PR #591, opened 2015-03-09
Improve plugin upload process backend

taylortom:
Switched to unzip2 to allow for __MACOSX folders (which are added if using the mac's built-in archive option in the right-click menu) - this folder could be removed during the unzip process, but I think we should take care of this, rather than expecting the user to do it

Added a more helpful error message for case where bower.json isn't in expected place

Changed the code to expect the folder to have been zipped from the inside, rather than to include a nested level, as this is what I've come to expect from similar systems.

### DIFF
--- a/plugins/content/bower/index.js
+++ b/plugins/content/bower/index.js
@@ -28,7 +28,7 @@ var origin = require('../../../'),
     _ = require('underscore'),
     util = require('util'),
     path = require('path'),
-    unzip = require('unzip'),
+    unzip = require('unzip2'),
     exec = require('child_process').exec,
     IncomingForm = require('formidable').IncomingForm;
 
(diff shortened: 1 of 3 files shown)


plugins/content/bower/index.js
```diff
@@ -779,30 +779,33 @@ function handleUploadedPlugin (req, res, next) {
...
-    });
-    ws.on('close', function () {
+    })
+    .on('close', function (data) {
       // enumerate output directory and search for bower.json
       fs.readdir(outputPath, function (err, files) {
-        if (err) {
-          return next(err);
-        }
+        if (err) return next(err);
```
dennis-learningpool: I'm not a huge fan of this formatting style, although we've used it in the frontend. I prefer my curlies.

plugins/content/bower/index.js
```diff
@@ -779,30 +779,33 @@ function handleUploadedPlugin (req, res, next) {
...
-        var canonicalDir = path.join(outputPath, files[0]);
-        try {
-          packageJson = require(path.join(canonicalDir, 'bower.json'));
-        } catch (error) {
-          return next(error);
+        var canonicalDir;
+
+        if(files.indexOf("bower.json") != -1) {
+          canonicalDir = outputPath
         }
```
dennis-learningpool: I like to cuddle my `else` to the end of the `if` brace.

plugins/content/bower/index.js
```diff
@@ -779,30 +779,33 @@ function handleUploadedPlugin (req, res, next) {
...
+        else {
+          var e = new Error("Cannot find bower.json file in the plugin root, please check your zip file and try again.");
+          return next(e);
+        }
+
+        try { packageJson = require(path.join(canonicalDir, 'bower.json')); }
```
dennis-learningpool: again, I'd prefer not to inline blocks like this.

dennis-learningpool:
Nice job. We don't have a complete coding style document (we originally looked at forking https://github.com/felixge/node-style-guide), so I'll not harp on too much about the formatting, but we should think about formalizing a style guide pretty soon.

+1

taylortom:
Thanks Dennis, have made those changes.

dennis-learningpool:
+1 those changes - thanks @taylortom

brian-learningpool:
+1

PR #601, opened 2015-03-18
Updated router to check isUserAuthenticated()

brian-learningpool:
This is required as not all routes were doing this, and when the user
isn't authenticated subsequent API calls can behave inconsistently.

This ensures that the user is redirected back to the login page in that
circumstance.

### DIFF
--- a/frontend/src/core/app/router.js
+++ b/frontend/src/core/app/router.js
@@ -86,6 +86,12 @@ define(function(require) {
         return $('.app-inner').append(new PermissionsView().$el);
       }
 
+      // Verify the user is authenticated
+      if (!this.isUserAuthenticated()  && (module !== 'user' && route1 !== 'login')) {
+        alert('Your session has expired, click OK to log on again');
+        return this.redirectToLogin();
+      }
+
       var routeArguments = arguments;
 
       // Set previous location object


frontend/src/core/app/router.js
```diff
@@ -86,6 +86,12 @@ define(function(require) {
         return $('.app-inner').append(new PermissionsView().$el);
       }
 
+      // Verify the user is authenticated
+      if (!this.isUserAuthenticated()  && (module !== 'user' && route1 !== 'login')) {
+        alert('Your session has expired, click OK to log on again');
```
dennis-learningpool: The alert is a bit bleh; we should print the message on the login page instead.

dennis-learningpool:
+1

frontend/src/core/app/router.js
darylhedley: Isn't there a way we could push them to the login page...

brian-learningpool: I agree that it's 'bleh', but thought leaving it as an alert() would aid the re-factoring when we revisit the entire error handling.

finbartracey1:
+1

darylhedley:
+1

PR #603, opened 2015-03-18
Fixes potential issue with copying course

brian-learningpool:
There's now a fall through when the _contentTypeParentId doesn't exist in
the parentIdMap variable.  This could happen with courses created in an
earlier version of the builder.


dennis-learningpool:
+1

taylortom:
+1

PR #608, opened 2015-03-23
Prototype for possible notify rework

taylortom:
This is an early version of what I was thinking.

A few key points: 
- Have split the notify into a singleton, which deals with all events, so rather than the `Origin.on(notify:xxxxx)` of olden-times, you have `(Origin).Notify.on('xxxxx')`. Similarly, triggering events is wrapped in this class, so this now becomes `Notify.error({ data })`, rather than `Origin.trigger('stuff', { data })`
- Separate plugins then listen in to events, and act as appropriate. I've set up a very basic [alert-type popup](https://github.com/taylortom/adapt_authoring/blob/feature/notifyRework/frontend/src/plugins/notify-prompt/index.js) based on the existing code, and a [console logger](https://github.com/taylortom/adapt_authoring/blob/feature/notifyRework/frontend/src/plugins/notify-console/index.js) as simple tests. The idea is that anyone can write one of these, so you could easily have logging to files, or a Facebook-style notifications list all working in-tandem.
- I've added several levels of notification, similar to console: fatal, error, warn, info, debug, to allow for easy plugin filtering (this also applies to type: push, alert etc.). I've also set up the wildcard * character to allow listening to all events (I've done this by triggering separate events, which isn't ideal, so any suggestions would be welcome - or indeed whether this is useful at all. I looked at dealing with a  listener queue manually, but not sure this provides any benefit). Notify in its current state triggers:
- Alert level (`fatal`, `error`, `warn` etc.)
- Alert type (arbitrary, plugins can deal with whatever they like here, in my two plugins, I've used `alert` and 'log')
- Wildcard `*`

Here are some example uses:
Triggering an event:
``` javascript
Origin.Notify.info({
   body: 'This is an info message',
   _type: "alert",
   _prompts: [{
      label: 'ok',
      _callback: callbackFunction
   }],
});
```
Listening to events:
``` javascript
Notify.on('alert', function(model) {
	// doing stuff
});
```
Adding/removing templates
``` javascript
Notify.addTemplate("alert", {
   _type: "alert",
   _prompts: [{
      label: 'ok'
   }]
});

Notify.removeTemplate("alert");
```

I've also added the ability to create notification templates, to allow for easier/cleaner triggering. These are just basically generic models which can be overridden. A good example of this is the alert popup type, which only needs one 'ok' button, and will look the same in all scenarios (with the exception of the message). Using a template, you can trigger an alert popup with:

``` javascript
Notify.fatal({
   body: 'this is the alert message',
   _template: 'alert'
});
```

One last thing I've done is to allow popup title/message fields to be polyglot keys (these are checked whenever the BackboneModel field is changed). 

E.g.
```javascript
Origin.Notify.addTemplate("alert", {
	title: 'app.errortitle',
});
```
Not sure is this is a useful feature or not. My thinking was that it would be easier to switch out translation libs if need be, although we should probably set up a separate global API which abstracts this.

Please let me know with regards to any suggestions, particularly any use-cases you have for using a notifications system, as I'd like to make it as flexible as possible. Similarly, let me know if you think there's a better solution.


brian-learningpool:
@taylortom, this looks cool.  Would it be possible for you to merge it into the new `develop` branch though?

taylortom:
Have created a branch here with all alerts replaced with these new notify tweaks:
https://github.com/adaptlearning/adapt_authoring/tree/feature/notifyIntegration

brian-learningpool:
@taylortom, I'm trying to get my head around how this hangs together.  Could you possibly implement this for `Origin.trigger('notify:prompt', deleteItem);` as found here (https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/core/editor/menu/views/editorMenuItemView.js#L210)?

taylortom:
Note to self: look into http://t4t5.github.io/sweetalert/

taylortom:
(and also actually finish this :\)

PR #619, opened 2015-03-27
Fixes #618

### LINKED ISSUE #618: /api/content/schema breaks if extension properties.schema don't define a pluginLocations attribute

dennis-learningpool:


taylortom:
+1, works for me

dennis-learningpool:
I think that we need to put something more in place to ensure an uploaded plugin is not malformed, and to do this at the point of upload. 

I think we need a bit more discussion about hard and fast plugin requirements before we do that, though.

finbartracey1:
+1

brian-learningpool:
+1

PR #621, opened 2015-03-30
Ab 332 performance bottlenecks

dennis-learningpool:
* Fixes a few uses of array.forEach in endpoints
* Fixes issue where an update of tenant details would be ignored by separate processes

### DIFF
--- a/lib/database.js
+++ b/lib/database.js
@@ -459,7 +459,7 @@ function getDatabase(next, tenantId, type) {
     }
 
     if (_databases[type] && _databases[type][tenantId]) {
-      if (_databases[type][tenantId].isStale(tenant.updatedAt)) {
+      if (_databases[type][tenantId].isStale(tenant)) {
         // database config has changed, need to refetch
         _databases[type][tenantId].disconnect();
         delete _databases[type][tenantId];
--- a/lib/database.js
+++ b/lib/database.js
@@ -39,6 +39,9 @@ Object.keys(errors).forEach(function(fnName){
   util.inherits(errors[fnName], Error);
 });
 
+// memoizes database instances
+var _databases = Object.create(null);
+
 // private hooks - only allow hooks defined here
 var _dbHooks = Object.create(null);
 _dbHooks['create'] = [];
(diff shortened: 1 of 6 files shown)


finbartracey1:
+1

brian-learningpool:
+1

lib/permissions.js
```diff
@@ -529,16 +523,57 @@ exports = module.exports = {
...
+            roles,
+            function (r, nextRole) {
+              async.each(
+                r.statement,
+                function (s, nextStatement) {
+                  s.resource = s.resource.replace('{{tenantid}}', tenantId);
+                  statements.push(s);
+                  nextStatement();
+                },
+                nextRole
```
taylortom: Is this a typo?

lib/database.js
```diff
@@ -328,19 +353,51 @@ function processDatabaseHooks (eventType, dbInstance, next) {
...
+ * 
+ * @param {string} type - e.g. 'mongoose'
+ * @param {object} tenant - a tenant record from DB
+ * @param {callback} next
+ */
+function createDatabaseInstance (type, tenant, next) {
+  var dbPath = path.join(getDriverPath(), type);
+  fs.exists(dbPath, function (exists) {
+    if (!exists) {
+      return next(new errors.notthere(dbPath));;
```
taylortom: Extra ;

lib/permissions.js
```diff
@@ -529,16 +523,57 @@ exports = module.exports = {
...
+                return callback(err);
+              }
+
+              // next retrieve bare policies
+              db.retrieve('policy', {user: userId}, function (error, records) {
+                if (err) {
+                  return callback(err);
+                }
+
+                if (!records || 0 === records.length) {
```
taylortom: Is having the comparison this way round the preferred format?

taylortom:
Looks good to me, +1

lib/permissions.js
brian-learningpool: I don't think there is a preferred format as such.  Some developers like to write it like this as it can sometimes accidental assignment when you really meant to compare values.

lib/permissions.js
brian-learningpool: `nextRole` looks to be the callback on line 532.  I think it might be more readable if it was written as `nextRole()` though.

taylortom: I thought you need the brackets to execute it though (this may not be correct)

dennis-learningpool: the third argument to async.each is a callback, so just passing in the named callback `nextRole` here. If it was `nextRole()` it would be invoked inline, which would be a bug.

The alternative would be to wrap it in an anonymous callback:
````
  function (err) {
    nextRole(err);
  }
````
but it would be a bit redundant, I think. It might improve readability.

lib/permissions.js
dennis-learningpool: I'm an adherent of [Yoda Notation](http://en.wikipedia.org/wiki/Yoda_conditions), but it's not an official style. I like my bugs to be explicit - chasing down an accidental assignment bug is a tiresome task.

lib/database.js
dennis-learningpool: Well spotted!

lib/permissions.js
taylortom: Ahhh yeah, I didn't realise it was being passed to the nested async.each, thought you meant to call it there and then.

taylortom:
+1

PR #623, opened 2015-04-01
Populate page attributes using schema defaults

taylortom:
Fixes https://adaptlearning.atlassian.net/browse/AB-355


darylhedley:
@taylortom - if we were to fix the schema issue on the backend we could have these defaults in the schema. This would then be created on the server and not on the front end. I think this is probably best discussed on a call

PR #631, opened 2015-04-10
Removes sidebar update class

### LINKED ISSUE #630: resetButtons method on sidebar doesn't remove class
When resetting sidebar buttons the class remains - still can cause problems when doing advanced plugins in the sidebar

darylhedley:
Also stringifies options object so it can work with nodejs v0.12

fixes #630

### DIFF
--- a/plugins/content/course/index.js
+++ b/plugins/content/course/index.js
@@ -47,6 +47,7 @@ function initialize () {
       var options = _.keys(req.body).length
       ? req.body
       : req.query;
+      var options = JSON.stringify(options);
       var search = options.search || {};
       var self = this;
       var orList = [];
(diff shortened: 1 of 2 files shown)


plugins/content/course/index.js
```diff
@@ -47,6 +47,7 @@ function initialize () {
       var options = _.keys(req.body).length
       ? req.body
       : req.query;
+      var options = JSON.stringify(options);
```
finbartracey1: you could remove the var. Already declared in line above var options = _.keys(req.body).length

dennis-learningpool: @darylhedley - actually, can you remove this bit? I think this is related to the issues with node v0.12, but we're only supporting 0.10.33 at the moment, and this workaround needs a closer look.

brian-learningpool:
+1

finbartracey1:
+1

PR #634, opened 2015-04-14
issue/#632

finbartracey1:
desc: update application.js to allow the use of express middleware by plugins.

Also removes the productmanger.js and its schema.


darylhedley:
+1

lib/application.js
```diff
@@ -278,12 +279,40 @@ Origin.prototype.createServer = function (options, cb) {
...
+
+ 
+
       server.use(auth.initialize());
       server.use(auth.session());
       server.use(express.static(path.join(app.configuration.serverRoot, 'frontend', 'src')));
       server.use(permissions.policyChecker());
-      server.use(server.router);
-      server.use(app.clientErrorHandler());
+        // loop through the plugins for middleware
```
dennis-learningpool: slight indentation issue on the following lines (extra 2 spaces)

finbartracey1:
good spot

lib/application.js
```diff
@@ -278,12 +279,40 @@ Origin.prototype.createServer = function (options, cb) {
...
+        var pluginManager = pluginmanager.getManager();
+        var plugins = pluginManager.getPlugins();
+        var plugin;
+        async.forEach(Object.keys(plugins), function (directory, cbOuter){
```
dennis-learningpool: since this process is only run once at server startup, you could use Array.forEach without worrying about blocking. It might help make this a bit more readable.

dennis-learningpool:
+1

brian-learningpool:
Sorry but it's -1 from me. There are a few issues with this:
1. As Dennis has said, `async` is not required on bootstrapping code such as this
2. This PR should be merged into `develop` not `master`
3. Error handling is required -- I think I read that at one stage it was thrown up into the calling `async.forEach` loop then ignored.

PR #638, opened 2015-04-15
Fixes around CKEditor

brian-learningpool:
- Changed to load CKEditor from their CDN
- Upgraded the version of JQuery to v1.1.2
- Added a hack for IE 9 and CKEditor
- Removed TinyMCE


brian-learningpool:
One thing at a time, @finbartracey1!  :)

There were some reports of more intermittent IE 10 issue, but it wasn't the same thing.  In IE 9, this would cause screens to freeze with the 'Loading...' message, while in IE 10 the issue was that the user couldn't enter text in any text inputs.

finbartracey1:
+1

darylhedley:
+1

PR #642, opened 2015-04-15
Fixes #620

### LINKED ISSUE #620: Preview displays default framework content.
When you first preview a new course from the authoring tool the preview is of the default framework build not the content that was entered into the authoring tool:

The second and subsequent times you click preview the authoring tool content is displayed.

The URL's in both instances are identical.

brian-learningpool:
This was due to the default framework course folder being copied over to
the temporary working/build directory.  The fix is to remove it on
installation.

### DIFF
--- a/install.js
+++ b/install.js
@@ -199,7 +199,15 @@ var steps = [
           return exitInstall(1, 'Framework install failed. See console output for possible reasons.');
         }
       
-        return next();
+        // Remove the default course
+        rimraf(path.resolve(__dirname, 'adapt_framework', 'src', 'course'), function(err) {
+          if (err) {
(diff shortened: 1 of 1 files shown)


darylhedley:
+1

dennis-learningpool:
+1 

Do we think there is a need to do a validation check on startup for cases where the tool is already installed?

taylortom:
+1

finbartracey1:
+1

brian-learningpool:
@taylortom, yes, I'll create a PR for develop.
@dennis-learningpool, for cases where the tool is already installed, it resolves itself after the first preview, but this does raise a point about how we manage file-system changes between versions.

PR #654, opened 2015-04-21
Use hashed filenames in output

brian-learningpool:
The hashed filename is unique so this gets around problems with spaces in
the filenames, special characters, etc.


nilslp:
+1

dennis-learningpool:
+1 for the change

+5 for the massive improvement in readability!

finbartracey1:
+1

PR #662, opened 2015-04-27
Added a check for NodeJS version

brian-learningpool:
This will generate a warning if not v0.10.33.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -308,6 +308,11 @@ Origin.prototype.startServer = function () {
     serverOptions.minimal = true;
   }
   
+  // Check the NodeJS version 
+  if (process.version !== 'v0.10.33') {
+    logger.log('warn', 'Incorrect version of NodeJS required -- this code requires v0.10.33 to run correctly.');
+  }
+
   var app = this;
   app.createServer(serverOptions, function (error, server) {
     if (error) {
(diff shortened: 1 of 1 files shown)


brian-learningpool:
Resolves #660.

taylortom:
+1
Did we want to have the version required pulled in from a config or anything? (not sure if this is needed anywhere else)

brian-learningpool:
I don't think it's needed anywhere else, and could possibly be removed when we re-work the code for v0.12.

lib/application.js
```diff
@@ -308,6 +308,11 @@ Origin.prototype.startServer = function () {
     serverOptions.minimal = true;
   }
   
+  // Check the NodeJS version 
+  if (process.version !== 'v0.10.33') {
+    logger.log('warn', 'Incorrect version of NodeJS required -- this code requires v0.10.33 to run correctly.');
```
dennis-learningpool: I think the message could worded a little better, here. I suggest:
"You are running an unsupported version of NodeJS (" + process.version + "). This application has only been fully tested on NodeJS v0.10.33. It is recommended that you use that version instead."

finbartracey1:
+1

lib/application.js
brian-learningpool: Me fail English?  That's unpossible.  

PR updated. ;)

dennis-learningpool:
+1 ralphie

PR #669, opened 2015-05-07
Changed user preferences to be session based

### LINKED ISSUE #663: Cached search/tags do not show on dashboard
After searching for courses, the search term is saved in a cookie, or local storage. When browsing to the dashboard later, this search term loads in the background, but does not appear in the search input. The user must click the "X" button next to the search input to clear it.

This also seems to happen with saved tags, but there is no way to remove saved tags via the UI if they do not show on the list.

darylhedley:
fixes #663


brian-learningpool:
+1

finbartracey1:
+1

PR #670, opened 2015-05-07
Fixes dashboard tags for grid view

### LINKED ISSUE #650: Restore display of "Tags" for projects/assets in thumbnail view
Tags were previously removed from thumbnails for projects and assets on the dashboard, as they could cause the vertical size of the thumbnail to grow, and thus throw off the grid layout for the dashboards.

We need to find a way to reintroduce the tags without affecting the layout. Tony Finlay suggested a horizontal scrolling element to list tags, and which would scroll when the tags list was long enough to start to wrap.

darylhedley:
fixes #650


nilslp:
+1

brian-learningpool:
+1

PR #671, opened 2015-05-08
Schema rework

taylortom:
Have created a PR for this to allow for easier discussion.


frontend/src/core/scaffold/scaffold.js
```diff
@@ -41,216 +37,276 @@ define(function(require) {
...
+	};
+
+	var setupArraySchemaField = function(field, key, schema, scaffoldSchema) {
+		if (field.items && field.items.properties) {
+			scaffoldSchema[key] = {
+				type: 'List',
+				itemType: 'Object',
+				subSchema: field.items.properties,
+				fieldType: 'List'
+			}
```
dennis-learningpool: missing `;`

frontend/src/core/scaffold/scaffold.js
```diff
@@ -41,216 +37,276 @@ define(function(require) {
...
-						itemType:field.items.inputType,
-						subSchema: field.items,
-						fieldType: field.items.inputType
-					}
+				scaffoldSchema[key] = {
+					type: 'List',
+					itemType:field.items.inputType,
+					subSchema: field.items,
+					fieldType: field.items.inputType
 				}
```
dennis-learningpool: missing `;`

frontend/src/core/scaffold/scaffold.js
```diff
@@ -41,216 +37,276 @@ define(function(require) {
...
-				// If value is an object - should give it some rights
-				// and add it as a field set - but not _extensions
-				if (fieldsets[key]) {
-
-					fieldsets[key].fields.push(key);
-
-				} else {
-
+			} else if (schema[key].type === 'object') {
+				debugger;
```
dennis-learningpool: should remove this line.

dennis-learningpool:
Looks pretty good to me. I like how you've implemented the frontend/backend separation with fieldset.json, but will defer to the opinions of @brian-learningpool and @darylhedley.

darylhedley:
I wonder if this should be separated further on the server. Having two schemas: One for db and one for frontend. This way we can have more flexibility when plugins or components want custom views without adjusting the schemas stored for the database.

Two suggestions:
1 - I think the schemas should be modules (js files) so we can have more flexibility and use mongoose schemas without a massive extraction layer. Mongoose is already extracted so I don't see the point in extracting it further to fit a schema standard into mongoose and frontend schemas. Plus imagine having to do this:

```
if (Schema.block.displayTitle === 'string') {
    Schema.block.displayTitle = String;
}
```

when we could have this:

```
"displayTitle": String
```

having to not run through a schema builder seems better to me. What do people think?

2 - Removing the properties object on components. This is important for security, validation and having schemas that force type. Mongoose has a way to enable multiple schemas into one collection.

This might completely change the way we set about the schemas though...

taylortom:
@dennis-learningpool: Have sorted your suggestions.

@darylhedley: 
1. Just to confirm - you're suggesting we swap the json files for JS files? 
2. I'd be up for removing the properties; these seem to make things unnecessarily complicated/confusing. Not sure how everyone else feels.

darylhedley:
Yeah by having the json fils as JS files - they're cached, can have JS in them so we can have subSchemas that automatically add id etc and we don't have to run through crazy schema parsers.

dennis-learningpool:
@darylhedley the reason the schemas are pure json at present is because the database module is pluggable. Right now, to add support for a different database engine, all you need to do is implement the Database interface, and provide a schema importer. I'm all for keeping things simple, but that means more than just simple to use - it also means simple to extend, which is something that would be lost if we removed the abstraction.

I'm prepared to accept that we don't need to support any other database than MongoDB, but it's not a call that any one person should make - I'd like to throw that question open to the other developers @brian-learningpool @finbartracey1 @dancgray.

taylortom:
@dennis-learningpool I think it's important that we're as 'decoupled' from specific plugins/libraries as possible, even if it means a bit more work for us.

PR #673, opened 2015-05-12
Fixes issue with plugins not running their JS in time

### LINKED ISSUE #672: Load order of plugins can sometimes cause an issue in IE9
The plugin require needs to be higher up the chain and the testing plugin doesn't need to be in there (strangely removing this and moving the plugin require up enables IE9 to load properly)

darylhedley:
fixes #672


brian-learningpool:
+1

finbartracey1:
+1

PR #677, opened 2015-05-14
Fixes issue with Firefox component list widths

brian-learningpool:
Noticed in 1024x768 in FF 36.


darylhedley:
+1 - nice catch!

PR #708, opened 2015-05-28
Fixes reference to files[0] to the actual folder name

### LINKED ISSUE #706: Uploading a zip on mac is painful
This is because of the __MACOSX file being read

darylhedley:
fixes #706


dennis-learningpool:
+1

plugins/content/bower/index.js
```diff
@@ -837,7 +837,7 @@ function handleUploadedPlugin (req, res, next) {
 
         // first entry should be our target directory
         var packageJson;
-        var canonicalDir = path.join(outputPath, files[0]);
+        var canonicalDir = path.join(outputPath, file.name.replace('.zip', ''));
```
dennis-learningpool: This will break if the zip file is not named the same as the contained plugin directory. It's not a normal scenario, but it might be likely.

Consider someone having downloaded the .zip for a plugin twice (resulting in a name like "adapt-contrib-foo(1).zip"), or if it has a timestamp or version number against it.

dennis-learningpool: maybe the solution is to iterate the directories and search for the existence of the bower.json file?

````
// pseudocode
var packagePath = false;
for each file in files {
  var bowerTest = path.join(outputPath, file.name, 'bower.json');
  if (exists(bowerTest)) {
    // found our man
    packagePath = bowerTest;
    break;
  }
}

if (packagePath) {
  // do stuff with packagePath
}
````

File: plugins/content/bower/index.js [hunk 2]
```diff
@@ -837,7 +837,7 @@ function handleUploadedPlugin (req, res, next) {
 
         // first entry should be our target directory
```
taylortom: Need to also update/remove the comment, as this no longer applies

dennis-learningpool:
I have a few comments against a commit here:

https://github.com/adaptlearning/adapt_authoring/commit/25a9aa0ea215f1eb757c508dd7f0bc2d4cf054c1

darylhedley:
@dennis-learningpool - fixed most of those now. Remaining one is the require - is the best way to do that with a try/catch?

dennis-learningpool:
@darylhedley yeah, I think it might be best to do a try/catch, and pass the error to asyncCallback:

````
try {
  packageJson = require(bowerPath);
} catch (e) {
  return asyncCallback(e);
}
````

File: plugins/content/bower/index.js [hunk 3]
```diff
@@ -835,43 +835,70 @@ function handleUploadedPlugin (req, res, next) {
...
+          async.some(directoryList, function(directory, asyncCallback) {
+            var bowerPath = path.join(outputPath, directory, 'bower.json');
+
+            fs.exists(bowerPath, function(exists) {
+              if (exists) {
+                canonicalDir = path.join(outputPath, directory);
+                try {
+                  packageJson = require(bowerPath);
+                } catch (e) {
+                  return asyncCallback();
```
dennis-learningpool: maybe should be:

````
return asyncCallback(null, e);
````

then on line 860:

````
    }, function (hasResults, err) {
          if (err) {
              return next(err);
          }
````

?

darylhedley: Don't think async.some passes back an error - if simply returns results

dennis-learningpool: yep - shame about that. In that case, maybe just logger.log the exception in the catch block, and pass a custom error to the callback on line 862.

dennis-learningpool:
+1
(although see previous comment about logger.log.)

brian-learningpool:
@darylhedley, can you please log the exception as @dennis-learningpool has suggested and we'll get this merged in?

brian-learningpool:
+1

dennis-learningpool:
sweet

PR #710, opened 2015-05-28
Separate the setting of assetType and metadata

### LINKED ISSUE #709: Wrong assetType set if ffmpeg isn't being used
See early return here: https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/filestorage/localfs/index.js#L380

Which results in certain file formats being categorised incorrectly (most document file formats are set as application for example).

One impact this has is that no document-type assets will be shown in the modal view for adding items to resources.

taylortom:
Fixes #709.


dennis-learningpool:
Haven't tried it locally, but looks good.

+1

darylhedley:
+1

brian-learningpool:
+1

PR #714, opened 2015-06-03
Fix to old asset not being able to remove themselves

### LINKED ISSUE #713: Remove asset button does not work with old asset links
This is because the model doesn't save when the course asset is not found.

darylhedley:
fixes #713

### DIFF
--- a/frontend/src/core/scaffold/views/scaffoldAssetView.js
+++ b/frontend/src/core/scaffold/views/scaffoldAssetView.js
@@ -255,6 +255,9 @@ define(function(require) {
                         console.log('error', error);
                     }
                 });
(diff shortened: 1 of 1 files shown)


frontend/src/core/scaffold/views/scaffoldAssetView.js
```diff
@@ -255,6 +255,9 @@ define(function(require) {
                         console.log('error', error);
```
dennis-learningpool: do we still remove `console.log`s when we find them?

dennis-learningpool:
+1

brian-learningpool:
+1

PR #717, opened 2015-06-04
Changes the way the modal window overlays

### LINKED ISSUE #716: Modal view allows double clicking on items - which for a matching component can break it
Double clicking on the items in a matching modal will cause the form to break and not update new data

darylhedley:
fixes #716


brian-learningpool:
+1

taylortom:
+1

PR #729, opened 2015-07-02
Corrected out-of-date tenant tests

brian-learningpool:


dennis-learningpool:
I have all our unit tests passing locally now, but the filepath issue with adapt-cli (https://github.com/adaptlearning/adapt-cli/issues/32) is preventing the install from completing.

PR #730, opened 2015-07-02
Re-worked language string compilation

brian-learningpool:
The en.json files is now the 'compiled' version.  I've renamed the original file to en-application.json and updated the Gruntfile and .gitignore.

This way, running a build won't flag the en.json file as a modified file, which was happening previously since the merge task was introduced.


darylhedley:
+1

dennis-learningpool:
+1

Issue #527, opened 2014-09-13
archiver  is in devDependencies it will be used in production


dancgray:
Should be moved to dependencies in packages.json

PR #732, opened 2015-07-03
Fixes #726 and #527

### LINKED ISSUE #726: issue with Phantomjs package install on some systems
See this thread: https://community.adaptlearning.org/mod/forum/discuss.php?d=778

It appears that phantomjs may be causing issues with some systems. 

Phantom may no longer be used in the tool, so perhaps should be removed from package.json. In any case, it should at least be moved to devDependencies, instead of dependencies.

brian-learningpool:
Tidied up the package.json to re-work devDependencies and update the README accordingly.

For the majority of end-users this will reduce the need to install PhantomJS, which had sometimes caused issues.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -162,7 +162,7 @@ module.exports = function(grunt) {
         // Save the configuration
         grunt.file.write(configFile, JSON.stringify(config, null, 2));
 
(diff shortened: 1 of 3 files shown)


dancgray:
+1

taylortom:
+1

dennis-learningpool:
+1

Issue #735, opened 2015-07-06
Adapt v2.0 - Make Publish/Preview with new framework Gruntfile


brian-learningpool:
The Gruntfile in the framework has changed and the code in the output plugin will require changing to support this.

NOTE: It should be possible to still work with version 1.x also, if the user has not upgraded.

taylortom:
What's needed in the server-build task? Currently consists of the following:
https://github.com/adaptlearning/adapt_framework/blob/develop/Gruntfile.js#L592-L593

brian-learningpool:
The `server-build` task is required to compile/build the JS, CSS and everything else required for a course to render, except the actual course JSON.  

The course JSON gets dropped into the correct folder during the publish/preview.

brian-learningpool:
Resolved in #801.

PR #740, opened 2015-07-08
Append build folder to output

taylortom:
Fixes: https://github.com/adaptlearning/adapt_authoring/issues/735

### DIFF
--- a/plugins/output/adapt/index.js
+++ b/plugins/output/adapt/index.js
@@ -147,7 +147,7 @@ AdaptOutput.prototype.publish = function (courseId, isPreview, request, response
 
             var args = [];
 
-            args.push('--outputdir=' + path.join(Constants.Folders.AllCourses, tenantId, courseId, Constants.Folders.Build));
+            args.push('--outputdir=' + path.join(Constants.Folders.AllCourses, tenantId, courseId));
             args.push('--theme=' + themeName);
             args.push('--menu=' + menuName);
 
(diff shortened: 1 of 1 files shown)


plugins/output/adapt/index.js
```diff
@@ -33,7 +33,9 @@ AdaptOutput.prototype.publish = function (courseId, isPreview, request, response
       tenantId = user.tenant._id,
       outputJson = {},
       isRebuildRequired = false,
+      // should this be Constants.Defaults.ThemeName?
```
dennis-learningpool: I think you're correct.

plugins/output/adapt/index.js
```diff
@@ -33,7 +33,9 @@ AdaptOutput.prototype.publish = function (courseId, isPreview, request, response
...
       themeName = 'adapt-contrib-vanilla';
+      // should this be Constants.Defaults.MenuName?
```
dennis-learningpool: ditto

dennis-learningpool:
Hey Tom - I think this is required for Framework 2.0 integration, right? Can you link an issue and update the description? Thanks.

taylortom:
Cool, I'll sort that now.

Another minor question: I've appended the build folder to the output, but is this actually needed? it doesn't look like anything else goes in there atm?
i.e. `path.join(Constants.Folders.AllCourses, tenantId, courseId, Constants.Folders.Build));`
or: `path.join(Constants.Folders.AllCourses, tenantId, courseId));`

brian-learningpool:
I don't think the build folder should be appended to the output, unless something else has changed in the framework?

brian-learningpool:
+1

dennis-learningpool:
+1

PR #763, opened 2015-07-23
Fixes #720 - Adds a confirmation to deleting an array item

### LINKED ISSUE #720: Confirmation when deleting an item such as a narrative item or question stem
I found it all to easy one day to hit the little orange X on a narrative item and lose all the content of it without any way to reinstate.

Can we have a confirmation box appear when we click on the orange Xs so this can't happen as easily?!

brian-learningpool:


dennis-learningpool:
+1

taylortom:
+1

PR #764, opened 2015-07-23
Fixes #719 - Allows <span> tags in editor

### LINKED ISSUE #719: Any <span> tag added in the wysiwyg editor is removed when going in to re-edit
If I add a <span> tag to the body text of a text element in the text component (using the source button in the wysiwyg editor), for example for tooltips - if I go in afterwards to make other changes to the body text the span tag and it's closing tag have disappeared and need re-adding.

brian-learningpool:


dancgray:
+1

dennis-learningpool:
+1

PR #770, opened 2015-07-29
Support for plugin globals

### LINKED ISSUE #765: Support plugin globals
Globals should be extracted from properties.schema and inserted into the course.json in the appropriate places. 

Possible touch-points are adding a component and adding an extension to a course.

brian-learningpool:
Fixes #765, adds support for plugin globals. 

This is dependent on the plugin implementing a `property.schema` file which has a `globals` section.


plugins/content/course/model.schema
```diff
@@ -31,256 +33,243 @@
       "required": true,
       "default": 0
     },
-    "_accessibility": {
```
dennis-learningpool: does moving this property into _globals break compatibility with v1.2?

lib/application.js
```diff
@@ -547,11 +547,16 @@ Origin.prototype.run = function () {
   // some modules are loaded async, so wait for them before actually starting the server
   this.once('modulesReady', function (app) {
     // Initialise the Polglot module so we can access language strings from the JSON
-    var phraseStrings = JSON.parse(fs.readFileSync(path.join(configuration.serverRoot, 'routes', 'lang', 'en.json'), {encoding: 'utf8'}));
-    app.polyglot = new polyglot({phrases: phraseStrings});
-    app.mailer = new Mailer();
-
-    app.startServer();
+    try {
+      var phraseStrings = JSON.parse(fs.readFileSync(path.join(configuration.serverRoot, 'routes', 'lang', 'en.json'), {encoding: 'utf8'}));
```
dennis-learningpool: might be a nice opportunity to read the desired lang file from config here

dennis-learningpool: This change will prevent correct installation of the tool. I've raised an issue to address this #777

brian-learningpool: I don't get it, that's been an issue since the first release.  That's why I wrapped it in a try-catch.

plugins/content/course/model.schema
brian-learningpool: No, it still compiles and renders ok.

dennis-learningpool:
+1

taylortom:
+1, works for me

PR #785, opened 2015-08-20
loading all dependencies using 'matchdep'

hrajotia:
Build process is failing after removing redundant packages from 'devDependencies'.(https://github.com/adaptlearning/adapt_authoring/commit/4891f56473c1e99680e11e5167b292e25826a948)


dennis-learningpool:
Nice catch! I guess we need to include the grunt build in our regression testing.

I have tested the fix and it works a treat.

+1

brian-learningpool:
+1

taylortom:
+1

PR #801, opened 2015-09-02
Issue/788

brian-learningpool:
Adds support for refreshing core plugins as part of the upgrade.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -363,6 +363,12 @@ Origin.prototype.createServer = function (options, cb) {
 };
 
 Origin.prototype.startServer = function (options) {
+
+  // Ensure that the options object is set.
+  options = typeof options === 'undefined' 
+    ? { skipVersionCheck: false } 
+    : options;
+
   // configure server
   var serverOptions = {};
   if (!this.configuration || !this.configuration.getConfig('dbName')) {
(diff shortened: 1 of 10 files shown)


lib/application.js
```diff
@@ -362,7 +362,7 @@ Origin.prototype.createServer = function (options, cb) {
   }, configuration.getConfig('dbName'));
 };
 
-Origin.prototype.startServer = function () {
+Origin.prototype.startServer = function (options) {
```
dennis-learningpool: options should be checked, and defaulted if not passed:
```
options = 'undefined' === typeof options ? { skipVersionCheck: false } : options;
```

or similar

lib/application.js
```diff
@@ -388,10 +388,10 @@ Origin.prototype.startServer = function () {
       callback();
     },
     function(callback) {
-      if (true === configuration.getConfig('isTestEnvironment')) {
+      if (true === configuration.getConfig('isTestEnvironment') || options.skipVersionCheck) {
         return callback();
       }
-
+
```
dennis-learningpool: if you enabled .editorconfig in your editor and merge the develop branch, you'll not add this kind of whitespace ;)

lib/application.js
```diff
@@ -449,13 +453,12 @@ Origin.prototype.startServer = function () {
...
 
-      // @TODO #771 - remove this check until the builder supports v2 of the framework
-      // if (installedFrameworkVersion == latestFrameworkTag) {
-      //   console.log(chalk.green('Adapt Framework %s'), installedFrameworkVersion);
-      // } else {
-      //   console.log(chalk.yellow('The Adapt Framework being used is %s - %s is now available'), installedFrameworkVersion, latestFrameworkTag);
-      //   isUpdateAvailable = true;
-      // }
+      if (installedFrameworkVersion == latestFrameworkTag) {
+        console.log(chalk.green('Adapt Framework %s'), installedFrameworkVersion);
```
dennis-learningpool: can we use the logger module rather than the console?

lib/application.js
```diff
@@ -571,7 +574,13 @@ Origin.prototype.restartServer = function () {
  * @api public
  */
 
-Origin.prototype.run = function () {
+Origin.prototype.run = function (options) {
+  if (!options) {
```
dennis-learningpool: just saw this - this would be better in the startServer function

lib/bowermanager.js
```diff
@@ -0,0 +1,338 @@
...
+
+           //is the module that loaded this modules requirement
+          if (modloaded === WAITFOR) {
+            app.bowermanager = instance;
+            preloader.emit('preloadChange', MODNAME, app.preloadConstants.COMPLETE);
+          }
+        }
+      };
+  }
+};
```
dennis-learningpool: With this addition, you should also remove plugins/content/bower.js, right? and update the plugins/content/component,extension,theme,menu plugins to inherit from here?

brian-learningpool: In time, yes.  I was going to come back to that  This would impact the plugin upload functionality and the initial install, etc. which is beyond what I was originally trying to resolve.

lib/application.js
brian-learningpool: I agree.

lib/bowermanager.js
dennis-learningpool: grand - that can come in a later pr then - might want to raise an issue for it?

brian-learningpool: I've added #802 and #803 to address the refactoring.

darylhedley:
+1 - all tested and worked with my courses. Although I reckon running npm install for the user would be a nice addition

brian-learningpool:
There's a separate issue for that here #804.  ;)

dancgray:
+1 works for me

dennis-learningpool:
+1 tested and working. nice one.

PR #805, opened 2015-09-03
Issue/#682

### LINKED ISSUE #682: Asset management: Unable to delete images uploaded in error (M - v0.2)
Suggested fix: add a delete button below the preview.

darylhedley:
fixes #682 - enables soft delete of assets ready for some sort of cron clean up

### DIFF
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -200,7 +200,12 @@ exports = module.exports = {
     }
 
     // Extend the search by _isDeleted to not return the soft deleted assets
-    search = _.extend(search, {_isDeleted: false});
+    // unless the api has a search of _isDeleted
+    if (options.search && options.search.hasOwnProperty("_isDeleted")) {
+      search = _.extend(search, {_isDeleted: options.search._isDeleted});
+    } else {
+      search = _.extend(search, {_isDeleted: false});
+    }
     
     database.getDatabase(function (error, db) {
       if (error) {
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -198,6 +198,9 @@ exports = module.exports = {
     } else {
       options.populate = _.extend(pop, options.populate);
     }
+
+    // Extend the search by _isDeleted to not return the soft deleted assets
+    search = _.extend(search, {_isDeleted: false});
     
     database.getDatabase(function (error, db) {
       if (error) {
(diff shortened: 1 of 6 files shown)


darylhedley:
Also adds deletedAt and deletedBy so we can clean up assets later. For now a soft delete is perfect!

lib/assetmanager.js
```diff
@@ -198,6 +198,9 @@ exports = module.exports = {
     } else {
       options.populate = _.extend(pop, options.populate);
     }
+
+    // Extend the search by _isDeleted to not return the soft deleted assets
+    search = _.extend(search, {_isDeleted: false});
```
dennis-learningpool: can we check search first to check if `isDeleted` is already set, and only default to false if not? That way, the api can still be used to fetch deleted assets if, for example, we want to add a gui to inspect assets that have been deleted.

lib/assetmanager.js
```diff
@@ -799,14 +842,22 @@ exports = module.exports = {
...
+    this.trashAsset(req.params.id, function(error) {
       if (error) {
         return next(error);
       }
 
       res.statusCode = 200;
       return res.json({success: true});
-    });
+    })
+    /*this.destroyAsset(req.params.id, function (error) {
```
dennis-learningpool: you can just delete this commented code

routes/lang/en-application.json
```diff
@@ -158,6 +158,7 @@
 	"app.upload": "Upload",
 	"app.assetfilters": "Filters",
 	"app.assetcollections": "Collections",
+	"app.assetconfirmdelete": "Are you sure you want to delete this asset? This cannot be undone",
```
dennis-learningpool: "This cannot be undone" is kinda fibbing?

darylhedley: To the end user it cannot be undone..but will remove it

lib/assetmanager.js
darylhedley: Was thinking about leaving it as this seems to be a temporary solution

lib/assetmanager.js
taylortom: +1 for this, a 'recycle bin'-type plugin could be useful

taylortom:
This is only a part-solution for #682. Is this what we want? 

We're just telling the user that they've deleted an asset when in actuality it's still there, we're just hiding it from them.

(I really like the soft delete feature, just seems a bit pointless without a 'hard' delete).

brian-learningpool:
I agree that we do need a 'hard' delete as well, but deleting assets completely from disk is a much more involved change.  Before doing so you would have to confirm that the asset isn't used in any course, because if it is and it's deleted, and the course is published again, it will have broken links.

I think something like the soft delete approach in this PR followed by a cron which checks if any 'soft' deleted assets not in use are physically deleted from disk makes sense.  I'm happy enough for that to wait to the next version though.

darylhedley:
Just pushed a bit of code to allow a trash bin approach for the frontend

brian-learningpool:
There's a slight UI issue as in if you delete an asset, it's still visible in the preview pane.  Then if you click Edit you get the Loading screen... and the page errors.  I think on successful deletion of the asset the preview pane should be removed.

taylortom:
Works well apart from issue mentioned by @brian-learningpool.

darylhedley:
@taylortom @brian-learningpool fixed that issue - could you test it again please?

taylortom:
Works for me, +1

brian-learningpool:
+1

PR #809, opened 2015-09-04
fix #807 add support for repeating arguments to logger

### LINKED ISSUE #807: Placeholder issue with logger.log() calls
When you call `logger.log()` with more than one placehoder, everything apart from the value of the first placeholder is ignored, e.g.

```
var string1 = 'foo';`
var string2 = 'bar';`
logger.log('info', 'This %s is a %s', string1, string2);
```
This gives the following output:
`This foo is a %s`

dennis-learningpool:


brian-learningpool:
+1, works a charm.

taylortom:
+1

PR #810, opened 2015-09-04
Issue/802

brian-learningpool:


darylhedley:
+1

lib/frameworkhelper.js
```diff
@@ -21,15 +23,14 @@ function flog(msg) {
 }
 
 function cloneFramework (next) {
-  var frameworkPath = FRAMEWORK_DIR;
-  fs.exists (frameworkPath, function (exists) {
+  fs.exists (path.join(process.cwd(), FRAMEWORK_DIR), function (exists) {
```
dennis-learningpool: instead of process.cwd, should use app.configuration.getConfig('root');

install.js
```diff
@@ -138,12 +138,12 @@ var configItems = [
     description: "Sender email address",
     default: ''
   },
```
dennis-learningpool: This might be a good opportunity to remove the prompt for SMTP settings also - it's not really required for most installs, at present.

lib/frameworkhelper.js
brian-learningpool: But the config file isn't written at that point?

dennis-learningpool: The serverRoot property is available as soon as the config module is required in - see https://github.com/adaptlearning/adapt_authoring/blob/develop/lib/configuration.js#L65

(This property needs to be available as early as possible, as much of the startup process depends on it)

dennis-learningpool:
+1 tested and working

PR #814, opened 2015-09-08
Fixes #813, issue with asset thumbnailPath when ffmpeg is not installed

### LINKED ISSUE #813: Bug: When ffmpeg is not installed, the thumbnailPath 'none' causes issues with routers
GET requests to /api/asset/thumb trigger the following error and result in broken images.

ENOENT, open '/Users/brian/Documents/Code/Adapt/adapt_authoring/data/master/none

brian-learningpool:
This was caused by thumbnailPath defaulting to the string 'none'.  I've added some language translation strings that were missing for the scaffoldAsset template.


darylhedley:
+1

dennis-learningpool:
+1

PR #815, opened 2015-09-09
Fixes #812 - associated assets now removed

### LINKED ISSUE #812: Bug: Deleting an article or page does not remove associated assets contained with in
During testing on `develop`, if you remove a block it seems to remove the assets used in the child component, but this doesn't seem to work for articles or courses.

brian-learningpool:
There was an issue where deleting an article did not remove any assets associated with components which were children of any blocks on that article.

No courseasset records were removed when a course was deleted.  This has now been corrected.


dennis-learningpool:
+1 - code looks ok, but have not tested yet

taylortom:
+1, works for me

PR #819, opened 2015-09-10
Fixes #818

### LINKED ISSUE #818: Bug: Assets on moved components hold reference to previous block
There is an edge case where assets added to components which are subsequently moved still hold a reference to their old block (in the `_contentTypeParentId` attribute).  In a future iteration we'll have to look at the courseassets collection -- I have a feeling it is due a re-factor -- but in the meantime this should be patched.

brian-learningpool:


plugins/content/component/index.js
```diff
@@ -153,12 +153,66 @@ Component.prototype.update = function (search, delta, next)  {
...
+      
+      ContentPlugin.prototype.update.call(self, search, delta, function (error, doc) {
         if (error) {
           return next(error);
         }
 
-        next(null);
+        // HACK -- This next 'if' block is required to keep the courseasset records
+        // in sync with the component's position.  This will be removed in a future
+        // re-factor to the courseasset collection when we can remove _contentTypeParentId.
```
dennis-learningpool: can you add the link to the ticket # here?

plugins/content/component/index.js
```diff
@@ -153,12 +153,66 @@ Component.prototype.update = function (search, delta, next)  {
...
+                  if (err) {
+                    return next(err);
+                  }
+
+                  next(null);
+                });
+
+              } else {
+                // Nothing to do.
+                next(null);
```
dennis-learningpool: you can get rid of this `else` and the following one by just returning on `database.getDatabase` on line 172 and `async.each`  on line 189 above, i.e.:

````
return database.getDatabase(//etc)
````
and
````
return async.each
````

plugins/content/component/index.js
brian-learningpool: Done.

darylhedley:
+1

dennis-learningpool:
+1

taylortom:
Works for me, +1

PR #830, opened 2015-09-15
Fixes #829 -- It should not be required to join on componenttype.id

### LINKED ISSUE #829: Bug: Component type label gets lost on plugin upgrade
This only seems to affect the page view,  as the edit screen still works as expected.  

I suspect this is due to the linkage on the `componenttype._id` value.

brian-learningpool:
The component name is unique enough.


dancgray:
+1

taylortom:
+1

PR #833, opened 2015-09-16
Asset scroll bug

### LINKED ISSUE #832: Issue with asset collection view being searched and a small amount of items returning
During asset management search - Chrome and only Chrome has an issue with re-rendering items into the DOM and showing them.

darylhedley:
fixes #832 - not a pretty fix but it does work. To test out - search for an asset you know that will only return one result. Then on this branch it should show.


nilssanderson:
+1

brian-learningpool:
+1

PR #844, opened 2015-09-18
Add support for responseText

### LINKED ISSUE #843: Plugin upload errors: add support for responseText
The [code that displays the error message](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/core/pluginManagement/views/pluginManagementUploadView.js#L34-L38) when uploading plugins seems to return a responseText value, but no responseJSON. Either there's a problem with the way that the server is returning the data, or we should handle both.

taylortom:
Fixes #843.

Note: not sure if this fix is actually required; there may be an issue with the way the server is returning the data rather than with the way the data is processed.

### DIFF
--- a/frontend/src/core/app/helpers.js
+++ b/frontend/src/core/app/helpers.js
@@ -165,6 +165,11 @@ define(function(require){
 
           return html;
         },
+        decodeHTML: function(html) {
+          var el = document.createElement('div');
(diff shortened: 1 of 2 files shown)


dennis-learningpool:
+1

brian-learningpool:
+1

PR #845, opened 2015-09-18
Various enhancements

hrajotia:
- changed ```Node.js``` to ```0.10```
-  explicitly specifying to use ```legacy infrastructure``` on Travis by putting ```sudo: required```
- only allow to build below branch names (case sensitive) on Travis
  - master
  - develop
- now limiting to ```depth: 10``` while cloning git repository
- caching ```node_modules``` directory content
- integrating with Gitter for notifications

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: node_js
 node_js:
   - '0.10'
 
-sudo: required
-
 branches:
   only:
     - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,44 @@
+---
 language: node_js
 node_js:
-  - "0.10.33"
-services: mongodb
+  - '0.10'
+
+sudo: false
+
+branches:
+  only:
+    - master
+    - develop
+    - /^(?i:release)\/.*$/
+    - /^(?i:hotfix)\/.*$/
+    - /^(?i:feature)\/.*$/
+    - /^(?i:issues?)\/.*$/
+
+git:
+  depth: 10
+
+services:
+  - mongodb
+
+cache:
+  directories:
+    - node_modules
+
 before_install:
-  - "npm install -g npm"
-  - "npm cache clean"
-  - "npm install -g adapt-cli"
+  - npm install -g npm
+  - npm install -g adapt-cli
+
+install:
+  - npm config set spin false
+
+matrix:
+  fast_finish: true
+
+notifications:
+  email: false
+  webhooks:
+    urls:
+      - https://webhooks.gitter.im/e/984c2f87fa167719344b
+    on_success: change  # options: [always|never|change] default: always
+    on_failure: always  # options: [always|never|change] default: always
+    on_start: false     # default: false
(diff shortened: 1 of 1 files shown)


dennis-learningpool:
Hi @hrajotia - I tried switching to container-based infrastructure before, but sudo is invoked somewhere in our dependencies installation. You may have to revert this, or find some way to remove the sudo requirement.

taylortom:
How do we feel about specifying branches? 

I'd have thought an all or nothing (just `master` and `develop`) situation would be more useful

hrajotia:
@dennis-learningpool As we are using ```mongodb``` service, which needs to be started using ```sudo```, that's why we have to add ```sudo: required``` over here.

@taylortom You are right! we should be targeting only those PR which are requested for ```master``` or ```develop``` branch.

I have updated the code. Please do let me know if there are any other concerns :)

.travis.yml
```diff
@@ -1,8 +1,41 @@
...
+
+install:
+  - npm config set spin false
+  - npm install
+
+matrix:
+  fast_finish: true
+
+notifications:
+  email: false
```
dennis-learningpool: Do we want to disable email notifications entirely?

hrajotia: I'm not sure for this configuration!
but are you really interested in receiving emails when there is success or failure for any PR?

dennis-learningpool: I believe that, by default, only the PR Author and the Committers are included in any email notification.

hrajotia: Definitely and those conversation email are generated by GitHub itself. As of now we haven't set any recipient for which Travis should dispatch email notification!

Please refer following docs - http://docs.travis-ci.com/user/notifications/#Email-notifications

This have clearly mentioned that we need to supply email id(s) of recipients who all are interested in receiving email when there is success and/or failure for PR(s).

dennis-learningpool: Under the default configuration, Travis-CI should email the PR author and committers on a failed build. If you do not specify any config for email under notifications, it should default as described here: http://docs.travis-ci.com/user/notifications/#Notifications

> By default, email notifications are sent to the committer and the commit author, if they are members of the repository (that is, they have push or admin permissions for public repositories, or if they have pull, push or admin permissions for private repositories).

Also, under the email notification section: http://docs.travis-ci.com/user/notifications/#How-is-the-build-email-receiver-determined%3F

> By default, a build email is sent to the committer and the author, but only if they have access to the repository the commit was pushed to.

dennis-learningpool: I can also confirm that I have received email notifications directly from travis on changes to the build resulting from PRs I have submitted up till now.

hrajotia: Thank you for explaining me this :+1:.
Previously I was thinking that people might not be interested in receiving the build status emails. Now I'm thinking about it may help them to check the status via email once build finished on Travis because sometime it takes time to get completed.

I have removed the email configuration, now it should work as you have mentioned above :)

dennis-learningpool:
+1

Thanks for this @hrajotia - this will really improve the build performance, and I'm looking forward to seeing the notifications on gitter!

brian-learningpool:
+1

taylortom:
+1

PR #849, opened 2015-09-18
Minor fixes

brian-learningpool:
- Scrambled HTML on hovering project title
- Thumbnail API now returns generic image rather than error

### DIFF
--- a/frontend/src/core/app/helpers.js
+++ b/frontend/src/core/app/helpers.js
@@ -188,9 +188,9 @@ define(function(require){
 
         ifImageIsCourseAsset: function(url, block) {
           if (url.length !== 0 && url.indexOf('course/assets') == 0) {
-            return block.fn(this);;  
+            return block.fn(this);
           } else {
-            return block.inverse(this);;
+            return block.inverse(this);
           }
         },
         
--- a/frontend/src/core/app/helpers.js
+++ b/frontend/src/core/app/helpers.js
@@ -175,10 +175,6 @@ define(function(require){
           }
         },
 
-        console: function(context) {
-          return console.log(context);
-        },
-
         getAssetFromValue: function(url) {
           var urlSplit = url.split('/')
           var fileName = urlSplit[urlSplit.length - 1];
@@ -190,6 +186,14 @@ define(function(require){
 
         },
 
+        ifImageIsCourseAsset: function(url, block) {
+          if (url.length !== 0 && url.indexOf('course/assets') == 0) {
+            return block.fn(this);;  
+          } else {
+            return block.inverse(this);;
+          }
+        },
+        
         getThumbnailFromValue: function(url) {
 
           var urlSplit = url.split('/')
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -794,6 +794,7 @@ exports = module.exports = {
           });
         });  
       } else {
+        // TODO -- Refactor and use a disk-based image (#850)
         // Return a generic image when no thumnail exists
         var img = new Buffer(DEFAULT_THUMBNAIL_BASE64, 'base64');
 
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -28,7 +28,8 @@ var MODNAME = 'assetmanager',
     WAITFOR = 'contentmanager',
     STREAM_BUFFER_SIZE = 64 * 1024,
     THUMBNAIL_WIDTH = '?',
-    THUMBNAIL_HEIGHT = '200';
+    THUMBNAIL_HEIGHT = '200',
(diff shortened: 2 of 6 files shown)


frontend/src/core/app/helpers.js
```diff
@@ -190,6 +186,14 @@ define(function(require){
 
         },
 
+        ifImageIsCourseAsset: function(url, block) {
+          if (url.length !== 0 && url.indexOf('course/assets') == 0) {
+            return block.fn(this);;
```
dennis-learningpool: couple of double semi-colons here and on line 193

frontend/src/core/app/helpers.js
```diff
@@ -190,6 +186,14 @@ define(function(require){
 
         },
 
+        ifImageIsCourseAsset: function(url, block) {
```
dennis-learningpool: the name of this is a bit weird. can you change it to something like "fixAssetURL" or something a bit more imperative

lib/assetmanager.js
```diff
@@ -28,7 +28,8 @@ var MODNAME = 'assetmanager',
...
...xveYxweY1xe45yfZB0fZF2gJN6hJZ6hZZ8hph9h5iBiZuDjJ6FjqCGj6CGkKGHkKKIkqSKk6WLlKWQmaqTnK2Una2Vna6Xn6+YobGZorKcpbSep7efqLihqrqkrLylrr2qssGwuca1vMq6ws+8xdLAyNXBydbFzdrHz9zI0N3K0t/L0t/M1ODN1ODQ2OPS2uXS2+bV3OjW3ejY3+ra4u3c5O/e5fDf5vHg6PPh6fTj6/bk6/bl7Pfm7ffm7vjn7vno7/np8PoQVD/gAAACnklEQVR4Ae3WbVMSURjG8bNUbmaY1EZqkgpYJhlFUWRkQhKgaQ+JmQ/hw/39P0MzNbfT5l4ehGv3Re3/9XXm9+rM3Ebky6sHXqr/vNnF5S2xZaR21QzexMqBxXljOI3Wuuc6Vwyryc55DrGRVjSOcZvYiQgybMji3MhdPAxhpyT20NvswmUAcZ2SbHoAIjvyLQ0grgMgugMgugMgugMgugMgugMgugMgugMgvgMgvgMgvgMgvgMgvgMgugOgFtsB0EiH7QBoskt1MFSjOhgaPaA6GFoJx1HI08FEqI5sXtLFVqiOLOhimeLcKYGyulikOPZmB3ISPTveQM7tnp3UQE4hIqftRONIOSLnR9lhO6BWIZ0YzFmvZMZcdyxTWZf+y9mc9pQ5baodlnNYNL6Kh6E4u9Pmr6Z3Q3COsuZM2SOqg39Fme5sOEGOs8F25kxgc2Rn24C2uU4dOXWuU0BOgevMIGeG66SRk+Y648gZ5zp55OS5ThE5Ra7TQE6D6+wNBTNDexLQx2fHfTn4uCgHMtfNIwBZnZ3hIGZ4J5gxALI7shrkrCIGQXZHKmeZCmYQZHek6vgVp4oZDNkdaST/ZJINwFgguyP7S8lTZWkfMwiyO9rJWvXp/PyT12snghkMYQcGGAAxHMAoxHMAoxDPAYxCPAcwCvEcwCjEciCjEMdBDIB6dJpNzGDowk7TdZt2BkJ2RxljfNAnxADI7iijEGYwhJ0ARiHAgJ7bHT+jEGJAOcEOYBRShusoo5AyRMfHKFRXhuAARqM6gOE7gOE7gOE7t0w0Tip2Yid2Yuffc/jFTuz8H07mbRh5fie8InZuRuTkI3JeROMkPoh5GIHzUsR8vx8687grYkTe3XVDRK7dey/yyxE5/vo5rDryu5+fSKzliGdNYwAAAABJRU5ErkJggg==';
```
dennis-learningpool: Can't we just ship a default image on file, and read it from disk? Seems like the default thumbnail is something a designer should be able to change, and we wouldn't lose the chunk of ram that storing this string will take up.

frontend/src/core/app/helpers.js
brian-learningpool: I'm not thrilled about the name myself but this function is a handlebars helper and the name is consistent with the other similar helpers in the same file, e.g. `ifValueEquals()`, `ifUserIsMe()`.  Handlebars doesn't support the ability to call a function with an `if` statement in front.

dennis-learningpool: Possibly those functions need named better too - the main problem I have is that the the function name looks like it tests something, but actually modifies the arguments.

lib/assetmanager.js
brian-learningpool: I've raised this as an enhancement (#850) for when we finally get a designer involved in the project.

dennis-learningpool:
+1

frontend/src/core/app/helpers.js
taylortom: This doesn't modify anything though does it?

Given: 
```
{{#customIf}}
    <div class="ifDiv"></div>
{{else}}
    <div class="elseDiv"></div>
{{/customIf}}
```
Retuning `block.fn(this)` will insert the `ifDiv`, whereas returning `block.inverse(this)` will insert the `elseDiv` (as I understand it).

darylhedley:
+1

frontend/src/core/app/helpers.js
dennis-learningpool: Hi @taylortom - you're correct. I misunderstood the handlebars helper.

Issue #811, opened 2015-09-08
Running preview in the authoring tool should run grunt dev not grunt build so we can have...


darylhedley:
SOURCE-MAPPING!

taylortom:
@darylhedley do we need anything more than this? https://github.com/taylortom/adapt_framework/blob/develop/Gruntfile.js#L609-L612

(with references changed here: https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/output/adapt/index.js#L158-L160)

taylortom:
Ollie's new LESS task should now also give us LESS sourcemaps :tada:

PR #857, opened 2015-09-22
Add support for adapt_framework -> grunt server-build:dev

taylortom:
Partial fix for #811


taylortom:
Also see https://github.com/adaptlearning/adapt_framework/pull/734/

brian-learningpool:
@taylortom, I really like this.  Source maps would be of major help to developers trying to debug client-side code.  The only thing I have issue with is using the `isProduction` config to toggle this.  This config is already used in determine whether the web application should run in production mode.

Perhaps a new course setting would be more applicable?

taylortom:
Fine by me - are you thinking another config.json option? (e.g. `generateOutputSourcemap`)

brian-learningpool:
Yes, either in Configuration Settings or Project Settings.

taylortom:
@brian-learningpool this work should be done now

taylortom:
TODO: test this in older versions of the framework to make sure the params don't cause any unexpected problems if not used.

brian-learningpool:
@taylortom, this is really useful.  I've just been testing this locally and I think there's one small thing it's missing though.  When the `_generateSourcemap` value changes you you should trigger a flag to indicate that a full rebuild is required.  Have a look at the `rebuildCourse` event:

```
 app.emit('rebuildCourse', tenantId, courseId);
```

plugins/output/adapt/index.js
```diff
@@ -215,14 +218,14 @@ AdaptOutput.prototype.publish = function (courseId, isPreview, request, response
...
+            { expand: true, cwd: path.join(FRAMEWORK_ROOT_FOLDER, Constants.Folders.AllCourses, tenantId, courseId, Constants.Folders.Build), src: ['**/*'] },
+          ]).finalize();
 
         } else {
           // No download required -- skip this step
           callback();
-        }        
-      }   
+        }
+      },
```
dennis-learningpool: usually wary of trailing commas in a JS array like this - is it ok?

taylortom: Think I ran the jscs grunt task at some point which may have added this.

taylortom:
+1 for @brian-learningpool's changes

brian-learningpool:
+1 for yours too! :+1:

dennis-learningpool:
+1 for everyone!

PR #860, opened 2015-09-23
Vagrant update

### LINKED ISSUE #767: Update vagrant file to make compatible with current version
The vagrant setup for the authoring tool is a wee bit out of date. We need to adjust it to take advantage of the new install process, and lock down the version of nodejs to the currently supported version.

dennis-learningpool:
fix #767 

This updates the Vagrant  provisioning to bring it in line with the current install process. This makes installing the tool a little bit simpler for most end users, and also gives developers a common development environment, which should make it easier to debug issues.

The steps for installing using this process are:

1. [Install vagrant](https://docs.vagrantup.com/v2/installation/)
2. [Install virtualbox](https://www.virtualbox.org/wiki/Downloads)
3. Launch a terminal and cd into the adapt_authoring repository
4. Run the command `vagrant up`
5. Go make a cup of tea/coffee

The `vagrant up` command should take around 10 minutes to complete. Once it finishes, you can browse to http://localhost:5000 to access the tool, and login with the following details:

 email: admin
 password: password

You don't need to keep your terminal window open. When you want to shut down, just cd into the adapt_authoring repository again, and run the command `vagrant halt` and restart later with `vagrant up`.


Vagrantfile
```diff
@@ -9,23 +9,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
-  # accessing "localhost:3000" on the host will access port 3000 on the guest machine.
-  config.vm.network "forwarded_port", guest: 3000, host: 3000
+  # accessing "localhost:5000" on the host will access port 5000 on the guest machine.
+  config.vm.network "forwarded_port", guest: 5000, host: 5000
   config.vm.network "forwarded_port", guest: 27017, host: 27027
 
-  # Enable provisioning with chef solo
-  # see http://berkshelf.com/ for details of this
-  # end
-  config.berkshelf.enabled = true
```
dennis-learningpool: Removed the chef dependency to reduce system configuration overhead for end users

install.js
```diff
@@ -114,30 +115,30 @@ var configItems = [
...
-    hidden: true
-  },
-  {
-    name: 'fromAddress',
-    type: 'string',
-    description: "Sender email address",
-    default: ''
-  },
+//  {
+//    name: 'smtpService',
```
dennis-learningpool: these settings can still be added to config.json later, but I think most users will not add them when installing the tool for the first time

brian-learningpool: I think these should be left in the manual install, otherwise we'll have to document how to enable email notifications for password resets.  Can these be uncommented but defaulted/bypassed for the Vagrant script?

dennis-learningpool: we can default them, but it'll have to be with "you@example.com" type values (can't use empty values, as far as I've seen, annoyingly)

brian-learningpool: If empty values can't be defaults I think defaulting to something like 'Gmail', 'youremail@gmail.com', 'yourgmailpasswordhere' is probably obvious enough.

brian-learningpool:
+1, nice job!

darylhedley:
+1

brian-learningpool:
This requires testing in a Windows environment.

lc-thomasberger:
installation of Vagrant on Windows 7 worked.

When running vagrant up in directory I got stuck with Warning: Connection timeout. Retrying ...
(see attached Image)
vagrantwindows

danjah:
Hmm, failed on Win8.1 (https://gist.github.com/danjah/7f228d03b37d34b2fbad), I'll try again just now.

lc-thomasberger:
after enabling VT-x in the Bios I am able to run `vagrant up` without connection errors. Vagrant runs a while, installs some ubuntu packages and installs node and npm. But vagrant fails with the following error message.
vagrant npm error

danjah:
I had a colleague run through the Vagrant/Vbox setup on Win7, enable VT-x and clone this branch. Vagrant up died on a 404 for the Ubuntu link:

vagrant-issue-win7

kyleberkland:
Following on from danjah last comment.
All I have done is typed in the command prompt "vagrant up" and let it run until I hit this error.
https://gist.github.com/anonymous/c393a790aab5942bfca3

kyleberkland:
Continuing on with my last post I have now "vagrant halt" the process. and re-ran the command "vagrant up" and have received this screen here: 
post2

Then After fixing the stdin: is not a tty with adding : config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"

I have now ran my "vagrant up" and this has fixed the tty problem. Now when I enable the server on my Virtual box I get this loading screen with the following errors.

post

brian-learningpool:
@kyleberkland, what do you see if you try a `vagrant reload`?

kyleberkland:
@brian-learningpool , thanks for the response.
First when I prompt "vagrant reload" it came up with this:
upload
With the same errors on reloading localhost:5000  :
upload1
 
I now have ran "vagrant box update" to check for updates in ubuntu/trusty64.
Downloaded the new version.
Ran the command vagrant up.
And the result is the same unfortunately.

danjah:
Quick update, I just tried this vagrant process again on a clean Win7 machine, I needed to manually install the berkshelf plugin, and also manually install chef, at which point `vagrant up` worked.

Felt like progress, but then I hit this error (ignore the BSON stuff):

vagrant@vagrant-ubuntu-trusty-64:/vagrant$ node server
{ [Error: Cannot find module '../build/Release/bson'] code: 'MODULE_NOT_FOUND' }
js-bson: Failed to load c++ bson extension, using pure JS version
info: [20 Oct 2015 01:47:23 +00:00] Adapt Builder v0.1.2
info: [20 Oct 2015 01:47:23 +00:00] The Adapt Framework being used is v2.0.0 - v2.0.3 is now available
info: [20 Oct 2015 01:47:23 +00:00] Run "node upgrade.js" to update to the latest version

/vagrant/node_modules/serve-favicon/index.js:55
  if (!path) throw new TypeError('path to favicon.ico is required');
                   ^
TypeError: path to favicon.ico is required
    at favicon (/vagrant/node_modules/serve-favicon/index.js:55:20)
    at Origin.createServer (/vagrant/lib/application.js:254:16)
    at /vagrant/lib/application.js:476:11
    at /vagrant/node_modules/async/lib/async.js:537:21
    at /vagrant/node_modules/async/lib/async.js:222:13
    at iterate (/vagrant/node_modules/async/lib/async.js:123:13)
    at process._tickDomainCallback (node.js:492:13)

brian-learningpool:
I've corrected some of the symlink issues which Windows has struggled with.  I'm still running into problems on Windows now with the clone of the adapt_framework part of the installation.

brian-learningpool:
@danjah, @lc-thomasberger, or @kyleberkland, would you be able to help testing this updated pull request on Windows?  If so, you should:
- Run the command prompt as an Administrator
- Test with a checkout of the repo to the root of your drive, e.g. C:\adapt_authoring

lc-thomasberger:
Log-File:
https://gist.githubusercontent.com/lc-thomasberger/1bca99c04e044e0385bd/raw/4d4995e19bd3f91a5267d0845edfcfd6e4e8596c/log.txt

davewallace:
Hi @brian-learningpool same set of cmds? Just clone & checkout the vagrant branch, then 'vagrant up'?

Same failure as @lc-thomasberger maybe? Here's my log: https://gist.github.com/davewallace/84ab0b1d88441bd23b3a

brian-learningpool:
Hi  @davewallace, thanks for checking it out.  Your error does look very similar to @lc-thomasberger's.

At this point I think we're going to have to take a different approach with Vagrant on Windows, which will probably involve removing the symlink and storing the code on the VM itself.  This is not ideal from a Windows developer's perspective, but it should be enough to provide end-users with a working installation.

davewallace:
What would windows devs be sacrificing? I don't know much at all about vagrant's inner workings, it seems like such an excellent idea having an agnostic dev environment setup. Where is a good starting point for understanding more about how vagrant works, particularly on windows?

brian-learningpool:
In this case Windows devs would be sacrificing the ability to make code changes on their Windows box that are reflected on the Vagrant (Ubuntu) server.  The authoring tool will still run when installed locally on Windows, it's when we try to use it with Vagrant that @lc-thomasberger and myself ran into all kinds of problems, e.g. http://blog.prolificinteractive.com/2015/01/21/getting-vagrant-nodejs-windows-play-well-together.

A good starting point to learn about Vagrant is probably https://www.vagrantup.com/.

lc-thomasberger:
I have updated [Vagrant](https://github.com/lc-thomasberger/adapt_authoring/tree/vagrant-windows) to copy the AT-files into the VM on Windows. For Mac/Unix Users nothing will change.

I have tested it on Windows 7 and Mac OSX.

Would be nice to have more people test the updates on different Platforms. Thanks

brian-learningpool:
+1, I've successfully tested this on Windows.

davewallace:
+1 Win8.1, it is a thing of beauty - thanks guys. So does this then come with the aforementioned caveats for Windows users?

lc-thomasberger:
@davewallace thanks for testing on Windows 8. Yes unfortunately right now there are the caveats for Windows users that @brian-learningpool described. You can however access all the files in virtual machine using `vagrant ssh` and `cd /home/vagrant/`.

lc-thomasberger:
+1 tested on Windows and Mac

taylortom:
+1

Works a charm for me on mac. Couple of suggestions:
- I had to run as `sudo` to get write access to a virtualbox folder (not sure if this is to be expected, and is probably a configurable virtualbox option).
- Add @dennis-learningpool's instructions to the README
- Possibly output the username/password in the console

PR #864, opened 2015-09-23
Fixes #863

### LINKED ISSUE #863: Broken preview and publish after deleting asset
# Description
If you delete an asset from the asset manager, the (soft) deletion is correctly applied in the 'assets' collection, but it is NOT applied in the 'courseassets' collection of the DB. Thus, when previewing or publishing, the reference to the asset still exists in the course, but the asset is not accessible (because it's soft-deleted), so there's an error (it doesn't find what it thinks should be there...)
I've been using Robomongo to see what's going on in the DB.

I'm using branch 'develop' of the authoring tool, 
commit: a8f8a4b69bc5fac6968ea4e03eb723fc2446a3ec
the system is an Ubuntu 14.04 (it's really an lxc container with ubuntu 14.04, running in an Ubuntu 14.04 host).

# Steps to reproduce
0. Create minimal course, just 1 page, whatever minimalist thing that you can preview.
1. Add an image using the asset manager (new asset appears in the 'assets' collection)
2. Somewhere in the course, e.g. in a page, use that asset
3. You should be able to preview the course normally
4. Delete the image using the asset manager. The corresponding asset in the 'assets' collection has the property '_isDeleted set to true, so the asset doesn't show up in the asset manager any more -correct-). BUT, the corresponding asset in the 'courseassets' collection still has the property '_isDeleted' set to 'false'
5. Try to preview. You should get an error box in the browser, that just says 'Error:' (with no message).
The errors I see in the server console are:
```
info: [23 Sep 2015 10:03:14 +00:00] Retrieving JSON
TypeError: Cannot read property 'filename' of undefined
    at /home/pabby/adapt_authoring/lib/outputmanager.js:642:70
    at Promise.<anonymous> (/home/pabby/adapt_authoring/lib/assetmanager.js:241:18)
    at Promise.<anonymous> (/home/pabby/adapt_authoring/node_modules/mongoose/node_modules/mpromise/lib/promise.js:177:8)
    at Promise.emit (events.js:95:17)
    at Promise.emit (/home/pabby/adapt_authoring/node_modules/mongoose/node_modules/mpromise/lib/promise.js:84:38)
    at Promise.fulfill (/home/pabby/adapt_authoring/node_modules/mongoose/node_modules/mpromise/lib/promise.js:97:20)
    at Promise.resolve (/home/pabby/adapt_authoring/node_modules/mongoose/lib/promise.js:114:23)
    at Promise.<anonymous> (/home/pabby/adapt_authoring/node_modules/mongoose/node_modules/mpromise/lib/promise.js:177:8)
    at Promise.emit (events.js:95:17)
    at Promise.emit (/home/pabby/adapt_authoring/node_modules/mongoose/node_modules/mpromise/lib/promise.js:84:38)
    at Promise.fulfill (/home/pabby/adapt_authoring/node_modules/mongoose/node_modules/mpromise/lib/promise.js:97:20)
    at Object.cb (/home/pabby/adapt_authoring/node_modules/mongoose/lib/query.js:1016:22)
    at Object._onImmediate (/home/pabby/adapt_authoring/node_modules/mongoose/node_modules/mquery/lib/utils.js:137:16)
    at processImmediate [as _immediateCallback] (timers.js:354:15)
```

I've noticed that if I delete the corresponding document in the 'courseassets' collection, thus cleaning up the erroneous asset reference in the course, things work ok, e.g. I can preview and publish the course.

brian-learningpool:
The _isDeleted flag should be exluded from the filter criteria whenever a preview or publish is run, otherwise assets which were added to a course before they were deleted will cause errors.

### DIFF
--- a/frontend/src/core/assetManagement/views/assetManagementCollectionView.js
+++ b/frontend/src/core/assetManagement/views/assetManagementCollectionView.js
@@ -95,6 +95,7 @@ define(function(require){
               this.collection.reset();
             }
 
+            this.search = {_isDeleted: false};
             this.search = _.extend(this.search, {
                 tags: {
                     $in: this.tags
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -198,20 +198,6 @@ exports = module.exports = {
     } else {
       options.populate = _.extend(pop, options.populate);
     }
-
-    // Extend the search by _isDeleted to not return the soft deleted assets
-    // unless the api has a search of _isDeleted
-    if (options.search && options.search.hasOwnProperty("_isDeleted")) {
-      search = _.extend(search, {_isDeleted: options.search._isDeleted});
-    } else {
-      search = _.extend(search, {_isDeleted: false});
-    }
-    
-    // A preview/publish should not care if an asset has been flagged as deleted,
-    // so remove this from the filter criteria.
-    if (options.isPublish && search.hasOwnProperty('_isDeleted')) {
-      delete search._isDeleted;
-    }
     
     database.getDatabase(function (error, db) {
       if (error) {
@@ -613,6 +599,14 @@ exports = module.exports = {
     var orList = [];
     var andList = [];
 
+    if (search.hasOwnProperty('_isDeleted')) {
+      if (search._isDeleted == 'false') {
+        search._isDeleted = false;
+      } else {
+        search._isDeleted = true;
+      }
+    }
+    
     // convert searches to regex
     async.each(
       Object.keys(search),
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -207,6 +207,12 @@ exports = module.exports = {
       search = _.extend(search, {_isDeleted: false});
     }
     
+    // A preview/publish should not care if an asset has been flagged as deleted,
+    // so remove this from the filter criteria.
+    if (options.isPublish && search.hasOwnProperty('_isDeleted')) {
+      delete search._isDeleted;
+    }
+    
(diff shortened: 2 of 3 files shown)


darylhedley:
+1

dennis-learningpool:
I think a better fix is to remove the _isDeleted flag options from the backend api function completely, and amend the frontend logic for the assetmanager/assetpicker so that they pass _isDeleted: true when hitting the api.

The backend api should be agnostic about the properties of the collection - the logic should really come from the client.

darylhedley:
@dennis-learningpool @brian-learningpool - I can update the API from this PR
https://github.com/adaptlearning/adapt_authoring/pull/805
Then put the logic in the frontend - although...I think we should also update the template to have a little deleted icon. So admins can see all the assets

taylortom:
That sounds v.useful @darylhedley :)

darylhedley:
@taylortom could be handy to toggle it if admin too...almost like the admin can bring back deleted assets

brian-learningpool:
@darylhedley I think if you're going to allow the admin to see ghost assets then an option to 'undelete' is essential.  I'd prefer that implemented in a separate PR though if you could fix the API so it doesn't break preview/publish.

taylortom:
Works for me, +1

darylhedley:
Just a note on this - @brian-learningpool is looking over this PR here:

https://github.com/adaptlearning/adapt_authoring/pull/868

which adds the ability to restore based upon permissions - and if super admin you get to see all the assets with a big cross over it
screen shot 2015-09-24 at 10 50 18

PR #869, opened 2015-09-25
Notify Rework 2: just when you thought it was safe to go back into the water.

taylortom:
Have simplified things quite a bit, and left heavy lifting up to plugins, rather than trying to implement everything ourselves.

Notify is now just a global object, that can easily be tapped into by other plugins using `Notify.register(NAME, FUNCTION)`, which allows you to attach any function to Notify. Calling a Notify plugin is as simple as `Origin.Notify.console(data)`. I was intending to integrate events (notification open, closed etc. eventually -- Notify already extends backbone events, so it's just a case of triggering this appropriately).

I've used a consistent model structure for all test, although in most cases, this isn't enforced in any way - we may want to introduce a specific JS model for this.

I've set up 3 different plugins to start with, which you can toggle through by clicking the global menu icon in the top left. So far, we have: 

**Alert** 
Supports titles, body messages, custom icons, nested/callback popups (i.e. 'Are you sure you want to delete' -> 'Thing deleted' -- see [notify-test/index.js#L30-L40](https://github.com/adaptlearning/adapt_authoring/blob/notify-rework/frontend/src/plugins/notify-test/index.js#L30-L40)). It's also possible to show an alert by simply passing a string, in which case the value is shown in the title field. Have a look at [sweetalert's demo page](http://t4t5.github.io/sweetalert/) for some more example configurations.
-- (powered by [sweetalert](https://github.com/t4t5/sweetalert)).

**Toast**
Shows non-intrusive notifications using toastr. Supports various notification types (warning, error, info, success), custom display times, animation types...see the [toastr demo page](http://codeseven.github.io/toastr/demo.html) for all configurations.
-- (powered by [toastr](https://github.com/CodeSeven/toastr))

**Console**
Just hooks into the `window.console`, and allows `log`, `error`, `warn` etc. If no type is specified, `log` is the default. Can also just pass a string (rather than an object), which also results in a `console.log` See [notify-test/index.js#L9](https://github.com/adaptlearning/adapt_authoring/blob/notify-rework/frontend/src/plugins/notify-test/index.js#L9).


frontend/src/core/notify/plugins/alert/index.js
```diff
@@ -0,0 +1,86 @@
...
+		}
+		data.type = "confirm";
+		openPopup(data);
+	};
+
+	var init = function() {
+		Origin.Notify.register('alert', Alert);
+		Origin.Notify.register('confirm', Confirm);
+
+		// shortcuts to override window methods
```
taylortom: Override window methods

frontend/src/core/notify/plugins/alert/index.js
```diff
@@ -0,0 +1,86 @@
...
+		window.alert = alert = Alert;
+		window.confirm = confirm = Confirm;
```
taylortom: Override window methods

brian-learningpool:
+1

brian-learningpool:
Merging this without two +1s as @dancgray, @taylortom and myself have all worked on this.

PR #882, opened 2015-10-06
Fixes #881 - bug in isAssetExternal()

### LINKED ISSUE #881: Bug: .match() is not a function error
There is a bug in the `isAssetExternal()` helper when the `url` string is empty.

brian-learningpool:
The .match() call will fail if the url is empty.


dancgray:
+1

darylhedley:
+1

PR #888, opened 2015-10-12
Fix preview security

brian-learningpool:
This PR closes preview URLs being accessible to anyone who had them.  Now when a request is made to 'main.html', the code checks that the user has access to the course.  If they do, the course is cached in the user's session, so that the permissions check is cached for subsequent requests (for images, CSS, etc.).

### DIFF
--- a/routes/preview/index.js
+++ b/routes/preview/index.js
@@ -7,9 +7,19 @@ var Constants = require('../../lib/outputmanager').Constants;
 var usermanager = require('../../lib/usermanager');
 var helpers = require('../../lib/helpers');
 var logger = require('../../lib/logger');
+var methodOverride = require('method-override');
+var util = require('util');
 
+server.use(methodOverride());
 server.set('views', __dirname);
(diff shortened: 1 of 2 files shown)


routes/preview/index.js
```diff
@@ -5,33 +5,103 @@ var server = module.exports = express();
...
+      logger.log('warn', 'Preview: User %s does not have permission to view course %s on tenant %s', user._id, courseId, tenantId);
+      
+      return next(new PreviewPermissionError());
+    }    
 
-    var isServerRequest = true;
-    if (isServerRequest || (currentUser && (currentUser.tenant._id == tenant))) {
-      res.sendFile(requestedFile);
+    // When requesting the first page, check the user has access to this course.
+    helpers.hasCoursePermission('*', user._id, tenantId, {_id: courseId}, function(err, hasPermission) {
```
dennis-learningpool: should the first param be `'read'` instead of `'*'`?

PR #893, opened 2015-10-16
Resolves #892

### LINKED ISSUE #892: Bug: IE 9 losing bullet or numbered list formatting after a save
Using any editor with the CKEditor control, for example, any 'body' field, if you add a bullet or numbered list and click Save, the next time you go to edit what you've entered the list formatting will be lost and the text entered will be contained within `<p>` tags.

brian-learningpool:
The important part here was to call .val(), rather than .html(), as the content is intially placed in a <textarea> control, then processed by CKEditor.

I've also updated CKEditor to the latest version.


nilssanderson:
+1

dennis-learningpool:
+1

PR #906, opened 2015-10-23
Notify updates

taylortom:
Lots of updates to the notifications to bring in line with latest Notify updates. Main points:
- Tidied up confirm a bit to allow for more straightforward calling
- Changed relevant Notify.confirm calls to Notify.alert
- Updated all window.alerts to Notify.alerts
- Externalised all messages in these Notify calls to the application.json

There were a lot of whitespace changes caused by my editor auto-removing spaces etc. so may be best to browse the commits rather than the overall diff.


dancgray:
+1, does the translation for default OK/cancel need to be made here https://github.com/adaptlearning/adapt_authoring/blob/develop/frontend/src/core/notify/plugins/alert/index.js#L19

taylortom:
Have made those changes here: 
https://github.com/adaptlearning/adapt_authoring/blob/notify-updates/frontend/src/core/notify/plugins/alert/index.js#L75-L79

brian-learningpool:
@taylortom, can you fix the conflicts in this PR when you get a chance?

brian-learningpool:
+1

Issue #588, opened 2015-03-04
Preview loads main.html


taylortom:
...which loads the course in an iframe, and so isn't responsive. The scrollbar also doesn't work.

brian-learningpool:
@taylortom, this is something we may improve with support for version 2.0 of the framework.  Do you have any thoughts?

If we done link to main.html we'll get the SCORM version of the course at index.html, which will throw JS errors about not being able to connect to the LMS.

taylortom:
I think leaving this for the v2.0.0 update is fine. 

Ollie's set up a SCORM harness html file in the new spoor which allows you to test basic tracking without an LMS - we could use this instead (there's also talk of swapping this in the default build when you have spoor installed, so that we get rid of those annoying 'not connected to LMS' messages completely).

taylortom:
Fixed in https://github.com/adaptlearning/adapt_authoring/commit/b81d576a6d8ad75b388b29001dd4d5a72e1cb8ef

PR #907, opened 2015-10-23
Fix the preview pop-up URL

juanpescador:
On clicking the "Preview" button, the popup window was pointing to the
course's 'main.html' file, which isn't generated by the grunt task. The
correct file is 'index.html'.

This happened to me on Windows, having cloned adapt_authoring v0.1.3. After following the installation steps and creating an initial course with some dummy content, the "Preview" button was opening a blank window, with the URL ending in 'main.html'. Chrome's developer tools showed a 404. After looking around the directory hierarchy I found the `repo_root\temp\<tenant-id>\adapt_framework\courses\<course-id>\build\` directory, which doesn't contain a 'main.html', but does have an 'index.html'. Manually changing the preview window's URL to index.html correctly displayed the course I had just created.

Searching issues and the wiki pages, #588 mentions that 'main.html' is/was the file that loads the course in an iframe (possibly for [SCORM previewing purposes][MainHtmlScormPreview]?), but the [changelog][ChangelogIframeRemoved] says iframe previews were removed back in January 2015. Is this a Windows-only issue maybe?

I saw that [adapt_framework's Gruntfile.js][AdaptFrameworkGruntfile] has a task that creates an 'index.html' so I `blame`d it to see if it was originally 'main.html', but it doesn't seem so.

I also noticed that in routes/preview/index.js, if there is no file specified it [defaults to `Constants.Filenames.Main`][PreviewRouteDefault], whose [value is 'main.html'][ConstantsFilenamesMain], so that might need updating too.

[MainHtmlScormPreview]: https://github.com/adaptlearning/adapt_framework/wiki/Setting-up-your-development-environment#viewing-the-build
[ChangelogIframeRemoved]: https://github.com/adaptlearning/adapt_authoring/blob/master/CHANGELOG.md#010---2015-01-26-dennis-heaney-dennislearningpoolcom
[AdaptFrameworkGruntfile]: https://github.com/adaptlearning/adapt_framework/blob/8f0e638add1d499e22c45ff3110b6f92802bd0de/Gruntfile.js#L72
[PreviewRouteDefault]: https://github.com/adaptlearning/adapt_authoring/blob/1f135b1f28621e63a03300fdf858f05c6bc9a1ec/routes/preview/index.js#L28
[ConstantsFilenamesMain]: https://github.com/adaptlearning/adapt_authoring/blob/1f135b1f28621e63a03300fdf858f05c6bc9a1ec/lib/outputmanager.js#L75

### DIFF
--- a/frontend/src/core/editor/global/views/editorView.js
+++ b/frontend/src/core/editor/global/views/editorView.js
@@ -108,7 +108,7 @@ define(function(require){
       var courseId = Origin.editor.data.course.get('_id');
       var tenantId = Origin.sessionModel.get('tenantId');
 
-      window.open('/preview/' + tenantId + '/' + courseId + '/main.html', 'preview');
+      window.open('/preview/' + tenantId + '/' + courseId + '/index.html', 'preview');
     },
 
     previewProject: function(event) {


taylortom:
Thanks for the feedback @juanpescador 

The main.html is copied across from [spoor's required folder](https://github.com/adaptlearning/adapt-contrib-spoor/tree/master/required). The main reason we launch from there is to avoid the 'cannot connect to LMS' errors/popups that show.

We're currently in the process of [updating the grunt build process](https://github.com/adaptlearning/adapt_framework/pull/781) to enable the exclusion of plugins, which will render this solution ineffective. 

I'm not sure what we're planning to do when the new grunt stuff comes into play, but this PR may come in useful then.

brian-learningpool:
Hi @juanpescador, thanks for raising this.  I've re-created the issue you've raised and it's not a Windows specific issue.

I've tracked the issue to a bug in v2.0.0 of the adapt_framework project.  Basically adapt-contrib-spoor was missing from the list of dependencies.  This plugin has some 'required' folders, which are essential to both the SCORM packaging and previewing without connecting to any LMS.  

Running `node upgrade.js` should have fixed this for you, but I've hot-fixed the master branch in v0.1.3 to refer to v2.0.1 of the framework instead now.  Thanks again for raising this.

PR #913, opened 2015-10-27
Fixes #912 - Extensions are now matched on name rather than _id

### LINKED ISSUE #912: Bug: Manage Extensions screen links on _id, rather than name
This is a problem as the name is constant but the _id changes with every new version of a plugin updated.

The issue was resolved with components but until now seems to have been missed on extensions.

brian-learningpool:
This fixes an issue whereby enabled extensions would appear not enabled on a course if the extensions were upgraded.


taylortom:
+1

dennis-learningpool:
+1

PR #917, opened 2015-10-28
Fixes #916, a full build is now triggered when customStyle changes

### LINKED ISSUE #916: Adding custom style does not trigger a full build
Any classes or styling added via the CSS/LESS field should be bundled with the .css file when preview or publish is clicked.

brian-learningpool:


dennis-learningpool:
+1

taylortom:
+1

PR #919, opened 2015-10-28
Fix for JavaScript coding error when alternativeAttribute is set

brian-learningpool:


frontend/src/core/scaffold/views/scaffoldAssetView.js
```diff
@@ -284,6 +284,7 @@ define(function(require) {
 
             // Check if alternative attribute should be used
             if (alternativeAttribute) {
+                attributesToSave = [];
```
dennis-learningpool: should `attributesToSave` only be initialized if it doesn't already exist? Can it exist at this point?

brian-learningpool: It's declared at the top of the function, but given a value of `null` in certain cases.

dennis-learningpool:
+1

darylhedley:
+1

PR #926, opened 2015-11-03
Allow assets to be defined in articles.json

### LINKED ISSUE #925: Asset paths defined in articles.json not converted
Use case: I have an extension which requires custom assets to be defined. When the course is built, these are copied into the correct place, but the filepath in the JSON is incorrect.

Only course, components and content objects work:
https://github.com/adaptlearning/adapt_authoring/blob/master/lib/outputmanager.js#L656-L657

taylortom:
Fixes #925

### DIFF
--- a/lib/outputmanager.js
+++ b/lib/outputmanager.js
@@ -232,7 +232,6 @@ OutputPlugin.prototype.writeCourseJSON = function(jsonObject, destinationFolder,
       var filename = (collectionType === 'config')
         ? path.join(destinationFolder, Constants.CourseCollections[collectionType].filename)
         : path.join(destinationFolder, outputJson['config']._defaultLanguage, Constants.CourseCollections[collectionType].filename);
-        debugger;
 
       fs.outputFile(filename, data, function(err) {
         if (err) {
--- a/lib/outputmanager.js
+++ b/lib/outputmanager.js
@@ -657,10 +657,12 @@ OutputPlugin.prototype.writeCourseAssets = function(tenantId, courseId, destinat
                   var lang = jsonObject['config']._defaultLanguage;
                   var newAssetPath = "course/" + lang + "/assets/" + encodeURIComponent(asset.filename);
 
-                  var objects = ['config', 'course', 'article', 'block', 'component'];
-                  for(var i = 0, count = objects.length; i < count; i++) {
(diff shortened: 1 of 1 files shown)


lib/outputmanager.js
```diff
@@ -653,14 +654,21 @@ OutputPlugin.prototype.writeCourseAssets = function(tenantId, courseId, destinat
                   // TODO -- This global replace is intended as a temporary solution
                   var replaceRegex = new RegExp("course/assets/" + asset.filename, 'gi');
 
-                  jsonObject['component'] = JSON.parse(JSON.stringify(jsonObject['component']).replace(replaceRegex, "course/" + jsonObject['config']._defaultLanguage + "/assets/" + encodeURIComponent(asset.filename)));
-                  jsonObject['course'] = JSON.parse(JSON.stringify(jsonObject['course']).replace(replaceRegex, "course/" + jsonObject['config']._defaultLanguage + "/assets/" + encodeURIComponent(asset.filename)));
+                  var lang = jsonObject['config']._defaultLanguage;
+                  var newAssetPath = "course/" + lang + "/assets/" + encodeURIComponent(asset.filename);
+
+                  var objects = ['config', 'course', 'article', 'block', 'component'];
```
taylortom: Set up to do the replace for:
- config
- course
- article
- block
- component

brian-learningpool: I don't think there was a need to declare the array of `objects` like this.  There is already a constant you can use: `Object.keys(Constants.CourseCollections)`

brian-learningpool:
+1

lib/outputmanager.js
```diff
@@ -229,9 +229,10 @@ OutputPlugin.prototype.writeCourseJSON = function(jsonObject, destinationFolder,
 
     async.each(Object.keys(Constants.CourseCollections), function(collectionType, callback) {
       var data = JSON.stringify(outputJson[collectionType], undefined, 2);
-      var filename = (collectionType === 'config') 
+      var filename = (collectionType === 'config')
         ? path.join(destinationFolder, Constants.CourseCollections[collectionType].filename)
         : path.join(destinationFolder, outputJson['config']._defaultLanguage, Constants.CourseCollections[collectionType].filename);
+        debugger;
```
dennis-learningpool: whoops

taylortom: :sweat_smile:

dennis-learningpool:
+1

PR #931, opened 2015-11-10
Fixed issue with _supportedLayout not working on component selection

### LINKED ISSUE #930: Issue with components list not filtering the correct layout options based upon their schemas
When trying to select a component that is full width and not single width the side bar still gives the option to add the component to any layout option

darylhedley:
fixes #930


brian-learningpool:
+1

taylortom:
+1

PR #938, opened 2015-11-17
Setup jscs to enforce code style

### LINKED ISSUE #342: Whitespace adjust grunt task or git hook
Add a grunt task or tasks (or git hook) to apply or revert 2 space indentation.

taylortom:
Fixes #342.

Have based the [jscs settings](https://github.com/adaptlearning/adapt_authoring/blob/issue/342/.jscsrc) on a node-inspired configuration. I've set the task to auto-enforce the rules where possible, so we may want to take some of the options out. 

See [here](http://jscs.info/rules) for the full list of available options.


taylortom:
There's also a git hook plugin that can run this task before commit

brian-learningpool:
+1. This throws 240 warnings on the current codebase right now.  Anyone fancy tidying them up?

dennis-learningpool:
+1

.jscsrc
```diff
@@ -0,0 +1,31 @@
...
+  "requireSpaceAfterLineComment": true,
+  "requireSpaceAfterBinaryOperators": true,
+  "requireSpaceBeforeBinaryOperators": true,
+  "requireSpaceBeforeBlockStatements": true,
+  "requireSpaceBeforeObjectValues": true,
+  "requireSpacesInFunction": {
+    "beforeOpeningCurlyBrace": true
+  },
+  "requireTrailingComma": {
+    "ignoreSingleLine": true
```
dennis-learningpool: does this mean adding a comma after the final entry in an array? this can be dodgy in front end JS, i think.

brian-learningpool: Perhaps we should use `"disallowTrailingComma": true` instead?

lc-thomasberger:
+1

PR #954, opened 2015-12-02
Fixes #953, resolves bug with new asset tagging

### LINKED ISSUE #953: Issues with uploading a new asset which has a new tag
As a Course Creator, if you upload a new asset and create a new tag at the same time (by entering a new tag followed by a comma) you will get an error because of a DB schema validation.

Furthermore, the Save and Cancel buttons will no longer be available.

brian-learningpool:
The 'tags' attribute was being passed up as an array of strings, rather than an array of ObjectId values.

### DIFF
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -489,6 +489,7 @@ exports = module.exports = {
               asset,
               function (createError, assetRec) {
                 if (createError) {
+                  logger.log('error', createError);
                   // if the record creation fails, remove the file that was uploaded
                   storage.deleteFile(storedFile.path, function (delErr) {
                     if (delErr) {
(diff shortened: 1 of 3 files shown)


lib/assetmanager.js
```diff
@@ -496,8 +497,8 @@ exports = module.exports = {
                       logger.log('error', 'Failed to delete stored file in assetmanager', storedFile.path);
                     }
 
-                    // return the creation error
-                    return next(createError);
+                    // Indicate that an error has occurred
+                    return res.status(500).json({success: false, message: "An error occured during upload, please contact an Administrator."});
```
taylortom: Should this be localised?

taylortom:
+1

dennis-learningpool:
+1

lib/assetmanager.js
brian-learningpool: Hi @taylortom, yes it should.  We don't seem to have localised any of the Node.js error messages right now though.  Should probably revisit them all with one ticket, like we did for the front end.

Issue #961, opened 2015-12-08
the vagrant setup isn't quite right for development


ryangadams:
It's not set up to turn on debugging by default, nor does it forward the debugging port to the host. Also, it would be nice if it reloaded the application when files are changed.

PR #962, opened 2015-12-08
Issues/961 - Vagrant tweaks

ryangadams:
fixes adaptlearning/adapt_authoring#961

1. Turn on debugging and forward the port to the host
2. Don't run the pm2 provision script as root, since that makes pm2 need to be managed as root, so you then would need to `vagrant ssh; sudo su; pm2 restart foo`, which is sub-optimal.
3. Make pm2 watch for changes in the project and reload when it sees them (I have not reliably confirmed that this works, but it should since I'm following the pm2 documentation)


brian-learningpool:
Hi @ryangadams, this looks fine, but can you merge into `develop` rather than `master` please?

ryangadams:
damnit Github. Closing to re-open on the right branch…

Issue #966, opened 2015-12-14
"download" referred to as "publish internally"


ryankilf:
The "download" button may have originally been called "publish" (and this is understandable), however it has been relabelled as "download".

PR #967, opened 2015-12-14
Publish re work

ryankilf:
This resolves issue #966 ( https://github.com/adaptlearning/adapt_authoring/issues/966 )

I've also snuck in a spelling fix.


brian-learningpool:
+1

dancgray:
+1

ryangadams:
This should be targeting develop not publish-re-work?

ryangadams:
Doing a search suggests there's a method on the OutputPlugin interface called publish. Does that need to be renamed? If so, any implementations will need to be updated too (which I think is only plugins/publish/Adapt at this point).  Additionally, routes/lang/en-application.json has some app.publish-like strings which might need updated.

PR #969, opened 2015-12-14
Issue/957 - Updated text for vagrant installations

ryan-lynch:
The vagrant installation instructed the user that they will be prompted for configuration settings, but this is not the case usually with vagrant installations, the args are taken from the vagrant setup. Tweaked the text here so it is clear.

Run into an issue where only the first 5 args were saved to config.json - copied optimist.argv code to the top of the install script to fix this.


install.js
```diff
@@ -414,7 +422,22 @@ var steps = [
   },
   // all done
   function finalize (next) {
-    console.log("Installation complete.\nRun the command 'node server' (or 'foreman start' if using heroku toolbelt) to start your instance.");
+    var passedArgs = process.argv.slice(2);
```
dennis-learningpool: lines 425 - 434 are calling out to be refactored into their own function.

install.js
```diff
@@ -24,6 +29,8 @@ var app = builder();
 var masterTenant = false;
 var superUser = false;
 
+var isVagrantInstallation = false;
```
dennis-learningpool: this var could be initialized here with the function suggested in my comment below.

install.js
```diff
@@ -414,7 +422,22 @@ var steps = [
...
+    for (var i = 2; i < passedArgs.length; i++) {
+      if (passedArgs[i] == undefined) {
+        isVagrantInstallation = false;
+        break;
+      } else {
+        isVagrantInstallation = true;
+      }
+    }
+
+    if (isVagrantInstallation) {
```
brian-learningpool: You're looping over each argument here and setting `isVagrantInstallation = true` multiple times.  It's probably better to just do something like:

`isVagrantInstallation = process.argv.length == 2 ? false : true;`

Also, can you rename this variable to `isVagrant`, for brevity?  The 'installation' part can be inferred from the fact that you're in the installation script (install.js).

You could also move this check further up the script (at the start) so other pieces of the code can know if if it's a Vagrant install.

ryan-lynch: @brian-learningpool see most recent commit, I have taken yours and Dennis' points on board.

install.js
```diff
@@ -24,6 +29,12 @@ var app = builder();
 var masterTenant = false;
 var superUser = false;
 
+var isVagrant = function () {
+  if (process.argv.length != 2) {
+      return true;
+    }
```
dennis-learningpool: minor indentation issue here. also, an explicit `return false;` would be a bit clearer:

````
if (process.argv.length != 2) {
      return true;
}

return false;
````

File: install.js [hunk 5]
```diff
@@ -24,6 +29,12 @@ var app = builder();
 var masterTenant = false;
 var superUser = false;
 
+var isVagrant = function () {
+  if (process.argv.length != 2) {
```
dennis-learningpool: I'd also be inclined to use `> 2` over `!= 2`

dennis-learningpool:
+1

brian-learningpool:
+1

PR #972, opened 2015-12-18
Resolves #959

### LINKED ISSUE #959: The output plugin should trigger events when a preview is successful and when a zip file is ready for download
This would be useful for allowing other parts of the code to tap into preview and download.

brian-learningpool:
Added 'zipCreated' and 'previewCreated' events.


dennis-learningpool:
+1

PR #975, opened 2015-12-23
Issue-973 - validate course content function moved to helpers.js

ryan-lynch:
Moved ValidateCourseContent in to helpers.js so it can be used throughout the application. Also removed it from editorView.js


frontend/src/core/app/helpers.js
```diff
@@ -238,7 +238,70 @@ define(function(require){
           } else {
             return block.inverse(this);
           }
+        },
+
+        validateCourseContent: function() {
+          // Store current course
+          var currentCourse = Origin.editor.data.course;
```
ryangadams: I wonder whether the course model should be injected as an argument to this function instead of relying on Origin? That would allow this function to have tests written for it and would allow additional usage in a non-course context.

frontend/src/core/editor/global/views/editorView.js
```diff
@@ -45,14 +46,14 @@ define(function(require){
       Origin.editor.pasteParentModel = false;
       Origin.editor.isPreviewPending = false;
       this.currentCourseId = Origin.editor.data.course.get('_id');
-
+      this.currentCourse = Origin.editor.data.course;
```
ryangadams: I'll give this PR a +1, but I wonder whether it's safe to keep hold of the current course object like this. I assume it is, but would appreciate a more experienced authoring tool developer confirming.

brian-learningpool:
+1

ryangadams:
:+1:

PR #978, opened 2015-12-23
issue/#976 Added an event to 'onAssetClicked' function

bell87:
Added an event which is triggered when an asset is selected within the assetManagementModalView.

The event passes the assetLink object so it can be used in other places.


frontend/src/core/assetManagement/views/assetManagementModalView.js
```diff
@@ -71,6 +71,7 @@ define(function(require) {
         		assetFilename: filename
         	}
 	        this.data = assetObject;
+            Origin.trigger('modal:assetSelected',assetObject);
```
dennis-learningpool: slight indentation issue here

dennis-learningpool:
+1

brian-learningpool:
+1, though it would be better if you could correct the indentation and spacing.

PR #992, opened 2016-01-19
Resolves #991

### LINKED ISSUE #991: The projectDetails view is a bit over-zealous with the patch .save() call
An issue has been noticed where the call to `this.model.changedAttributes()` does not detect brand new properties, e.g. the `_globals` introduced with version 2.0 of the framework.  This can lead to quirks in published courses, as the schema renders the UI correctly but the data does not save as expected.

A resolution to this would be to submit all model attributes back up, if nothing is changed.

brian-learningpool:
Passes the full model up to the .save() when only tags have been passed.

This will correct the issue whereby brand new properties were not being
detected by the changedAttributes function.


dennis-learningpool:
+1

lc-thomasberger:
+1

PR #993, opened 2016-01-20
Resolves #989, adds 'Copy ID to clipboard' functionality

### LINKED ISSUE #989: Allow the user to view the _id of of contentobjects, articles, blocks, components
We should provide a user friendly way of displaying the _id value to the user, so that it can be used in plugins that require it, e.g. assessment, or within inline HTML for things like routing to other course pages from hyperlinks.

One possible way of doing this is via a Copy ID to Clipboard type option, which would be part of the settings cog available on pages/menus, articles, blocks and components.  This makes sense as the _id is the ObjectID straight from MongoDB and not particularly readable from a usability perspective.  Persisting it on the clipboard might be enough, rather than cluttering the UI with such_id values.

brian-learningpool:
This commit adds a new option to the context menu for pages/menus, articles,
blocks and components.


lc-thomasberger:
+1

dennis-learningpool:
+1

dancgray:
+1

PR #996, opened 2016-01-20
Fixes #995, sets the JSON payload limit to 5Mb

### LINKED ISSUE #995: JSON payload size is limited to 100Kb
The `body-parser` seems to revert to its default because of the way it's invoked.  Thought the limit is indicated as 50Mb in code, this is not the case.  

I'm proposing a sensible limit of 5Mb.

brian-learningpool:
Previously this was defaulted to 100k.


dennis-learningpool:
+1

lc-thomasberger:
+1

Issue #999, opened 2016-01-21
Possible to drag component to the same position as an existing component


dancgray:
Paste zones remain enabled after adding a second component to a block.

Create a new block
Add a component to right
Add a component to left
Can still drag left or right component to the left, right or full position.

PR #1001, opened 2016-01-22
Issue/#870

dancgray:
Adds ability to move components within a block without drag and drop. Option only added to component item settings bar.

Also fixes issue #999.

### DIFF
--- a/lib/contentmanager.js
+++ b/lib/contentmanager.js
@@ -202,7 +202,7 @@ ContentPlugin.prototype.retrieve = function (search, options, next) {
       if (err) {
         return next(err);
       }
-      
+
       return next(null, records);
     });
   }, (options && options.tenantId));
@@ -229,7 +229,7 @@ ContentPlugin.prototype.update = function (search, delta, next) {
   if (!delta.hasOwnProperty('_id') && search.hasOwnProperty('_id')) {
     delta._id = search._id;
   }
-  
+
   self.hasPermission('update', user._id, user.tenant._id, delta, function (err, isAllowed) {
     if (!isAllowed) {
       return next(new ContentPermissionError());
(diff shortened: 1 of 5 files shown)


lib/contentmanager.js
```diff
@@ -1029,6 +1029,38 @@ ContentManager.prototype.setupRoutes = function () {
...
+
+    that.update(type, { _id: switchId }, switchVal, function (error, updatedDocument) {
+      if (error) {
+        res.statusCode = error instanceof ContentTypeError ? 400 : 500;
+        return res.json({ success: false, message: error.message });
+      }
+    });
+
+    res.statusCode = 200;
+    return res.json({ success: true });
```
dennis-learningpool: lines 1044 to 1060 could result in 3 responses to the client, should the updates fail. Also, it's possible that the first update fails, and the second would succeed, which would be undesirable, I think. 

I think that the update in lines 1052-1057 should be moved inside the first update callback, just after line 1049, and the return on 1059-1060 moved inside that second callback if there is no error.

Also, if there is an error on the second callback, the first update should be rolled back before returning the error.

dancgray:
Thanks @dennis-learningpool, corrected mistake. Added rollback, if that fails it's error is sent to logger.

dennis-learningpool:
Tested and works very nicely +1

dennis-learningpool:
Also, more fun to play with than is really reasonable

markl55:
Dennis gave me a quick preview of this earlier - really nice work Dan.

taylortom:
+1, really *really* cool! :tada: :balloon:

dancgray:
PR is to master, should be to develop. Will recreate for correct branch

PR #1005, opened 2016-01-25
Issue/#870

dancgray:
Adds ability to move components within a block without drag and drop. Option only added to component item settings bar.

Also fixes issue #999.


dennis-learningpool:
+!

brian-learningpool:
+1

taylortom:
+1

PR #1008, opened 2016-01-27
Fixes #994 (part one)

### LINKED ISSUE #994: Support for arrays of strings
Support for arrays of strings is patchy at the moment. There are a couple of blockers which require the most attention:

1. **Removing items**

 ~~You cannot remove a string from an array, even if a `confirmDelete` message is specified in the schema, likely due to this not being handled in [scaffold.js](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/core/scaffold/scaffold.js#L81-L86). I don't think a confirm message is needed for this trivial task anyhow.~~ See #1008.

2. **Default items**

 ~~You cannot specify a default array of strings. If a default array is specified this becomes a placeholder string for each item added to the array.~~ See #1052.

Schema:
```json
"colours": {
  "type": "array",
  "default": [
    "red",
    "blue"
  ],
  "title": "Favourite colours",
  "inputType": "List",
  "help": "List your favourite colours."
}
```
Output:

tomgreenfield:


brian-learningpool:
+1

taylortom:
+1

Might need to add brackets to those single-line ifs...@brian-learningpool?

PR #1014, opened 2016-01-28
Resolves #1011, removes non-essential properties from JSON output

### LINKED ISSUE #1011: Non-essential attributes are set in the output course JSON
In our effort to produce faster and smaller build package, we should should also ensure that the course JSON output does not contain any non-essential attributes which may have only been required on the server side.

brian-learningpool:
Added `omitProps` array to CourseCollection object -- currently only
set for 'course' and 'config' right now.

### DIFF
--- a/lib/outputmanager.js
+++ b/lib/outputmanager.js
@@ -31,27 +31,35 @@ var Constants = {
     CourseCollections : {
         'course': {
             tag: null,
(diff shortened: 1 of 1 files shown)


dennis-learningpool:
+1

taylortom:
+1

PR #1017, opened 2016-02-01
fix #1016 pm2 watch is too agressive

### LINKED ISSUE #1016: pm2 "watch" appears to kill and restart during preview
I've discovered and issue where the "watch" option in processes.json causes pm2 to restart the app on the vagrant box when doing a preview - apparently because the watch is picking up file changes from the build folder!

The "ignore_watch" attribute needs to be added to filter out those directories that we know will change, but shouldn't force a restart, namely "data" and "temp". There may be others.

dennis-learningpool:
See #1016

### DIFF
--- a/processes.json
+++ b/processes.json
@@ -1,8 +1,12 @@
 {
   "apps" : [{
-  "name"        : "adapt",
-  "script"      : "server.js",
-  "watch"       : true,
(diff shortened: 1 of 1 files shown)


brian-learningpool:
+1

dancgray:
+1

PR #1021, opened 2016-02-05
Resolves #1020, added caching to /asset/serve and /asset/thumb routes

### LINKED ISSUE #1020: Consider caching requests for assets
On reviewing the code there seems to be some code in `rest.js` which effectively disables any caching.  

As a performance enhancement we should revisit this for serving assets, as once an asset is uploaded, though attributes of it can be updated, the actual file itself cannot.

brian-learningpool:
This should speed up performance by reducing the number of requests
to the server.


dennis-learningpool:
+1

taylortom:
+1

PR #1023, opened 2016-02-05
Resolves #1022, reduces DB calls in writeCourseAssets()

### LINKED ISSUE #1022: Improve performance in OutputPlugin.prototype.writeCourseAssets()
I noticed some performance issues with the `writeCourseAsset()` function.  In an effort to provide faster build times as part of the 0.1.5 milestone I'm proposing that we reduce the number of database calls required to output the assets for a course.

brian-learningpool:
Using a combination of a 'distinct' on the 'courseassets' collection
and an '$in' operator on the 'assets' collection we can determine the
assets for a given course using only two DB calls.

Previously the code was getting the 'courseassets' (including any
duplicated assets) and making a new request to the database in order
to get the details of the asset.


dennis-learningpool:
+1

dancgray:
+1

PR #1026, opened 2016-02-09
Resolves #1025, removes default 'displayTitle' on adding a new block

### LINKED ISSUE #1025: Remove defaulting of displayTitle for blocks
Blocks are usually only used for structural purposes and defaulting the `displayTitle` to 'Block title', while consistent with other elements, is unnecessary as it's typically removed right away.  I'm proposing that we remove the defaulting of this.

brian-learningpool:


dennis-learningpool:
+1

dancgray:
+1

PR #1028, opened 2016-02-09
Resolves #1027, corrects upgrade.js

### LINKED ISSUE #1027: Bug in upgrade.js
Running `node upgrade.js` fails, on `app.polyglot.t()`.

This is because the helper function is accessed before `app` had initialised.

brian-learningpool:
Call to the polyglot helper function should be deferred until the server has actually started.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -364,8 +364,8 @@ Origin.prototype.createServer = function (options, cb) {
 Origin.prototype.startServer = function (options) {
 
   // Ensure that the options object is set.
-  options = typeof options === 'undefined' 
-    ? { skipVersionCheck: false } 
+  options = typeof options === 'undefined'
+    ? { skipVersionCheck: false, skipStartLog: false}
     : options;
 
   // configure server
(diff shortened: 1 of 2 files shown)


dennis-learningpool:
+1

dancgray:
+1
In terminal the "info: [09 Feb 2016 15:21:47 +00:00] Server started listening on port 5000" log gets inserted immediately after the "Would you like to continue?" prompt. 

Does not affect anything but I guess it might confuse some people.

taylortom:
+1, also to @dancgray's comment.

There's quite a block of output when you first run the script, and it took me a few seconds to realise what I was supposed to do. Can we add a line or two of whitespace before the upgrade output?

brian-learningpool:
I agree with @dancgray's comment.  I've noticed that a `skipVersionCheck` flag is passed in.  Do we have a similar flag to disable the logging during bootstrapping?

dancgray:
@brian-learningpool cannot find one. I'm assuming we would not want to suppress all logging so will apply just to startServer. Shall i commit to issue/1027 branch?

brian-learningpool:
Hi @dancgray, yeah, just commit on to his branch.

brian-learningpool:
+1

PR #1031, opened 2016-02-10
Resolves #1029, corrects sequencing issue on output plugin

### LINKED ISSUE #1029: Sequencing issue on output plugin
With the re-work to the Gruntfile in the framework, and the extra steps in the authoring tool to generate the `build`property there is a slight issue which can fail to apply the expected theme.

brian-learningpool:
This is to resolve issues where a theme wih customisations may have been used.


dennis-learningpool:
+1

PR #1034, opened 2016-02-11
Resolves #1033, corrects dependency issue with hot graphic

### LINKED ISSUE #1033: JavaScript error on courses published with a hot graphic but no narrative
This is due to the fact that on a small viewport a hot graphic component becomes a narrative.  Unfortunately the hot graphic has a dependency on narrative and must be bundled along with it.

brian-learningpool:


bell87:
+1

taylortom:
+1, thought that check seemed off at first :)

brian-learningpool:
Yeah, it still worked, but I couldn't let it be.

PR #1055, opened 2016-02-19
Fixes #1054, looks up on 'component' attribute rather than '_id'

### LINKED ISSUE #1054: Intermittent error in copy and pasting component
I've noticed this is an issue with older content, for example where the component plugin has since been updated.

brian-learningpool:
The 'component' value is constant, whereas the '_id' changes when newer versions are installed.  This had caused problems with Copy and Paste functionality.


dennis-learningpool:
+1

taylortom:
+1

PR #1058, opened 2016-02-25
Added implementation of CopyAsset() for localfs

brian-learningpool:
This helper function means we don't have to use read/write streams.

### DIFF
--- a/plugins/filestorage/localfs/index.js
+++ b/plugins/filestorage/localfs/index.js
@@ -13,6 +13,7 @@ var FileStorage = require('../../../lib/filestorage').FileStorage,
     async = require('async'),
     probe = require('node-ffprobe'),
     logger = require('../../../lib/logger'),
+    ncp = require('ncp').ncp,
     FFMpeg = require('fluent-ffmpeg');
 
 function LocalFileStorage() {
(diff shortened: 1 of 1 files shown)


dennis-learningpool:
+1

plugins/filestorage/localfs/index.js
```diff
@@ -290,6 +291,57 @@ LocalFileStorage.prototype.getFileStats = function (filePath, callback) {
...
 /**
+ * Copies an asset from one tenant to another.
+ * 
+ * @param {object} asset - A valid 'asset' record
+ * @param {string} sourceTenantName - Source tenant.name
+ * @param {string} destinationTenantName - Destination tenant.name
+ */
+LocalFileStorage.prototype.copyAsset = function(asset, sourceTenantName, destinationTenantName, callback) {
+  // Verify the destination folder exists
+  var assetFolder = path.join(configuration.serverRoot, configuration.getConfig('dataRoot'), destinationTenantName, asset.directory);
```
ryangadams: You could make this a bit more readable (with shorter lines) by declaring
`var dataRoot = path.join(configuration.serverRoot, configuration.getConfig('dataRoot');`

There are 5 instances it could be used.

plugins/filestorage/localfs/index.js
```diff
@@ -290,6 +291,57 @@ LocalFileStorage.prototype.getFileStats = function (filePath, callback) {
...
+      return callback(error);
+    }
+
+    // Copy the asset
+    var sourceFile = path.join(configuration.serverRoot, configuration.getConfig('dataRoot'), sourceTenantName, asset.path);
+    var destinationFile = path.join(configuration.serverRoot, configuration.getConfig('dataRoot'), destinationTenantName, asset.path);
+
+    ncp(sourceFile, destinationFile, {clobber:false}, function(error) {
+      if (error) {
+        logger.log('error', 'ncp error while copying ' + asset.filename);
```
ryangadams: should we also log the error message here (or is it logged elsewhere)? it could be file exists, or a file permissions problem or something else like disk space. I know from painful experience that “couldn't save $filename" can be a frustrating error message unless you know why.

plugins/filestorage/localfs/index.js
```diff
@@ -290,6 +291,57 @@ LocalFileStorage.prototype.getFileStats = function (filePath, callback) {
...
+
+        var sourceThumbnailFile = path.join(configuration.serverRoot, configuration.getConfig('dataRoot'), sourceTenantName, asset.thumbnailPath);
+        var destinationThumbnailFile = path.join(configuration.serverRoot, configuration.getConfig('dataRoot'), destinationTenantName, asset.thumbnailPath);
+
+        ncp(sourceThumbnailFile, destinationThumbnailFile, function(error) {
+          if (error) {
+            logger.log('error', 'ncp error while copying ' + asset.thumbnailPath);
+            logger.log('error', 'error copying (thumbnail) ' + sourceThumbnailFile + ' to ' + destinationThumbnailFile);
+
+            // TODO If thumbnails fail it's not the end of the world -- for now
```
ryangadams: would it be reasonable to call [`createThumbnail()`](https://github.com/adaptlearning/adapt_authoring/pull/1058/files#diff-75d6b374d5cc336322b6494e003f6e1eR353) here if the copy of the thumb fails?

plugins/filestorage/localfs/index.js
brian-learningpool: Good point.  I've updated the code.

plugins/filestorage/localfs/index.js
brian-learningpool: As an enhancement, maybe.

ryangadams:
+1

PR #1059, opened 2016-02-25
User management

taylortom:
#### Features
- Can add users
- Can edit existing users (click on user in the list to display edit controls)
  - Can disable/restore users (uses `_isDeleted`)
  - Can delete users (completely remove from DB)
  - Can edit user information
    - Email
    - Password
      - Change directly in a Notify popup
      - Send out password reset email
    - Tenant
    - Role
    - Reset `failedLoginCount` to 0/unlock locked accounts
- Password reset emails have been fixed, and styled to match the UI refresh colouring (requires https://github.com/adaptlearning/adapt_authoring/pull/1352)
#### Fixes
- #1220 
- #1332 
- #1346 (requires https://github.com/adaptlearning/adapt_authoring/pull/1371)  
- #1351  
- #1366 
- #1367 

**Notes:**<br>
- Feature only accessible to users with super admin rights
- Users are only able to change password/reset `failedLoginCount` on their own accounts; all other options are hidden.
#### Implementation notes
- **Everything's self-contained in the plugin folder, with the exception of a few amends to usermanager.** -- should this be moved to core for the release?
- **Users are only able to have one role/tenant at a time.**
- Uses backbone models' save function to persist any changes to the DB (with a couple of exceptions: changing role and password)


taylortom:
screen shot 2016-06-14 at 10 09 16

dancgray:
@taylortom looks and works really well. 
Just found a couple of issues (tested in chrome 54 OSX 10.11.6):
- I was able to edit a users email address to that of an existing user. Causes mongo duplicate key error.
- the locked user class was not added after a user exceeded login attempts and refresh button clicked.
- similarly the locked class was not removed after unlocking

It would be good if the refresh button went back to the off state once the view was refreshed.
At some point in the future it would be good to add filter/search features.
Otherwise much awesomeness.

taylortom:
Thanks for that @dancgray. I think the refresh button's a bit broken by the looks, will sort that asap.

There are a couple of other things I've been thinking about with this:
- The ability to bulk add users (using a box similar to tags to list the email addresses), which then  sends out an email requesting a user to set a password.
- User creation screen removes the password field, and just sends out an email as per above.

We would obviously need to fall back to password fields if STMP isn't enabled (and maybe disable the bulk add).

dancgray:
+1 for those suggestions. I guess the questions are: does it make user import easier? and is it more secure?

Sending a set password email is certainly more secure as it prevents the super user setting a generic password for all users they are importing. Which is probably what most people do.

dancgray:
Noticed that running `grunt build:prod`, or when installing, the requirejs:compile task gives an error:

Running "requirejs:compile" (requirejs) task
Error: Cannot uglify2 file: /var/www/deltanet.adaptbuilder/frontend/src/adaptbuilder/js/origin.js. Skipping it. Error is:
Unexpected token: name (key) (line: 30394, col: 25, pos: 1161500)

Running `grunt build:dev` works OK as it misses uglify

taylortom:
@dancgray thanks, I've just seen this too -- sorting it now 😄

dancgray:
Looking awesome,  just getting the following error after clicking save button on change password.

POST http://localhost:5000/api/user/resetpassword net::ERR_EMPTY_RESPONSE

lc-thomasberger:
Looks very good.

As a Tenant-Admin I was able to manage the Super-Admin. Is this by design, or will this change with tenant-management.

Can we update the user-screen to display the current role and or permissions?

simondate:
When I try to change the password of my user or another user the save button is not responsive (No errors in console) after about a minute this message appears 
screen shot 2016-10-24 at 15 21 05

taylortom:
@simondate that's what @dancgray reported

taylortom:
@dancgray, @simondate have fixed the above issue.
@lc-thomasberger like your suggestion about adding role information to the profiles screen, don't have time to look at this today though unfortunately.

lc-thomasberger:
@taylortom awesome, I can help with this. It's probably just a update to the view?

routes/lang/en-application.json
```diff
@@ -15,9 +16,29 @@
   "app.returninguser": "Returning User",
   "app.createnewaccount": "Create a new account - it's free and the button doesn't work",
   "app.createnewaccountbutton": "Create an account",
-  "app.alreadyloggedin": "You are logged in, please go to the ",
+  "app.alreadyloggedin": "Looks like you're are already logged in! Click the button below to go to your dashboard. ",
```
dancgray: Should be "Looks like you are already" or remove extra are.

taylortom:
Thanks @dancgray, fixed that.

I've submitted a new PR for a 'bulk user add' feature. Not sure if this is something we want to get in with this release. 

See here for more info: https://github.com/adaptlearning/adapt_authoring/pull/1444

aniketdharia:
👍  This seems good to go! :)

taylortom:
@aniketdharia I think we probably need to look into considerably reworking the build process; there are various things that could do with being refined. Are you ok with this being approached as a separate task?

Alternative could be to move the code into core.

aniketdharia:
@taylortom works for me... we can go ahead with this right now... 👍

PR #1061, opened 2016-02-26
Resolves #1060, removes the hasPermissions() check from retrieveAsset()

### LINKED ISSUE #1060: Performance issue with retrieveAsset()
The `retrieveAsset()` function does not require checking read permissions on every asset returned.  It's a given that the Course Creator role already has access to read all these assets.

brian-learningpool:
This check is not required as the Course Creator role has access to read all assets.

### DIFF
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -211,29 +211,7 @@ exports = module.exports = {
           return next(error);
         }
 
-        var readable = [];
-
-        // we have to iterate records, :/
-        if (records && records.length) {
-          async.eachSeries(
-            records,
-            function (rec, nextRec) {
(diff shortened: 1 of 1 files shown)


dennis-learningpool:
+1

bell87:
+1

taylortom:
Is there any need for a role (or the ability to create a role) without read access to assets (can't really think of any useful real-world examples, but thought I'd ask the question).

brian-learningpool:
No, it's a fair point.  There is the 'Authenticated User' role which is pretty redundant now.  

I think roles and permissions are probably due a revisit.

taylortom:
Fair enough. 
I can't think of a decent enough reason to lock down this function, so +1 (with the variables at the top of the function removed :smile:)

ryangadams:
+1

PR #1079, opened 2016-03-03
Removes duplicate DELETE event requests for blocks and components

brian-learningpool:
There was some code here which listened twice to deletes.


dennis-learningpool:
+1

taylortom:
+1

PR #1080, opened 2016-03-03
Fixes export to only allow one zip on the server per user

### LINKED ISSUE #1056: Need to delete old exports
At the moment, these are all stored indefinitely. 

Simple solution is to only keep one export per course. Any other ideas?

taylortom:
This is then overwritten with each export.

Fixes #1056.


taylortom:
Maybe need to revaluate this in conjunction with https://github.com/adaptlearning/adapt_authoring/issues/1091

taylortom:
@dancgray could you have a look at this? I think it's still a valid issue.

dancgray:
@taylortom this is still a valid issue. Single zip per user should be fine, it would only be a problem if users share login details and export source for different courses at the same time. It will reduce disk space compared with current setup.

routes/export/index.js
```diff
@@ -41,58 +41,63 @@ server.get('/export/:tenant/:course', function (req, res, next) {
...
-      res.statusCode = 401;
-      return res.json({
+      return res.status(401).json({
         success: false,
         message: app.polyglot.t('app.errorusernoaccess')
       });
     }
   });
 });
 // TODO probably needs to be moved to download route
```
canstudios-louisem: This is a good time to address this TODO.

Issue #1087, opened 2016-03-07
preview fails running vagrant on windows


lc-thomasberger:
Error in console:

adapt-0 (err): error: [07 Mar 2016 11:19:15 +00:00] Error: EACCES: permission denied, mkdir '/home/vagrant/temp/56dd62fe69beb9c538c4dff1/adapt_framework/courses'
adapt-0 (err): error: [07 Mar 2016 11:19:15 +00:00] Error: EACCES: permission denied, mkdir '/home/vagrant/temp/56dd62fe69beb9c538c4dff1/adapt_framework/courses'
adapt-0 (err): error: [07 Mar 2016 11:21:06 +00:00] Error: EACCES: permission denied, mkdir '/home/vagrant/temp/56dd62fe69beb9c538c4dff1/adapt_framework/courses'
adapt-0 (err): error: [07 Mar 2016 11:21:06 +00:00] Error: EACCES: permission denied, mkdir '/home/vagrant/temp/56dd62fe69beb9c538c4dff1/adapt_framework/courses'

brian-learningpool:
@lc-thomasberger, does the account have write permission on that folder?

lc-thomasberger:
ran cmd as admin
files located at: C:\vagrant\

Issue #1085, opened 2016-03-07
vagrant setup on windows processes.json not found


lc-thomasberger:
PM2 throws an error due to missing processes.json

processes.json is not copied into the VB on windows so it can't be found
https://github.com/adaptlearning/adapt_authoring/blob/master/vagrant_setup_win.sh#L9

PR #1092, opened 2016-03-09
vagrant windows fix

lc-thomasberger:
solves: 
issue #1087
issue #1085 

have tested on 2 windows 7 machines

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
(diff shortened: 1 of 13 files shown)


brian-learningpool:
+1, thanks!

brian-learningpool:
Tested on OSX Yosemite.  @lc-thomasberger, it might be worth merging the 32-bit branch into this PR too?

lc-thomasberger:
Thanks for testing on OSX @brian-learningpool. I think we should stick with the 64-bit version for now and update the documentation that vt-x must be enabled on windows.

dancgray:
+1  tested on default install of windows 8 64-bit

PR #1093, opened 2016-03-09
Resolves #1090, displays alert on delete project error

### LINKED ISSUE #1090: Notify.alerts in callbacks don't always fire
This may only occur when testing locally, but if a server response fires too quickly, any nested alerts in the callback of another alert won't fire (think the sweetalert may be calling the global .close function which is stopping it)

Example: dashboard delete button on shared project.

brian-learningpool:
This was caused by two sweet alert dialog boxes in quick succession.  Adding a delay seems to resolve this.


taylortom:
+1

PR #1096, opened 2016-03-11
Fixes #1095

### LINKED ISSUE #1095: Checkbox value always reverts to truthy schema default
If the schema default is truthy for a checkbox, it will always be checked on form revisit even if the user has previously unchecked it.

tomgreenfield:
With the override removed, the original [function](https://github.com/adaptlearning/adapt_authoring/blob/develop/frontend/src/core/libraries/backbone-forms.js#L1554-L1560) is called.

Initial population of schema defaults is handled [here](https://github.com/adaptlearning/adapt_authoring/blob/develop/lib/helpers.js#L31).

### DIFF
--- a/frontend/src/core/scaffold/scaffoldOverrides.js
+++ b/frontend/src/core/scaffold/scaffoldOverrides.js
@@ -99,14 +99,6 @@ define(function(require) {
 		this.$el.val(value);
 	}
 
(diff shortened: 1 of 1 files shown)


brian-learningpool:
+1

taylortom:
+1

PR #1098, opened 2016-03-11
Resolves #1097, added full support for download on other output plugin

### LINKED ISSUE #1097: Support pluggable 'download' option in UI and tidy up preview/download code
The 'download' functionality was not revisited the way that 'preview' was to support different output plugins.  Some other things:
- We're missing a permissions check on the download.
- An error during preview or download could result in a 'stuck' state in the UI.
- The `buildFlagExists()` function actually checks for the .rebuild flag then deletes it.  We need a distinct `clearBuildFlag()` in cases where the build errors before it finishes.  (Also this code is using `fs.exists()`, which has been deprecated.)

brian-learningpool:
- Fixed buildFlagExists()
- Added clearBuildFlag()
- Removed deprecated fs.exists()
- Added missing permissions check to download route


dancgray:
+1

dennisheaney:
+1

PR #1103, opened 2016-03-15
Issue/1102

lc-thomasberger:
run node upgrade after node install


brian-learningpool:
+1

johnniezen:
That is what I missed till now, could not do it manually somehow. If it helps, a +1 from me.

PR #1123, opened 2016-03-24
Added 'login:loaded' event

brian-learningpool:
This allows plugins to tap into the login page, in the same way as the dashboard.

### DIFF
--- a/frontend/src/core/user/templates/login.hbs
+++ b/frontend/src/core/user/templates/login.hbs
@@ -12,7 +12,7 @@
         <div class="login-new-user col-1-2">
 
             <div class="login-new-user-inner">
-                  <h2 class="welcome-blurb">Welcome to the {{t 'app.productname'}}.</h2v>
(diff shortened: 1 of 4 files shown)


dennis-learningpool:
+1

frontend/src/core/user/templates/login.hbs
```diff
@@ -12,22 +12,13 @@
         <div class="login-new-user col-1-2">
 
             <div class="login-new-user-inner">
-                  Welcome to the {{t 'app.productname'}}.<br/><br/>
+                  <h2 class="welcome-blurb">Welcome to the {{t 'app.productname'}}.</h2v>
```
dancgray: Should be </h2>

brian-learningpool: Good catch -- missed that completely.

dancgray:
+1

PR #1131, opened 2016-04-11
Fix issue with install script grabbing versions of plugins which aren't compatible with the framework core

### LINKED ISSUE #1110: Install should not exit if plugin framework supported version is greater than the installed framework version.
Example:

Authoring tool installs framework v2.0.1
latest version of Trickle requires framework v2.0.6
the install fails with output something like
```Unable to install adapt-contrib-trickle as it is not supported in the current version of of the Adapt framework ERROR: Unable to install adapt-contrib-trickle as it is not supported in the current version of of the Adapt framework Install was unsuccessful. Please check the console output.```

Suggest that this should just log an error and go to the next core plugin in adapt.json.

taylortom:
Fixes #1110 

Queries the bower repo to find the latest version that's compatible with the installed framework core. Any plugins with semver `*` in `adapt.json` will use the new `Bowermanager.installLatestCompatibleVersion` function, whereas specific versions use `Bowermanager.installPlugin` as before.

_**Testing tip**: add a `next();` at line 208 to skip the framework install. Depending on your internet connection, this can take a while._

### DIFF
--- a/install.js
+++ b/install.js
@@ -352,22 +352,12 @@ var steps = [
       var plugins = Object.keys(json.dependencies);
 
       async.eachSeries(plugins, function(plugin, pluginCallback) {
-        app.bowermanager.installPlugin(plugin, json.dependencies[plugin], function(err) {
-          if (err) {
-            return pluginCallback(err);
-          }
-
-          pluginCallback();
-        });
-
-      }, function(err) {
-        if (err) {
-          console.log(err);
-          return next(err);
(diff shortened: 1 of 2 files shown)


taylortom:
Testing tip: add a `next();` at line 208 to skip the framework install. Depending on your internet connection, this can take a while.

dancgray:
+1 works for me

brian-learningpool:
+1

PR #1133, opened 2016-04-13
Resolves #1129, adds <legend> to array items

### LINKED ISSUE #1129: Extend legend styles to arrays
I have an extension that is based on an array not an object. I have not found a way in _properties.schema_ to trigger UI formatting that is applied to "legend". Doesn't matter whether I use "legend" or "title", the current presentation is awkward. (My properties.schema: https://github.com/chucklorenz/adapt-favicons/blob/master/properties.schema)
at-before

I'm suggesting a way be found to apply the same styling of "legend" for objects to arrays. I was able to produce what I am looking for by adding this code to line 176 of _frontend/src/core/scaffold/scaffold.js_:

```
} else if (field.type === 'array') {  
      scaffoldSchema[key].title = '';  
      scaffoldSchema[key].legend = field.legend;  
}
```

at-after

I'm not submitting a PR because my knowledge of the application is not strong and because I suspect there will need to be an adjustment to CSS padding, etc.

brian-learningpool:
Thanks to @chucklorenz for raising this.

### DIFF
--- a/frontend/src/core/scaffold/scaffold.js
+++ b/frontend/src/core/scaffold/scaffold.js
@@ -170,7 +170,7 @@ define(function(require) {
 		if (field.title) {
 			scaffoldSchema[key].title = field.title;
 			
-		} else if (field.type === 'object') {
+		} else if (field.type === 'object' || field.type === 'array') {
 			scaffoldSchema[key].title = '';
 			scaffoldSchema[key].legend = field.legend;
 		}


brian-learningpool:
This is purely functional.  Styling, etc. will be revisited as part of the 0.2 UI/UX refresh where I suspect `scaffold.js` will change substantially.

dancgray:
+1

lc-thomasberger:
+1

chucklorenz:
+1 thanks Brian

PR #1140, opened 2016-04-22
Properties not appearing on-screen in same order as properties.schema

barry-learningpool:
Resolved the source of the issue with inverted properties. Turns out it is a quirk in MongoDB. The property 'retainKeyOrder' needed to be set on the schema options 'toObject' attribute (see here: http://mongoosejs.com/docs/api.html#document_Document-toObject). Properties will now be listed in the same order as defined in the schema file. For this to take effect on a component, a version bump is required.


dennis-learningpool:
+1 - Nice find!

dancgray:
+1 awesome find

ryangadams:
+1

ryan-lynch:
+1

brian-learningpool:
+1, great job!

PR #1143, opened 2016-04-22
Issue-1141 - Fixed typo which caused db.update on role to lose the statement.

ryan-lynch:


dennis-learningpool:
+1 - wonder which silly ned is responsible for that bug ... *cough*

barry-learningpool:
+1

brian-learningpool:
+1, nice find!

PR #1148, opened 2016-04-25
Resolves issue 1147

ryan-lynch:
Removed a hack that checks for @learningpool.com e-mail address to give permissions to resetPassword and replaced it with a call to hasPermission. Also removed resetPassword endpoint from the ignoreRoutes array.


lib/usermanager.js
```diff
@@ -751,17 +766,20 @@ exports = module.exports = {
         return res.json({success: false, message: 'request body was not a valid object'});
       }
 
-      // TODO - Verify the current user has rights
-      // HACK -- Allowing @learningpool.com email addresses for now
-      if (user.email.indexOf('@learningpool.com') > -1) {
+      self.hasPermission('update', user, delta, function(err, isAllowed) {
+        if (err) {
```
brian-learningpool: What if `isAllowed` is `false`?

lib/usermanager.js
```diff
@@ -53,6 +52,22 @@ exports = module.exports = {
...
+   * @param {objectid} userId
+   * @param {objectid} tenantId
+   * @param {object} a content item
+   * @param {callback} next (function (err, isAllowed))
+   */
+
+  hasPermission: function(action, user, contentItem, next) {
+    permissions = require('./permissions');
+    var resource = permissions.buildResourceString(user.tenant._id, '/api/user/resetpassword/' + contentItem);
+    permissions.hasPermission(user._id, action, resource, next);
```
brian-learningpool: I don't think the `contentItem` parameter is correct here if you're only looking to restrict access to the `/api/user/resetpassword` endpoint.

lib/usermanager.js
```diff
@@ -53,6 +52,22 @@ exports = module.exports = {
...
+   *
+   * @param {string} action
+   * @param {objectid} userId
+   * @param {objectid} tenantId
+   * @param {object} a content item
+   * @param {callback} next (function (err, isAllowed))
+   */
+
+  hasPermission: function(action, user, contentItem, next) {
+    permissions = require('./permissions');
```
brian-learningpool: Hmmm.

ryan-lynch: Without a rather large refactor, I think this is the only way to resolve the problem right now.

lib/usermanager.js
ryan-lynch: That's true, I don't think contentItem is necessary here.

lib/usermanager.js
ryan-lynch: Woops, I'll handle this

lib/usermanager.js
```diff
@@ -53,6 +52,22 @@ exports = module.exports = {
     'DuplicateUserError': DuplicateUserError
   },
 
+  /**
+   * checks if a user is permitted the action on the resource
+   *
+   * @param {string} action
```
ryangadams: this jsdoc needs updated to match the function details

lib/usermanager.js
```diff
@@ -53,6 +52,20 @@ exports = module.exports = {
...
+   * @param {object} user
+   * @param {callback} next (function (err, isAllowed))
+   */
+
+  hasPermission: function(action, user, next) {
+    permissions = require('./permissions');
+    var resource = permissions.buildResourceString(user.tenant._id, '/api/user/resetpassword');
```
ryangadams: Now that I look at this shouldn't this resource be injected into hasPermission? 

Something like 
```
hasPermission: function(action, user, resource, next) {
  permissions = require('./permissions');
  var resource = permissions.buildResourceString(user.tenant._id, resource);
// etc.
```
Then on line 767
```
self.hasPermission('update', user, '/api/user/resetpassword', function(err, isAllowed) {
   // do stuff
}
```

lib/usermanager.js
```diff
@@ -53,6 +52,21 @@ exports = module.exports = {
...
+   * checks if a user is permitted the action on the resource
+   *
+   * @param {string} action
+   * @param {object} user
+   * @param {string} resource
+   * @param {callback} next (function (err, isAllowed))
+   */
+
+  hasPermission: function(action, user, resource, next) {
+    permissions = require('./permissions');
```
brian-learningpool: You shouldn't do this:
https://github.com/alanjames1987/Node.js-Best-Practices#never-require-modules-inside-of-functions

brian-learningpool: Remove the circular dependency by removing the `require('./usermanager') and replacing calls to `usermanager` with `app.usermanager`.

brian-learningpool:
+1

ryangadams:
+1

PR #1203, opened 2016-05-17
updated node.js version

hrajotia:
- added Node.js version `4`, `5` and `6`.
- logging `git`, `mongodb` and `adapt` versions.
- removed caching for `node_modules` directory.
- removed `on_start` from `notifications` as its showing deprecated on [Travis WebLint](https://lint.travis-ci.org/).

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
   },
   "main": "index",
   "engines": {
-    "node": ">=0.10"
+    "node": ">=4.0.0"
   },
   "scripts": {
     "test": "make test"
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
   },
   "main": "index",
   "engines": {
-    "node": ">=4.0.0"
+    "node": "4.x"
   },
   "scripts": {
     "test": "make test"
--- a/plugins/content/config/model.schema
+++ b/plugins/content/config/model.schema
@@ -125,7 +125,7 @@
           "type": "string",
           "required": true,
           "default": "easeOutQuart",
-          "title": "Hide Easing",
+          "title": "Show Easing",
           "inputType": {
             "type": "Select",
             "options": [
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -128,7 +128,7 @@
   },
   "main": "index",
   "engines": {
-    "node": ">=0.10"
+    "node": ">=4.0.0"
```
brian-learningpool: Should this be >= or `4.x`?

hrajotia: Travis seems happy with [Node.js v6.1](https://travis-ci.org/adaptlearning/adapt_authoring/jobs/130845301#L280). So we can also make `"node": ">=4.0.0 <6.x"`.

What do you think?

taylortom:
I'd be tempted to stick with specifying the LTS/@brian-learningpool's suggestion just to be safe.

hrajotia:
@brian-learningpool, @taylortom done!

taylortom:
This all seems ok to me, @brian-learningpool?

plugins/content/config/model.schema
```diff
@@ -125,7 +125,7 @@
           "type": "string",
           "required": true,
           "default": "easeOutQuart",
-          "title": "Hide Easing",
```
taylortom: Should these be the other way round? We have easeOutQuart for the '_showEasing' animation.

hrajotia: yes we have `easeOutQuart` for `_showEasing` and `easeInQuart` for `_hideEasing` [see config.json](https://github.com/adaptlearning/adapt_framework/blob/master/src/course/config.json#L11-L12). but mistakenly both titles was labeled as `Hide Easing` that's why I have corrected it!

brian-learningpool:
+1

PR #1216, opened 2016-06-17
Resolves #1215, corrects a caching issue with course schemas

### LINKED ISSUE #1215: Extensions going missing from a project
This was reported on the community site:
https://community.adaptlearning.org/mod/forum/discuss.php?d=1442

I tracked the issue down to what looks like some kind of caching problem. To re-create: 

1. Create two dummy courses, Course A and Course B
2. When editing Course A, add two extensions, e.g. resources, and search.
3. When editing Course B, add three extensions, e.g. resources, spoor, and page level progress. The important point is to add a common extension. In this case I've used resources.
4. While on the Dashboard, repeatedly go out and in of Course A and Course B and view their Project Settings page.

**Expected result:**
Course A would show two extensions, Course B will show three.

**Actual result:**
The code gets confused and does not display all extensions. Depending on the order of your visit to each course you may only see the Resources extension.

brian-learningpool:
This could manifest itself when editing Project Settings whereby extensions or global text strings would apparently disappear from the course.

There were two fundamental problems with the previous code:
- The _.clone() function only provides a shallow-copied clone. Nested objects, such as '_extensions' are copied by reference.
- In scaffold.js, the builtSchemas variable which is being used as a cache for previously generated schemas, should not persist the 'course' schema becuase it is dependent on which plugins are enabled or in use on a course.

### DIFF
--- a/frontend/src/core/scaffold/scaffold.js
+++ b/frontend/src/core/scaffold/scaffold.js
@@ -191,7 +191,11 @@ define(function(require) {
         
       });
       
-      builtSchemas[type] = scaffoldSchema;
+      b// The 'course' schema is a special case as it is augmented depending
+      // on what plugins are enabled on the course.
+      if (type !== 'course') {
+        builtSchemas[type] = scaffoldSchema;
+      }
       
       return scaffoldSchema;  
     } catch (e) {
--- a/frontend/src/core/scaffold/scaffold.js
+++ b/frontend/src/core/scaffold/scaffold.js
@@ -178,29 +178,30 @@ define(function(require) {
 
 	var buildSchema = function(schema, options, type) {
 
-    try {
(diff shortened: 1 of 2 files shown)


taylortom:
Untested, but +1 in principle

dancgray:
+1, could reproduce issue. PR fixes issue

taylortom:
I'm still getting the issue atm.
In my case, I've got one course with resources and spoor, another with resources and dev tools. I've got the situation now whereby both are showing settings for resources and spoor.
Both courses were created prior to me testing this fix (to replicate the issue beforehand)

brian-learningpool:
`config` and the rest.  I've updated the code.

taylortom:
+1, works for me!

PR #1219, opened 2016-06-21
Fixes form validation

### LINKED ISSUE #1209: Form validation broken
None of the form validation is working at present, which leads to the very un-useful error message:
> Something went wrong while saving your data. Please try again.

taylortom:
Fixes #1209


taylortom:
Shouldn't be going into master

PR #1281, opened 2016-07-28
Ui refresh - Button widths

simondate:
Applied fixed widths to two buttons to make them not break onto two lines.


simondate:
These issues only seems to occur on Chrome for Windows.

PR #1297, opened 2016-08-08
Resolves #1289, adds support for negative numbers in 'number' validation

### LINKED ISSUE #1289: Validation error with Assessment extension preventing save
**Steps to re-create**
1. Add the assessment extension to a course.
2. Add a new page, change some fields, set the Assessment ID and click Save.

**Expected result**
The page would save.

**Actual result**
The page flags an error in the Assessment extension that Attempts Allowed must be a number.  It's set to -1 (the default).

brian-learningpool:
This has been patched in backbone-forms.js directly.  We should consider upgrading the entire library in a future release.


taylortom:
+1

dancgray:
+1

PR #1335, opened 2016-08-18
Fixed issue with uploading plugins

davetaz:
Tested on Ubuntu. 

There is a bug that it looks for files in the wrong place (or the guide on how to build plugins is wrong).


davetaz:
No longer relevant

PR #1347, opened 2016-08-23
Proposed: move function declarations out of helpers object for readability

taylortom:
More like lib/helpers.js. Gives an 'index' of what's in the object without having to search.

(have separated this from the bug fix as is more a suggestion)


taylortom:
Defunct.

PR #1381, opened 2016-09-13
pm2 processes.json fix

lc-thomasberger:
when uploading plugins, pm2 restarts the app while uploading the plugin. this causes the plugin upload to fail

removed "plugins" folder frow watch list to fix this issue


brian-learningpool:
Good catch @lc-thomasberger!

+1

dancgray:
+1

PR #1401, opened 2016-09-23
Update to page editor screen

simondate:
Fix to 'Add block' button that is not fully responsive in width and does not align with the block above.


simondate:
screen shot 2017-01-04 at 12 53 51 Notice the add block button is not aligned centrally

PR #1404, opened 2016-09-23
Make asset description not required

### LINKED ISSUE #1120: An asset's description shouldn't be required

taylortom:
Fixes #1120

### DIFF
--- a/lib/dml/schema/system/asset.schema
+++ b/lib/dml/schema/system/asset.schema
@@ -14,7 +14,7 @@
     },
     "description": {
       "type": "string",
-      "required": "true"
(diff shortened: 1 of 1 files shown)


davetaz:
Hello all, Unfortunately this doesn't work. 

As noted in the README.md it states:

"Note that the schema defined here are not in the JSON Schema format - they are Mongoose Schemas"

As they are not JSON then you cannot have "required": "false" with the "false" in quotes. In quotes this evaluates to true. So it needs to be:

"required": true 
or
"required": false 

no quotes. 

This then works. I've closed this for assets in #2010

Issue #1408, opened 2016-09-25
Course -> Manage extensions should be alphabetised


taylortom:
List seems to be in a random order -- would be easier to navigate if this was alphabetised.

taylortom:
Fixed

PR #1409, opened 2016-09-25
Hook plugin._isAvailableInEditor to frontend

### LINKED ISSUE #922: Plugin management: enabled checkbox has no effect
I may have misunderstood what this does, but I can't see it doing anything useful at the moment.

I expected unchecking this to then hide that plugin from selection in the course edit tool.
- Seems to have no effect on extensions whatsoever
- Components appear disabled when in course edit view, but you can still add them as normal

taylortom:
Hides any plugins which have been disabled from:
- Manage extensions
- Menu picker
- Theme picker
- Component list view

Fixes #922 and #1408 

Doesn't do anything about disabling any used plugins when the _isAvailableInEditor is set to false, simply stops being added afterwards.


frontend/src/core/editor/extensions/views/editorExtensionsEditView.js
```diff
@@ -66,38 +55,56 @@ define(function(require) {
         this.setViewToReady();
     },
 
+    postData: function(url) {
+        var self = this;
+        $.post(url + this.model.get('_id'), { extensions: this.currentSelectedIds },
+            _.bind(function(result) {
+            if (result.success) {
+                self.refreshData();
```
taylortom: Yeah, good point.....

taylortom:
Fixed above, also a few whitespace amends

PR #1417, opened 2016-10-07
Strip tab and newline characters from WYSIWYG

### LINKED ISSUE #1416: Rich Text Editor content - accessibility issues
When I create content using the rich text editor it inserts lots of extra \n and \t characters (e.g. between paragraphs).

These don't affect rendering but they do seem to become individually tabbable if accessibility is turned on, which means tabbing between two paragraphs takes two tabs.

Manually removing the extra characters fixes the problem.

tomgreenfield:
Resolves #1416.


oliverfoster:
+1

PR #1428, opened 2016-10-12
Remove grandchild courseassets on content object delete

taylortom:
i.e. article & blocks


plugins/content/courseasset/index.js
```diff
@@ -42,38 +42,67 @@ CourseAssetContent.prototype.getModelName = function () {
...
               function(callback){
+                if (contentType !== 'contentobject') {
+                  return callback(null, 'Not processing a content object');
+                }
+                db.retrieve('article', { _courseId: itemForDeletion._courseId, _parentId: itemForDeletion._id }, function(err, articles) {
+                  if (err) {
+                    return callback(err, 'Unable to retrieve child articles of contentobject ' + itemForDeletion._id);
+                  }
+                  var parentIds = [];
+                  if (articles && articles.length !== 0) {
```
brian-learningpool: @taylortom, I think there's an issue here if you attempt to delete a page which has no articles.  It's an edge case, I know.  Can you add something like the following instead?
```
                  if (articles.length == 0) {
                    return callback(null, 'A contentobject with no articles is being removed');
                  } 

```

plugins/content/courseasset/index.js
```diff
@@ -42,38 +42,67 @@ CourseAssetContent.prototype.getModelName = function () {
...
+                      }
+                      if(parentIds.length > 0) {
+                        criteria.$or = [
+                          { _contentTypeParentId: { $in: parentIds } },
+                          { _contentTypeId: itemForDeletion._id }
+                        ];
+                      } else {
+                        criteria._contentTypeId = itemForDeletion._id;
+                      }
+                      callback(null, 'Child content objects added to criteria object');
```
brian-learningpool: `cb` is never called here.  I don't think this will work as expected for articles which have multiple blocks.

dancgray: Getting header errors with multiple articles or blocks possible due to this.
`Error: Can't set headers after they are sent.`

dancgray:
On courses with extension assets on contentObjects, articles or blocks this still leaves orphaned assets.

taylortom:
@dancgray have merged your PR here

PR #1437, opened 2016-10-27
Show roles

lc-thomasberger:
might link to a wiki page that describes each role

### DIFF
--- a/frontend/src/core/user/templates/userProfile.hbs
+++ b/frontend/src/core/user/templates/userProfile.hbs
@@ -10,6 +10,19 @@
     <div class="editing-overlay-panel-content">
         <div class="editing-overlay-panel-content-inner">
             <form class="forms">
+                
+                {{#if rolesAsName}}
+                <div class="form-group">
+                    <label>{{t 'app.role'}}
+                        <span id="titleErrorMessage" class="error"></span>
+                    </label>
+                    <dl>
+                    {{#each rolesAsName}}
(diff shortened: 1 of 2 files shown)


lib/usermanager.js
```diff
@@ -748,8 +761,28 @@ exports = module.exports = {
...
+
+            if (roles.length > 0) {
+              data.rolesAsName = [];
+
+              for (var i = 0; i < roles.length; i++) {
+                data.rolesAsName.push(roles[i].name);
+              }
+            } else {
+              data.rolesAsName = false;
+            }
```
tomgreenfield: The entire conditional logic on lines 772 & 778–780 is optional.

If there are no `roles` then `data.rolesAsName` would remain as `[]` which equates to falsy in the Handlebars.

lc-thomasberger: 👍

frontend/src/core/user/templates/userProfile.hbs
```diff
@@ -10,6 +10,19 @@
     <div class="editing-overlay-panel-content">
         <div class="editing-overlay-panel-content-inner">
             <form class="forms">
+                
+                {{#if rolesAsName}}
+                <div class="form-group">
+                    <label>{{t 'app.role'}}
+                        <span id="titleErrorMessage" class="error"></span>
```
taylortom: Is this `span` used?

PR #1455, opened 2016-11-24
Test new experience in content authoring 

hibaIT:


taylortom:
@hibaIT I'm assuming this was a mistake?

PR #1461, opened 2016-12-01
Delete assets hidden from user and super admin

### LINKED ISSUE #1450: Deleted assets can be added by Super Admin to a course

aniketdharia:
fixed #1450


taylortom:
I don't think we should be filtering out deleted asset globally, as this then prevents admins from restoring assets at a later date.

My suggestion is to only hide assets in the modal view.

ambadasavhad:
@taylortom the above comments have been implemented in the recent comment.

PR #1463, opened 2016-12-05
Issue/1411

### LINKED ISSUE #1411: Access rights
Only the super user should be able to add users in all tenants, and the tenant admin  should be able to see only the users and tenants to which he/she is an admin. The other users should not be able to view the user management link and should get access denied message.

aniketdharia:
fixed #1411


taylortom:
Lots of other commits are coming through here, could you resubmit with just the relevant ones please?

Issue #1480, opened 2017-01-13
implement onScreen support


lc-thomasberger:
Feature landed in the Framework. Needs to be added to the schemas for articles/blocks/components 
https://github.com/adaptlearning/adapt_framework/issues/1100

PR #1490, opened 2017-01-25
implement onScreen support

lc-thomasberger:
#1480

### DIFF
--- a/plugins/content/article/model.schema
+++ b/plugins/content/article/model.schema
@@ -52,6 +52,41 @@
       "title": "Is this available?",
       "help": "If set to 'false' this will not be displayed"
     },
+    "_onScreen": {
+      "type": "object",
+      "title": "",
+      "properties": {
+        "_isEnabled": {
+          "type": "bool",
+          "default": "false",
+          "isSetting": true,
+          "inputType": {"type": "Boolean", "options": [true, false]},
+          "validators": [],
+          "title": "Enable",
+          "help": "If set to 'true', a CSS-Class will be added once the Element is scrolled into view."
+        },
+        "_classes": {
+          "type": "string",
+          "default": "",
+          "isSetting": true,
+          "required": false,
+          "inputType": "Text",
+          "validators": [],
+          "title": "Classes",
+          "help": "The CSS-Class that is added to the Element."
+        },
+        "_percentInviewVertical": {
+          "type": "number",
+          "required": false,
+          "default": 50,
+          "isSetting": true,
+          "title": "Percent in view",
+          "inputType": "Number",
+          "validators": ["number"],
+          "help": "Percent of the Element in View."
+        }
+      }
+    },
     "_extensions": {
     	"type":"object"
     }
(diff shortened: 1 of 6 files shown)


plugins/content/article/model.schema
```diff
@@ -52,6 +52,41 @@
...
+      "title": "",
+      "properties": {
+        "_isEnabled": {
+          "type": "bool",
+          "default": "false",
+          "isSetting": true,
+          "inputType": {"type": "Boolean", "options": [true, false]},
+          "validators": [],
+          "title": "Enable",
+          "help": "If set to 'true', a CSS-Class will be added once the Element is scrolled into view."
```
moloko: I think this would be worded better as 'If enabled, the CSS class(es) specified in the Classes field will be added to the article element when it is within the browser's viewport'

plugins/content/article/model.schema
```diff
@@ -52,6 +52,41 @@
...
+        },
+        "_classes": {
+          "type": "string",
+          "default": "",
+          "isSetting": true,
+          "required": false,
+          "inputType": "Text",
+          "validators": [],
+          "title": "Classes",
+          "help": "The CSS-Class that is added to the Element."
```
moloko: I think this would be better worded as 'The CSS class(es) to add to the article element'

plugins/content/article/model.schema
```diff
@@ -52,6 +52,41 @@
...
+        },
+        "_percentInviewVertical": {
+          "type": "number",
+          "required": false,
+          "default": 50,
+          "isSetting": true,
+          "title": "Percent in view",
+          "inputType": "Number",
+          "validators": ["number"],
+          "help": "Percent of the Element in View."
```
moloko: I think this would be better worded as 'Controls what percentage of the article's height needs to be in the viewport in order for the CSS class(es) to be added'

plugins/content/block/model.schema
```diff
@@ -59,6 +59,41 @@
...
+      "title": "",
+      "properties": {
+        "_isEnabled": {
+          "type": "bool",
+          "default": "false",
+          "isSetting": true,
+          "inputType": {"type": "Boolean", "options": [true, false]},
+          "validators": [],
+          "title": "Enable",
+          "help": "If set to 'true', a CSS-Class will be added once the Element is scrolled into view."
```
moloko: see above comments only with 'block element' instead of 'article element'

plugins/content/block/model.schema
```diff
@@ -59,6 +59,41 @@
...
+        },
+        "_classes": {
+          "type": "string",
+          "default": "",
+          "isSetting": true,
+          "required": false,
+          "inputType": "Text",
+          "validators": [],
+          "title": "Classes",
+          "help": "The CSS-Class that is added to the Element."
```
moloko: see above comments only with 'block element' instead of 'article element'

plugins/content/block/model.schema
```diff
@@ -59,6 +59,41 @@
...
+        },
+        "_percentInviewVertical": {
+          "type": "number",
+          "required": false,
+          "default": 50,
+          "isSetting": true,
+          "title": "Percent in view",
+          "inputType": "Number",
+          "validators": ["number"],
+          "help": "Percent of the Element in View."
```
moloko: see above comments only with 'block's height' instead of 'article's height'

Issue #1447, opened 2016-11-08
Allow for custom SMTP servers


taylortom:
Currently we only support services listed here: https://nodemailer.com/2-0-0-beta/setup-smtp/well-known-services/

Easiest solution is to use a connection URL:
```
smtps://user%40gmail.com:pass@smtp.gmail.com/?pool=true
```

canstudios-louisem:
We are running into an issue with this. What was the outcome of the reference PR @lc-thomasberger @taylortom I see it was close without much discussion?

taylortom:
Looks like the fix was added to https://github.com/adaptlearning/adapt_authoring/pull/1615

PR #1491, opened 2017-01-25
custom smtp Service

lc-thomasberger:
refactored so that a custom connection URL (`smtps://user%40gmail.com:pass@smtp.gmail.com/?pool=true`) can be passed as a smtpService config option.

#1447

### DIFF
--- a/lib/mailer.js
+++ b/lib/mailer.js
@@ -22,14 +22,13 @@ var Mailer = function () {
     var errors = [];
 
     if (this.service === '' && !this.useSmtpUrl) {
+    	// neither valid service nor smtpUrl defined, fail imediately 
       errors.push('smtpService as a well-known Service or a smtpConnectionUrl is required');
-    } else {
-      if (!this.useSmtpUrl) {
-        if (this.service === '') errors.push('smtpService as a well-known Service');
-        if (this.user === '') errors.push('smtpUsername');
-        if (this.pass === '') errors.push('smtpPassword');
-        if (this.from === '') errors.push('fromAddress');
-      }
+    } else if (!this.useSmtpUrl) {
+    	// check if required options for servie are available 
+			if (this.user === '') errors.push('smtpUsername');
+			if (this.pass === '') errors.push('smtpPassword');
+			if (this.from === '') errors.push('fromAddress');
     }
 
     if (errors.length > 0) {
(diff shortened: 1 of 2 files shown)


taylortom:
Solution looks good to me!

My only slight concern is that we're using the `smtpService` for two completely different uses -- it may be more transparent if we just add a new `smtpConnectionUrl` or something similar.

dancgray:
+1 for a seperate smtpConnectionUrl

lib/mailer.js
```diff
@@ -13,30 +13,46 @@ var Mailer = function () {
...
 
-    if(this.service === '') errors.push('smtpService');
-    if(this.user === '') errors.push('smtpUsername');
-    if(this.pass === '') errors.push('smtpPassword');
-    if(this.from === '') errors.push('fromAddress');
+    if (this.service === '' && !this.useSmtpUrl) {
+      errors.push('smtpService as a well-known Service or a smtpConnectionUrl is required');
+    } else {
+      if (!this.useSmtpUrl) {
+        if (this.service === '') errors.push('smtpService as a well-known Service');
```
tomgreenfield: Line 24:
```js
if (this.service === '' && !this.useSmtpUrl) {
```
and lines 27 & 28:
```js
if (!this.useSmtpUrl) {
  if (this.service === '') errors.push('smtpService as a well-known Service');
```
is exactly the same logic?

lc-thomasberger: The first one checks if at least one of the two options is defined. If there is, it checks the required values for the well known service.

taylortom: @lc-thomasberger I don't see how the second check will ever execute, if both conditions to line 24 are satisfied, it won't ever go into the else.

lib/mailer.js
```diff
@@ -13,30 +13,46 @@ var Mailer = function () {
...
+    smtpConfig = {
+      service: this.service,
+      auth: {
+        user: this.user,
+        pass: this.pass
+      }
+    };
+  }
+
   // Configure the credentials for the specified sevice
```
taylortom: Can you change `sevice` -> `service`

lib/mailer.js
taylortom: I'm finding this block quite hard to read, could we do something like:
```
if(!this.useSmtpUrl) {
        if (this.service === '') errors.push('smtpService as a well-known Service');
        if (this.user === '') errors.push('smtpUsername');
        if (this.pass === '') errors.push('smtpPassword');
        if (this.from === '') errors.push('fromAddress');
    }
```

PR #1510, opened 2017-02-08
Bulk user add

### LINKED ISSUE #1391: Bulk add new users
The ability to bulk add users from a list of email addresses (using an input box similar to when adding tags), which then sends out an email requesting a user to set a password.

### LINKED ISSUE #1392: Remove password input from create user
Remove the password field from the user creation screen, and just send out an email to the user to request a password to be set.

taylortom:
Allows admins to quickly create multiple users (fixes #1392).

The big new feature which enables this is the ability to create a user without specifying a password. An email is sent out to the account requesting that a password be set (uses the password reset API). *Note that the backend generates a temporary password at the point of creation, but this is never used.*

screen shot 2016-11-07 at 12 58 29


taylortom:
Superseded by https://github.com/adaptlearning/adapt_authoring/pull/1615

Issue #1519, opened 2017-02-13
Manage Extensions - do we really need the are you sure? prompt?


hmaffin:
As it's extremely easy to remove a plugin that you just added in Manage Extensions by mistake and it seemingly has had no impact if you added it by mistake, is it acceptable to remove the prompt asking if you really want to add this extension? I just enabled a bunch of them of a new project and it got a bit tedious!!

By all means keep it for when you click the remove button ads that has a lot more impact if you remove one in error!

PR #1543, opened 2017-02-24
Remove are you sure prompt when adding new extensions

canstudios-nicolaw:
Fix for #1519


taylortom:
Awesome, thanks @canstudios-nicolaw!

PR #1550, opened 2017-02-27
Add framework updater plugin

taylortom:
Allows updating of the framework via the frontend


lc-thomasberger:
Would it be possible to select the framework version from a list of releases?

taylortom:
I guess it could -- I've written my own update code which just performs a git reset, so would be easy to implement.

I do have some usability concerns around being able to freely up/downgrade though -- while thinking about this, I've realised I need to limit updates to the same major version number.

canstudios-louisem:
Could we get a config variable to disable this (hide it from the menu and disable the api).

We like to manage our versions and lock to specific ones so that we know all out instances are running the same.

taylortom:
@tomgreenfield this could be one to revisit (or close) now we've got **v0.6.3** in

PR #1558, opened 2017-03-01
Add Logger colour support

taylortom:
Adds some nice colours to logger


taylortom:
Agreed, need to look at colour contrast on all OSes.

taylortom:
@adaptlearning/authoring-tool-push how does everyone feel about this being added? Will close this if there's no interest.

aniketdharia:
👍

PR #1559, opened 2017-03-01
Refactor ffmpeg thumb generation

taylortom:
Makes the code a bit more consistent/readable.

Also reworked the video thumb slightly to give a better preview.

### DIFF
--- a/plugins/filestorage/localfs/index.js
+++ b/plugins/filestorage/localfs/index.js
@@ -401,7 +401,7 @@ LocalFileStorage.prototype.createThumbnail = function (filePath, fileType, optio
 };
 
 function isThumbnailTypeSupported(type) {
-  switch(fileType) {
+  switch(type) {
     case 'video':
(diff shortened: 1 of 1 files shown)


plugins/filestorage/localfs/index.js
```diff
@@ -352,70 +352,65 @@ LocalFileStorage.prototype.copyAsset = function(asset, sourceTenantName, destina
...
+    return next(null, self.getRelativePath(imgThumbPath));
+  });
 
-  // can't do thumb
-  return next(null, false);
+  return ff.run();
 };
 
+function isThumbnailTypeSupported(type) {
+  switch(fileType) {
```
dancgray: Should be using type not fileType.

canstudios-louisem:
This was merged as I was about to post a comment. The video gif's have some bad artifacts in comparison to the old ones I think i needs to be looked a

PR #1601, opened 2017-03-16
Issue/1512

### LINKED ISSUE #1512: exported course can't be used for translation with manual modifications
There's a reasonable expectation from users that you should be able to export a course from the AT so that you can then use the language export/import feature of the framework - however, since the AT doesn't export the core .schema files this doesn't work unless the user manually adds them to the exported course.

Could the export feature be updated to include the .schema files?

Also, the versions of the .schema files in the AT have not been updated to have `translatable` properties.

dancgray:
fixes #1512  

Also added missing values for:
contentobject.pageBody
course.description

Is anything else needed?


dancgray:
@taylortom, that does make more sense. Added check for folders.

PR #1612, opened 2017-04-07
fixes #1611

### LINKED ISSUE #1611: _supportedLayout not working
I was trying to restrict a new component to full width only and it wasn't working. I couldn't figure out why until I started to dig into the code.

https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/core/editor/page/views/editorComponentListView.js#L82-L83

The string representations of the keys should not have a "." and L83 should not have a hasOwnProperty call as it is the value itself we are trying to return.

I will try and PR when I get a chance

danielstorey:
_supportedLayout was not working at all meaning left, right and full were always available even if a plugin's properties.schema was limited to `half-width` or `full-width` only. This PR addresses the issue in two areas:

Sidebar component list - when a component is selected only the options specified in the properties schema are available. 
Page editor - if a component is full-width only then the `component-move` arrow icons will not be rendered. If half width only then the full width arrow will not be rendered

### DIFF
--- a/frontend/src/core/editor/page/views/editorComponentListView.js
+++ b/frontend/src/core/editor/page/views/editorComponentListView.js
@@ -82,11 +82,7 @@ define(function(require) {
         var properties = componentType.get('properties');
         if (properties && properties.hasOwnProperty('_supportedLayout')) {
           var supportedLayout = properties._supportedLayout.enum;
-          var availablePositions = {
-            left: true,
-            right: true,
-            full: true
-          };
+          var availablePositions = _.clone(this.availablePositions);
 
           // Prune the available positions
           if (_.indexOf(supportedLayout, 'half-width') == -1) {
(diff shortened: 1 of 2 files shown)


frontend/src/core/editor/page/views/editorComponentListView.js
```diff
@@ -77,30 +77,36 @@ define(function(require) {
...
       Origin.trigger('editorComponentListView:removeSubviews');
       // _.each(this.collection, function(componentType) {
+
       this.collection.each(function(componentType) {
         var properties = componentType.get('properties');
-        if (properties && properties.hasOwnProperty('._supportedLayout')) {
-          var supportedLayout = properties.hasOwnProperty('._supportedLayout').enum;
+        if (properties && properties.hasOwnProperty('_supportedLayout')) {
+          var supportedLayout = properties._supportedLayout.enum;
+          var availablePositions = {
```
dancgray: @danielstorey I think availablePositions is also used to find positions that are not already occupied as set in setupFilters function. Assigning a new value to availablePositions here allows the user to insert two left (or right) positioned components.

danielstorey: Ok good point...so just need to combine both checks

dancgray: 👍

danielstorey: Latest commit should do it!

Issue #1620, opened 2017-04-22
Add attribution metadata to asset catalogue


davetaz:
1) Add attribution metadata to asset catalogue consisting of 
- Asset attribution (text)
- Asset source (url)
- Asset licence
None are mandatory

2) Add plugin to authoring tool to auto fill the attribution field of components that have it when selecting an asset.

PR #1621, opened 2017-04-22
Issues/1620

davetaz:
Closing issue #1620 

**Important:**
All fields in the asset manager are required regardless of config. There is a related issue that needs closing for this to be fixed (see #1120).

**Notes**
I have added attribution to most components in adapt-framework that have graphics so this works with more than just adapt-contrib-graphic. (For reference these were adaptlearning/adapt-contrib-accordion/pull/81, adaptlearning/adapt-contrib-gmcq/pull/108 and daptlearning/adapt-contrib-narrative/pull/171, hotgraphic still to do).


davetaz:
Travis build will fail due to all asset metadata being required even when required="false" in config. See #1120

Issue #1517, opened 2017-02-10
Router improvements


taylortom:
- Should the router handle adding views? (for some reason, the sidebar handles the adding of its views, which makes more sense)
- I think we should maybe try to decouple the router from the specific Backbone.Router functionality if possible 
- [Need less hard-coded routes](https://github.com/adaptlearning/adapt_authoring/issues/1516)

taylortom:
Fixed as part of 0.4.0

Issue #1369, opened 2016-09-05
Standardise editor page container div


taylortom:
Specifically look into rendering all pages into the `.editing-overlay` by default. As part of the UI refresh, there have been various LESS fixes which have had to be duplicated across various different screens/files as this isn't the case.

PR #1627, opened 2017-05-02
Major frontend refactor

taylortom:
This PR aims to reorganise the `frontend` folder to make it a bit easier to work with.

### Bugs
- [x] Moving a block in page editor doesn't remove it from its current article, resulting in duplicate blocks (refreshing the page fixes this)
- [x] In page editor drag and drop of articles and blocks does not work correctly (Components works OK)
- [x] The navbar help link redirects to an empty page when clicked on the dashboard, asset management, user management and plugin management.
- [x] If you go to the user profile, and press 'Change password' the little eye icon that shows the password is on the far right hand side of the page, instead of inside the box.
- [x] Deleting a sub-menu's page page in the menu editor redirects you to page editor. It should stay on the menu editor (deleting a page at the top level works fine)
- [x] Server 500 error when trying to copy an article, block or component in the page editor. Error occurs when you select the past location. [error] - More than one parent found
- [x] In the menu editor copying a page in a submenu pastes it to the top menu not the submenu that was selected
- [x] After uploading a plugin the upload spinner and saving text is not refreshed
- [x] User edit - after changing the role the new role is not displayed
- [x] Plugins don't show correct enabled/disabled style on change
- [x] Notify destructive animation out of position
- [x] Notify destructive not clearing timer on close
- [x] Close context menu on click outside
- [x] Add copy ID to project
- [x] Context menu buttons sometimes do not work
- [x] Plugin management: Button reverts to 'check for update' after update
- [x] User manager: keep user selected after enable
- [x] Component edit loads config edit
- [x] Theme editor "back to page" link breaks when accessed from page editor
- [x] Missing translation (app.editorarticle, app.editorblock, app.editorcomponent) from ABC editors
To reproduce:
  Change the role of a user.
  Click the tick to save
  Role displayed reverts to previous value
  Refresh the page and the new role is displayed.

### Enhancements
- [x] Change styling of 'Click to update' plugin management button for emphasis

### Out with the old
For reference, here's the current structure:
```
src/
  adaptbuilder/     -- output/build
  core/             -- all of the code
  images/           -- some global assets
  less/             -- LESS overrides
  plugins/          -- any extra (non-core) modules - userManagement is in here
  templates/        -- built handlebars templates
```
### In with the new
Front-end folder now looks like the following:
```
build/            -- renamed adaptbuilder, moved out of src
src/
  core/           -- previously core/app
  libraries/      -- all third-party libs, moved from core/app
  modules/        -- all contents of core except app
  templates/
```

## Notes

- All `src/less` files have been merged with their module LESS files, and the folder removed
- All `src/images` have been moved to appropriate module asset folders, and the folder removed
- Made sure all *used* assets in `adaptbuilder` are present in the `src` folder
- Dashboard renamed to 'projects'
- `src/core/project` files have been redistributed to more sensible locations (mostly the projects folder)
- `userManagement` has been moved into `modules`.
- All `modules` sub folders are loaded dynamically, rather than having to add the path to `app.js` as we go at the moment. I've done this by building a separate bundled module file `modules.js` which works in a similar way to the `plugins.js`.
- Tidied RequireJS config:
  - Removed mystery root `config.js` file
  - Removed `editorXXX` paths
  - Removed `coreJS` -- this never made any sense, as path was simply `core`
- Set up LESS compiler to check in `src/core/less`, so any LESS files here can be `@import 'variables'`, rather than needing to specify a long relative path
- I've modified which folders are served statically by the app in `lib/application.js`. In production mode, this is only the `build` and `libraries` folders, in dev mode we also allow the entire `src` folder.
- Have rewritten `app.js` and `origin.js` for readability. `origin.js` takes over the remaining view code from `router.js` (i.e. loading code, 'remove:views' event)
- Abstracted polyglot to `Origin.l10n` in case we want to customise/change later
- Refactor router.js:
  - Abstracted `navigate()` calls
  - Added shortcut functions (`navigateToHome`, `navigateToLogin`)
- Refactored editor code. Main points:
  - Add preloading of origin.editor.data
  - Replace editingOverlay with new contentPane, which now applies to *all* views (fixes inconsistencies with how scrolling etc. is dealt with, see https://github.com/adaptlearning/adapt_authoring/issues/1369)
  - Removed `coreJS` requireJS path (this was longer than the actual path `core`...)
  - Editor folder cleared out: 
    - Appropriate files moved to `core/app/` and `editor/global` (mostly models)
    - sub folders now more appropriately map to pages
    - page renamed to contentObject, as more representative of function
    - Editor routing moved to editor submodules (see `editor/**/index.js`)
    - Removed `editorXXX` requireJS paths in favour of relative paths

## Fixes

- #1517 
- #1369


frontend/src/modules/projects/views/projectsSidebarView.js
```diff
@@ -0,0 +1,157 @@
...
+      this.collection.each(function(tag) {
+        var availableTagsTitles = _.pluck(availableTags, 'title');
+        var usedTagTitles = _.pluck(this.usedTags, 'title');
+        if (!_.contains(availableTagsTitles, tag.get('title')) && !_.contains(usedTagTitles, tag.get('title'))) {
+          availableTags.push(tag.attributes);
+        }
+      }, this);
+
+      Origin.trigger('sidebar:sidebarFilter:add', {
+        title: Origin.l10n.t('app.filterbytag'),
```
dancgray: Should be app.filterbytags

dancgray:
👍 looking good, but getting error creating course:
Uncaught ReferenceError: Helpers is not defined :5000/modules/editor/course/index.js:30

canstudios-nicolaw:
This looks really good to me so far. I've found a few issues, hopefully they will be straightforward to fix :)

- I get an error when previewing a course:
The following modules share the same URL. This could be a misconfiguration if that URL only has one anonymous module in it:
.../adapt_authoring/temp/590d9f2821d82ab30592335f/adapt_framework/src/core/js/adapt.js: core/js/adapt, coreJS/adapt
.../adapt_authoring/temp/590d9f2821d82ab30592335f/adapt_framework/src/core/js/offlineStorage.js: core/js/offlineStorage, coreJS/offlineStorage

- And it looks like there is a similar error for download.

- When attempting to choose a menu and save (in this case just choosing box menu) I get:
Uncaught TypeError: this.remove is not a function
    at Object.<anonymous> (editorOriginView.js:191)
    at editorDataLoader.js:104

- When removing an extension from the course I have to refresh the page to see that the extension has been removed

taylortom:
@canstudios-nicolaw have fixed the last two issues, haven't been able to replicate the preview bugs though, will keep looking for those.

canstudios-nicolaw:
@taylortom  Hm I don't seem to be able to reproduce those now either, I was looking from home before, so it must be a problem on my home machine I wouldn't worry about it.

I will have another look at the PR later and hopefully approve :)

taylortom:
@canstudios-nicolaw cool :) It looks like a framework build issue anyway, so unlikely to be something I've introduced 🤞

taylortom:
@tomgreenfield I should have fixed all of your issues bar the plugin update one.

tomgreenfield:
Brilliant, plugin updates now appear to be functioning.

Something to note though is that upon a successful update, the button state immediately returns to _Check for update_, rather than sticking at _Up to date!_.

canstudios-nicolaw:
With the 'Restrict project tile height to ensure neat grid' commit the tile just cuts off if the course title is long. The project tags are then not visible. Since the project tiles being different heights is not strictly a bug, and this PR is so large already, would it be best to leave that commit out of this PR?

restrict tile height

canstudios-nicolaw:
Thanks for all your work on this Tom :+1: 

I've noticed a couple of minor things

The help link in the nav bar at the top of the page seems to redirect you to an empty github page. I've noticed this when clicking the help link on the dashboard, asset management, user management and plugin management. 
Both the dashboard and user management help links take you to a populated page on develop.

If you go to the user profile, and press 'Change password' the little eye icon that shows the password is on the far right hand side of the page, instead of inside the box.

dancgray:
Looking good. Regarding "Deleting a page in the dashboard redirects you to page editor"  this can be reproduced by deleting a page in a submenu. Deleting a page at the top level works OK.

taylortom:
@canstudios-nicolaw yep happy to revert the project styling, although the current styling gives you layouts like this:

screen shot 2017-07-03 at 16 11 38

canstudios-nicolaw:
@taylortom Yes we have noticed courses like this, it isn't beautiful but I also wouldn't say it is broken. I think an ellipsis approach would be better, but is it needed as part of this PR?

taylortom:
Nope, I agree it's not breaking anything per se, so doesn't have to be fixed immediately.

canstudios-nicolaw:
@taylortom Thanks Tom, I'll take another look at this today

canstudios-nicolaw:
I odn't think the project tile height has been fully reverted. It looks like there is a "height: 341px;" rather than a "min-height: 341px;"

taylortom:
@canstudios-nicolaw I can't find that anywhere, can you link me please? The [revert commit](https://github.com/adaptlearning/adapt_authoring/pull/1627/commits/d97d24797f2d9d7be115796787f0171bc8ee0ccc) looks ok, so may have happened later

canstudios-nicolaw:
@taylortom I've just checked and this seems to be fine now

canstudios-nicolaw:
Looks good to me :) I've just noticed one thing:

When uploading a plugin I get the following in the console:

`Uncaught ReferenceError: data is not defined
    at Object.<anonymous> (pluginManagementUploadView.js:49)
    at index.js:45
    at success (schemasModel.js:20)
    at Object.options.success (backbone.js:433)
    at fire (jquery.js:3143)
    at Object.fireWith [as resolveWith] (jquery.js:3255)
    at done (jquery.js:9309)
    at XMLHttpRequest.callback (jquery.js:9713)`

taylortom:
Good spot, have removed.

PR #1632, opened 2017-05-11
Fixes #1631

### LINKED ISSUE #1631: Full width only components can still be dragged left or right into an empty block
This is related to #1611 which was partially fixed in #1612.

**Steps to reproduce:**
* Create a  block
* Add a full width only component to the block
    *Placement and movement within the initial block will be restricted to full only which is the desired effect*
* Add a new block
* Drag the above component into the new block

**Expected result:**
Only the full width option should be enabled

**Actual result:**
Left and right options are enabled as well

danielstorey:
Ensure only the supported drop zones are displayed when dragging a component from one block to another.


canstudios-nicolaw:
This is currently going to the master branch, should it not be going to develop?

taylortom:
Thanks @danielstorey, we'll roll this into the next release :)

frontend/src/modules/editor/contentObject/views/editorPageComponentListView.js
```diff
@@ -79,30 +79,32 @@ define(function(require) {
...
-            model: componentType,
-            availablePositions: this.availablePositions,
-            _parentId: this.model.get('_parentId'),
-            $parentElement: this.$parentElement,
-            parentView: this.parentView,
-            searchTerms: componentType.get('displayName').toLowerCase()
-          }).$el);
+        this.$('.editor-component-list-sidebar-list').append(new EditorComponentListItemView({
+          model: componentType,
+          availablePositions: availablePositions,
```
tomgreenfield: `availablePositions` could now potentionally be `undefined`. Suggest line 87 is moved out to sit above the if statement.

taylortom: Like that?¿

frontend/src/modules/editor/contentObject/views/editorPageComponentListView.js
```diff
@@ -78,31 +78,33 @@ define(function(require) {
...
 
-        this.$('.editor-component-list-sidebar-list').append(new EditorPageComponentListItemView({
-            model: componentType,
-            availablePositions: this.availablePositions,
-            _parentId: this.model.get('_parentId'),
-            $parentElement: this.$parentElement,
-            parentView: this.parentView,
-            searchTerms: componentType.get('displayName').toLowerCase()
-          }).$el);
+        this.$('.editor-component-list-sidebar-list').append(new EditorComponentListItemView({
```
tomgreenfield: Getting an error when I try to add a component. I don't think the view should have been renamed here?

PR #1458, opened 2016-11-30
Tenant management MVP 0.0.1

aniketdharia:
Tenant management MVP 0.0.1

### Requirements:
- [x] Super admin: Able to view existing tenancies
- [x] Super admin: Able to create new tenancies
- [x] Super admin: Able to update existing tenancies
- [ ] Super admin: Able to delete existing tenancies
- [x] Super admin: Able to transfer an individual user from one tenant to another
- [ ] Tenant admin: Able to view users in the tenancy
- [ ] Tenant admin: Able to create new users in the tenancy
- [ ] Tenant admin: Able to update existing users in the tenancy
- [ ] Tenant admin: Able to delete existing users from the tenancy
- [ ] Tenant admin: Able to assign tenancy-scope system roles to users within the tenancy
- [ ] Tenant admin: Able to unassign tenancy-scope system roles to users within the tenancy
- [ ] Tenant admin: Able to remove any course in the tenancy
- [ ] Tenant admin: Able to change the owner of any course in the tenancy

### Other
- [ ] Update to fit with final user management UI (specifically the removal of edit mode)
- [ ] Remove space plugin
- [ ] Should tenant admin be able to access tenant management for their own tenant? (to change name etc.)

PR #1637, opened 2017-05-17
Tenant management

taylortom:
Tenant management MVP 0.0.1

### Requirements:
- [x] Super admin: Able to view existing tenancies
- [x] Super admin: Able to create new tenancies
- [x] Super admin: Able to update existing tenancies
- [ ] Super admin: Able to delete existing tenancies
- [x] Super admin: Able to transfer an individual user from one tenant to another
- [ ] Tenant admin: Able to view users in the tenancy
- [ ] Tenant admin: Able to create new users in the tenancy
- [ ] Tenant admin: Able to update existing users in the tenancy
- [ ] Tenant admin: Able to delete existing users from the tenancy
- [ ] Tenant admin: Able to assign tenancy-scope system roles to users within the tenancy
- [ ] Tenant admin: Able to unassign tenancy-scope system roles to users within the tenancy
- [ ] Tenant admin: Able to remove any course in the tenancy
- [ ] Tenant admin: Able to change the owner of any course in the tenancy

### Other
- [ ] Update to fit with final user management UI (specifically the removal of edit mode)
- [ ] Remove space plugin
- [ ] Should tenant admin be able to access tenant management for their own tenant? (to change name etc.)

Also see https://github.com/adaptlearning/adapt_authoring/pull/1458 for previous conversation thread.


frontend/src/plugins/tenantManagement/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+    if (data.allTenants.length > 0) {
+      var tenantId = Origin.sessionModel.get('tenantId');
+
+      window.model = data.allTenants.where({_id:tenantId});
+
+      if (Origin.permissions.hasPermissions(data.featurePermissions) && model.length > 0 && model[0].get('isMaster') == true) {
+        var globalMenuObject = {
+          "location": "global",
+          "text": "Tenant Management",
```
taylortom: Can we localise this please?

frontend/src/plugins/tenantManagement/views/tenantView.js
```diff
@@ -0,0 +1,190 @@
...
+      this.listenTo(this, 'remove', this.remove);
+    },
+
+    render: function () {
+      OriginView.prototype.render.apply(this, arguments);
+      this.applyStyles();
+    },
+
+    applyStyles: function () {
+      if (this.model.get('_isDeleted') === true) {
```
taylortom: We could maybe rewrite this using [$.toggleClass](http://api.jquery.com/toggleclass/#toggleClass-className-state)

frontend/src/plugins/tenantManagement/views/tenantView.js
```diff
@@ -0,0 +1,190 @@
...
+      });
+    },
+
+    onError: function (error) {
+      Origin.Notify.alert({
+        type: 'error',
+        text: error.message || error
+      });
+    }
+
```
taylortom: Whitespace looks a bit off down here

lib/tenantmanager.js
```diff
@@ -137,18 +137,6 @@ exports = module.exports = {
                 nextKey();
               }, function () {
                 var query = {};
-                if (orList.length) {
```
taylortom: Has AND/OR support been removed for a reason?

package.json
```diff
@@ -72,6 +72,7 @@
     "bcrypt-nodejs": "0.0.3",
     "body-parser": "^1.13.3",
     "bower": "1.6.5",
+    "du": "^0.1.0",
```
taylortom: Is this still used?

routes/lang/en-application.json
```diff
@@ -320,5 +320,28 @@
   "app.list": "List",
   "app.ascending": "A to Z",
   "app.descending": "Z to A",
-  "app.visitpluginhomepage": "Visit plugin homepage"
+  "app.visitpluginhomepage": "Visit plugin homepage",
+  "app.ismaster":"IsMaster",
+  "app.createdat":"Created On",
+  "app.totaltenantsize":"Total Size",
```
taylortom: Can you please remove any unused strings?

frontend/src/plugins/tenantManagement/templates/addTenant.hbs
```diff
@@ -0,0 +1,28 @@
...
+    <div class="form-group">
+      <label for="tenantName"> {{t 'app.tenantname'}}
+      <span class="req">*</span>
+      <span id="tenantNameError" class="error"></span>
+    </label>
+      <input type="text" id="tenantName" name="name" value="{{tenantName}}" autocomplete="off">
+    </div>
+
+    <div class="form-group">
+      <label for="tenantDisplayName">{{t 'app.displayname'}}
```
canstudios-louisem: `app.displayname` no longer exists I believe it should be `app.tenantname`

frontend/src/plugins/tenantManagement/templates/tenantManagement.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="container">
+  <div class="inner">
+    <div class="tb-heading col-row tb-row">
+      <div class="col-15">
+        <h4 class="tb-col-inner">{{t 'app.tenant'}}</h4>
+      </div>
+      <div class="col-15">
+        <h4 class="tb-col-inner">{{t 'app.displayname'}}</h4>
```
canstudios-louisem: `app.displayname` no longer exists I believe it should be `app.tenantname`

frontend/src/plugins/tenantManagement/templates/tenantManagementSidebar.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="sidebar-item-inner">
+  <div class="action-button-group">
```
canstudios-louisem: `<div class="separator"></div>` needs to be added here so the spacing matches the other sidebar buttons.

taylortom:
The 'edit' functionality seems to have been dropped (probably during the upgrade to the updated UI). Can we have something similar to user management?
screen shot 2017-05-17 at 12 25 59

taylortom:
I've updated the requirements TODO list to reflect the latest changes. Looks to be quite a bit to do still.

canstudios-louisem:
If you add a tenant then go straight to user management the tenant you created is not in the list, you have to refresh to get it show up.

dancgray:
Should we capture the requirement to move/copy a course and it's assets from one tenant to another. I don't think the last requirement (Able to change the owner of any course in the tenancy) matches this.

aniketdharia:
@canstudios-louisem the point regarding the tenant name not displayed, we would need to trigger an event from TM to UM to update the drop-down list. This will add a dependency to have both the plugins installed for this to work. Would it be fine to do it or should we just refresh the page once a tenant is added?

canstudios-louisem:
@aniketdharia I think we should do it properly there is no where else in the tool that refreshes the page

aniketdharia:
ok! @ambadasavhad will get in touch with the solution that we have used, we can then decide which approach to use moving forward.

taylortom:
@aniketdharia @canstudios-louisem sounds like there's a bug in the way the list of tenants is generated, either in user management, or at the backend

ambadasavhad:
@taylortom edit functionality is already there, by clicking row you can edit tenant same as user management.

taylortom:
@ambadasavhad ah, I didn't see that when I tested. I don't think we have the same rollover styling as the user list items.

ambadasavhad:
@taylortom  I will add it in next commit

frontend/src/plugins/tenantManagement/views/tenantView.js
ambadasavhad: @taylortom as we are adding or removing class on the basis of condition other than having a class or not, I  don't think we can use[ $.toggleClass](http://api.jquery.com/toggleclass/#toggleClass-className-state) here

tomgreenfield: @taylortom will have been looking for code along the lines of:
```js
this.$el
  .toggleClass('inactive', this.model.get('_isDeleted') === true)
  .toggleClass('selected', this.isSelected);
  
this.$('.edit-mode').toggleClass('display-none', !this.isSelected);
this.$('.write').addClass('display-none');
```
(untested, just refactored from your current logic)

canstudios-louisem:
Currently, when a tenant is created even a none master one a new DB is created, this is not great if you want to run multiple installs from one mongo cluster.

canstudios-louisem:
Currently, a tenant admin can upload plugins, It was agreed when we had the requirements meeting that should not be the case. Should we just hide the plugin management area total from anyone who is a not a super admin?

taylortom:
@canstudios-louisem we haven't included the 'ability to enable plugins at the tenant-level' in the MVP, so that's a fine approach for this release.

canstudios-louisem:
Has anyone made @ambadasavhad aware of the above?

frontend/src/plugins/userManagement/views/userManagementView.js
```diff
@@ -42,42 +42,50 @@ define(function(require){
...
       this.setViewToReady();
       this.$('.users').fadeIn(300);
     },
 
-    refreshUserViews: function(event) {
+    fetchUsers: function () {
+      if (this.model.get('globalData').hasSuperAdminPermissions) {
+        this.users.fetch();
+      } else if (this.model.get('globalData').hasTenantAdminPermissions) {
+        this.users.fetch({ data: $.param({ _tenantId: Origin.sessionModel.get('tenantId') }) });
```
taylortom: I think the data is serialised automatically by the fetch (I may be wrong...)

frontend/src/core/app/permissions.js
```diff
@@ -72,6 +72,48 @@ define(function(require) {
 			
 		},
 
+		hasSuperPermissions:function(){
+			var sessionModelPermissions = Origin.sessionModel.get('permissions');
+			var hasWildCard = false;
+
+			_.each(sessionModelPermissions, function(permission) {
```
taylortom: It's probably a bit more readable if you use `the permissions.hasPermissions` function, and pass `['*/*:delete']` (just specifying delete should be enough?)

frontend/src/core/app/permissions.js
```diff
@@ -72,6 +72,48 @@ define(function(require) {
...
+
+			if (hasWildCard) {
+				return true;
+			}else{
+				return false;
+			}
+
+		},
+
+		hasTenantAdminPermission:function(){
```
taylortom: See above, but use `['{{tenantid}}/*:delete']`

frontend/src/core/app/permissions.js
ambadasavhad: By using  `['{{tenantid}}/*:delete']` and `['*/*:delete']`  not able to differentiate between tenant Admin and super Admin, as  `the permissions.hasPermissions`  returns true for both

canstudios-nicolaw:
Hi ambadasavhad,

Looks good to us so far, thanks for your work on this :) 

We have noticed a few issues:

- Tenant admins can change a super admins role to something else
- Tenant admins don't always see a full list of the users in their tenant. Log in as a tenant admin, navigate to user management and notice that not all users in the tenant are in the list. Each time you refresh the page more users become visible. (This only seems to happen once per log in)
- Tenant Admins can only create users within their tenant. As part of this, the "Tenant" dropdown is grayed out in the "Add new user" form. However, the newly created user does not appear in the user management list for the Tenant Admin. Logging in as Super Admin will reveal that the new user is not assigned to a particular tenant at all.

canstudios-louisem:
XSS! When creating or editing a tenant the name and display name are not sanitised for script tags and when the data is displayed the tags are not escaped this also occurs on the user page if there is a user in the affected tenant.

Upon inspection of the hbs files `{{{` has been used in a lot of places that `{{` should have been.

ambadasavhad:
@taylortom sure.

lib/tenantmanager.js
```diff
@@ -276,23 +255,30 @@ exports = module.exports = {
...
+   * @param {object} tenant - tenant details, only details on create tenant whitelist will be accepted
    * @param {function} callback - function of the form function (error, tenant)
    */
   createTenant: function (tenant, callback) {
     var self = this;
 
+    // Strip out fields that are not on whitelist
+    Object.keys(tenant).forEach(function(key) {
+      if (!CREATE_TENANT_WHITELIST.includes(key)) {
+        delete delta[key];
```
dancgray: getting ReferenceError: delta is not defined on install. Should this be `delete tenant[key]; `

canstudios-nicolaw:
I've just had a look at this branch, I've checked things off that have been completed. I haven't checked off all items for the following reasons:
- A super admin can disable a tenant but they do not have an option to delete the tenant
- A tenant admin can select the super admin role when creating a user
- A tenant admin can edit an existing user so that they have the super admin role

I have also noticed a bug:
- As a super admin, if you create a tenant and then navigate to the create user page you must refresh the page before the newly created tenant appears in the drop down

taylortom:
Thanks @canstudios-nicolaw

taylortom:
The biggest outstanding bit of work is to merge the latest changes to develop into here.

ambadasavhad:
@canstudios-nicolaw I have fixed above listed issues in the recent commit, can you please have look at it?

taylortom:
@ambadasavhad there are still major merge conflicts with develop, can you fix those please?

canstudios-nicolaw:
thank you @ambadasavhad I will take a look at those bug fixes next week :) As Tom said, it would be brilliant if you could fix those conflicts with develop

frontend/src/plugins/userManagement/templates/user.hbs
```diff
@@ -58,10 +58,21 @@
               <a class="edit read edit-mode"><i class="fa fa-pencil-square-o"></i></a>
               <div class="write">
                 <select class="input" data-modelKey="roles">
-                  {{#each globalData.allRoles.models}}
-                    {{!NOTE this won't do if we introduce multiple roles (../roles.[0])}}
-                    <option value="{{this.attributes._id}}"{{#ifValueEquals ../roles.[0]._id this.attributes._id}} selected{{/ifValueEquals}}>{{this.attributes.name}}</option>
-                  {{/each}}
+                  {{#if globalData.hasTenantAdminPermissions}}
+                    {{#each globalData.allRoles.models}}
+                      {{#ifValueNotEquals this.attributes.name "Super Admin"}}
```
taylortom: See above.

frontend/src/plugins/userManagement/templates/addUser.hbs
```diff
@@ -24,9 +24,17 @@
       <div class="form-group">
         <label for="role">{{t 'app.role'}}</label>
         <select name="role">
-          {{#each globalData.allRoles.models}}
-            <option value="{{this.attributes._id}}">{{this.attributes.name}}</option>
-          {{/each}}
+          {{#if globalData.hasTenantAdminPermissions}}
+            {{#each globalData.allRoles.models}}
+              {{#ifValueNotEquals this.attributes.name "Super Admin"}}
```
taylortom: This check should be permissions-based, rather than role-based. 
(see [Helpers.ifHasPermissions](https://github.com/adaptlearning/adapt_authoring/blob/develop/frontend/src/core/helpers.js#L191))

frontend/src/plugins/userManagement/helpers.js
```diff
@@ -66,6 +69,14 @@ define(function (require) {
       } else {
         return block.inverse(this);
       }
+    },
+
+    ifValueNotEquals: function (value, text, block) {
```
taylortom: Please remove this in-line with the below changes.

PR #1647, opened 2017-06-01
Add sequential locking on sub a menu, fixes #1625

### LINKED ISSUE #1625: Cannot lock a sub menu
In the authoring tool _lockType attribute can only be set on course model. This prevents sub menus from being locked as _lockType needs to be set on the menu contentObject.

See framework [wiki](https://github.com/adaptlearning/adapt_framework/wiki/Locking-objects-with-'_isLocked'-and-'_lockType') :

> The normal use case has the menu at the root level of the course and so the _lockType should be added to course.json. For a sub-menu, the _lockType should be added to the content object (in contentObjects.json) which is the parent of the pages to be locked. 

[_lockType](https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/content/course/model.schema#L598) attribute should be added to contentObject [schema](https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/content/contentobject/model.schema).

Unfortunately this will also be displayed on page type contentObjects.

"The schema needs to be updated to allow this to be set on content objects too."

dancgray:
Add sequential locking on sub a menu. Has the disadvantage of being shown on pages as well as menu objects.


plugins/content/contentobject/model.schema
```diff
@@ -73,6 +73,23 @@
...
+        "options": [
+          "",
+          "custom",
+          "lockLast",
+          "sequential",
+          "unlockFirst"
+        ]
+      },
+      "title": "Menu lock type",
+      "help": "Defines which (if any) locking mechanism will be used. For menu items only.",
```
tomgreenfield: Could we rephrase to say only submenus? _Menu items_ implies pages as well.

dancgray:
@tomgreenfield, corrected misleading wording.

canstudios-louisem:
Are we sure we are happy with the option being in places where it has no affect? I can see it confusing users.

lc-thomasberger:
Should we use `editorAttrs` to show the setting only when the contentObject is not a page?

PR #1548, opened 2017-02-27
Upgrade Backbone Forms; move overrides out

tomgreenfield:
Resolves #1544.

PR #1660, opened 2017-06-16
Added a Scaffold override to allow for specifying 'min’, 'max' values…

### LINKED ISSUE #1659: Ability to limit 'Number' editor to a specific range
It would be good to be able to specify min and max attributes for Number editors in the schemas.

This could be achieved via a Scaffold override, perhaps making use of the Backbone Forms validators i.e.
      "validators": [
        {"type": "range", "min": 1, "max": 100}
      ]

If detected, the Scaffold override could then apply the 'min' and 'max' attributes to the editor element. This ensures there is also a validation rule applied to the editor.

It seems that 'step' can be defined within an 'editorAttrs' object

barrymckay397:
… for Number inputs. 'Step' is also configurable. These values can be supplied in the schema like so:
      
```
{
    "step": 1,
    "validators": [
        {"type": "range", "min": 1, "max": 100}
    ]
}
```
Fixes #1659

### DIFF
--- a/frontend/src/core/scaffold/scaffold.js
+++ b/frontend/src/core/scaffold/scaffold.js
@@ -107,7 +107,7 @@ define(function(require) {
 			_.each(objectSchema, function(field, key) {
 				setupSchemaFields(field, key, objectSchema, scaffoldObjectSchema);
 			});
-			
+
 		} else if (field.type != 'object' || field.inputType) {
 
 			var validators = [];
@@ -129,7 +129,7 @@ define(function(require) {
                 console.log('No validator of that sort - please register: "' + validator + '" by using Origin.scaffold.addCustomValidator(name, validatorMethod)');
               }
               // If match is found - add the method
-              validators.push(customValidator.validatorMethod);  
+              validators.push(customValidator.validatorMethod);
             }
 					}
 				}
@@ -152,12 +152,6 @@ define(function(require) {
 
 			scaffoldSchema[key] = fieldObject;
 
-			var editorAttrs = field.editorAttrs;
-
-			if (editorAttrs) {
-				scaffoldSchema[key].editorAttrs = editorAttrs;
-			}
-			
 		} else {
 
 			scaffoldSchema[key] = {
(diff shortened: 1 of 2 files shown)


frontend/src/core/scaffold/scaffold.js
```diff
@@ -149,8 +149,15 @@ define(function(require) {
...
 			}
+
 			scaffoldSchema[key] = fieldObject;
 
+			var editorAttrs = field.editorAttrs;
+
+			if (editorAttrs) {
+				scaffoldSchema[key].editorAttrs = editorAttrs;
+			}
+
```
tomgreenfield: Can this be moved to sit after the mammoth if statement (currently ends on line 175), so that `object` and `array` types can get involved in the fun too?

frontend/src/core/scaffold/scaffoldOverrides.js
```diff
@@ -206,5 +206,4 @@ define(function(require) {
 
 	    return _.isEmpty(errors) ? null : errors;
 	}
-
```
tomgreenfield: I'd avoid changing whitespace in this file since it may cause conflicts with the Backbone Forms upgrade PR.

ryan-lynch:
Hey @barry-learningpool @brian-learningpool - Using a range validator, I was able to include a max of 10 and 10,000 without issues. However, I've discovered that if you do not supply a max, it seems to default to 100. Take a look at this: https://github.com/powmedia/backbone-forms/blob/master/distribution/backbone-forms.js#L578

barry-learningpool:
Damn that's annoying. This may need an override for the 'range' validator to disregard that default (or the one for 'min' if we were only specifying 'max').

The range validator is actually only introduced in a more recent version of Backbone Forms, however there's an outstanding PR to update Backbone Forms, so this will become a relevant issue in the future. PR is here: https://github.com/adaptlearning/adapt_authoring/pull/1548

@tomgreenfield any thoughts on this?

tomgreenfield:
Yeah, Backbone's range validator is designed to have a defined `min` and `max`.

If you don't want the default `100`, the least hacky way would be to just add a large `max` to the schema. This is similar to how we simulate 'infinite' `_attempts` on questions.

tomgreenfield:
Superseded by https://github.com/adaptlearning/adapt_authoring/tree/feature/scaffoldRefactor,

Issue #1511, opened 2017-02-08
Remove build includes from export


taylortom:
As we only add the relevant plugins, this often adds extra complexity to those not familiar with the feature.

moloko:
when this is done and released, please could you delete [the section on manually removing this property](https://github.com/adaptlearning/adapt_authoring/wiki/Exporting-Course-Source-Code#remove-the-includes-array-from-the-configuration-file) from the wiki page on exporting?

PR #1664, opened 2017-06-26
Fix #1623 - Add build includes to output manager

### LINKED ISSUE #1623: All components and extensions are included in preview and  download
Removing the build includes from lib/outputmanager.js results in all components and extensions being included in preview and download.

Export source code only includes enabled extensions.

canstudios-nicolaw:
Revert #1586

### DIFF
--- a/lib/outputmanager.js
+++ b/lib/outputmanager.js
@@ -85,6 +85,11 @@ var Constants = {
       Rebuild: '.rebuild',
       CustomStyle: 'zzzzz.less',
       Variables: 'variables.less'
+    },
+    Modes: {
+      export: 'EXPORT',
+      preview: 'PREVIEW',
+      publish: 'PUBLISH'
     }
 };
 
@@ -200,9 +205,11 @@ OutputPlugin.prototype.getCourseJSON = function(tenantId, courseId, next) {
 
 /**
  * Sanatizes the component
- *
+ * @param mode string A string describing the type of output (eg export, publish, preview)
+ * @param json Course json
+ * @param next callback
  */
-OutputPlugin.prototype.sanitizeCourseJSON = function(json, next) {
+OutputPlugin.prototype.sanitizeCourseJSON = function(mode, json, next) {
   try {
     var outputJson = json;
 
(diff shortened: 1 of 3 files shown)


taylortom:
Can we check that this still resolves https://github.com/adaptlearning/adapt_authoring/issues/1511

canstudios-nicolaw:
Sorry Tom I'm not sure I follow. #1511 says "As we only add the relevant plugins, this often adds extra complexity to those not familiar with the feature."

I don't think it is a bug is it? It was a thought of how we could simplify something, removing it caused a bug, so my revert fixes that bug?

dancgray:
The build includes is required for download (and probably preview) unless you want all plugins include. It is part of the framework grunt process. However Tom is correct in the original 1511 ticket that export source code does not require build includes.  Export source is a seperate processes in plugins/output.

From memory part of the original reason for #1511 and the wiki note is that the AT added incorrect values in build.includes for the theme making it necessary to remove this code from the export source before building. I think this might have happened only if you added custom styles, possibly from [this](https://github.com/adaptlearning/adapt_authoring/blob/master/lib/outputmanager.js#L656) line.

I think the output plugin and lib/outputmanager should be cleaned up so that build includes are added for preview/download only. That would fix #1511 and #1623.

There are other problems with outputmanager. For instance these two lines cause problems if you have theme or menu settings:
https://github.com/adaptlearning/adapt_authoring/blob/master/lib/outputmanager.js#L566
https://github.com/adaptlearning/adapt_authoring/blob/master/lib/outputmanager.js#L855

For a course with custom menu that has settings, preview would try and build a course with a folder named 596d10de4cda3fb9033d6703-596ce2474bbc5efc1935aa5d which would not exist. Builds with no warning but no menu templates :(. I will make a seperate ticket for that.

canstudios-nicolaw:
@dancgray @taylortom I've added another commit. 

So now the build includes is only added if it is not an export.

The `isPreview` flag was true when it was a preview or export and false for download. I've changed the flag to be a "mode" which can describe which of the three processes is happening. The rest of the publish function is functionally unchanged.

canstudios-nicolaw:
I think it would be good to get this in 0.4.0 if we had time to test it, because it would fix a couple of issues.

taylortom:
Going to hold off merging this until `thebigbugfix` has gone in in case we've got conflicts

Issue #1626, opened 2017-04-26
Install script needs to be updated to encompass latest changes


canstudios-louisem:
- `rootUrl`: used for the password reset email. Currently, the install script does not set it in the config file.
- `useSmtp`
- `root`

dancgray:
The config wiki is [here](https://github.com/adaptlearning/adapt_authoring/wiki/Configuring-the-Authoring-Tool)

canstudios-louisem:
Do we want to add `trackingId`? most people add it after although some people may use if you do programmatic installs.

taylortom:
Don't see why not.

canstudios-louisem:
this is done in #1667

Issue #1570, opened 2017-03-03
Upgrade to specific versions


canstudios-louisem:
Currently the upgrade script always just gets the latest version.

It would be really useful to be able to explicitly specify what version's to upgrade or even downgrade to via arguments and if no arguments are input the current behaviour should remain.

PR #1667, opened 2017-07-03
Can upgrades to upgrade and install script

canstudios-louisem:
Install and upgrade script can now use any git repository framework and AT can continue to be updated automatically if they use the adapt community repo's or be updated to a specific git tag or branch.

Error's on the install and upgrade scripts are now passed to stderr, and they return none 0 error codes much less often when erroring.

By default, the latest framework version is now installed.

Version.json is not tracked and has been removed from VCS it is now generated when the install script is run, from the version in the package.json

To use it on install:
1. Pull this branch to an install.
1. `node install`
1. When asked to give your own AT and framework repo.
1. Pass a git branch or tag in the form `tags/[tag]` that you want the framework on or use the default it will get the latest. (what ever you pass is passed to `git checkout`)

To use on upgrade:
1. Enter `n` when asked if you wish to do an automatic upgrade.
2. Enter a git tag or branch for the AT and Framework. (what ever you pass is passed to `git checkout`)

If you chnage the repos in the config file they will be used on your next update.

Fixes:
* https://github.com/adaptlearning/adapt_authoring/issues/1626
* https://github.com/adaptlearning/adapt_authoring/issues/1570

Known Issues:
1. Github api is used to get the latest version tags but without authenticaton it limits your IP to 60 requests an hour this is not a lot when testing and currently the install fails once the 60 request limit is it hit.


canstudios-louisem:
Looks like the tests are failing because we removed version.json and install or upgrade have not been ran first to generate it. It will have to be created in the before and removed in the after.

canstudios-louisem:
Docs need to be updated for the config file and install also we should document the none interactive install method.

canstudios-louisem:
This was brought in as part of #1726

Issue #1673, opened 2017-07-10
Ability to set mongo auth source


canstudios-louisem:
Currently, your mongo user's auth source has to be the master tenant db. This causes issues as all the default roles are in the admin db.

PR #1674, opened 2017-07-10
Fix auth source issue

canstudios-louisem:
Fixes: https://github.com/adaptlearning/adapt_authoring/issues/1673

Allows you to set auth source as you do other database configs. `dbAuthSource` is the config attribute.

Docs would need to be updated here https://github.com/adaptlearning/adapt_authoring/wiki/Configuring-the-Authoring-Tool on merge.

### DIFF
--- a/lib/dml/mongoose/index.js
+++ b/lib/dml/mongoose/index.js
@@ -67,11 +67,13 @@ MongooseDB.prototype.connect = function(db) {
   
   // Construct the authentication part of the connection string.
   authenticationString = dbUser && dbPass ? dbUser + ':' + dbPass + '@' : '';
-  
+
+  var connectionString = '';
+
   // Check if a MongoDB replicaset array has been specified.
   if (dbReplicaset && Array.isArray(dbReplicaset) && dbReplicaset.length !== 0) {
     // The replicaset should contain an array of hosts and ports   
-    this.conn = mongoose.createConnection('mongodb://' + authenticationString + dbReplicaset.join(',') + '/' + dbName, options);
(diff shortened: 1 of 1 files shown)


lib/dml/mongoose/index.js
```diff
@@ -67,20 +67,28 @@ MongooseDB.prototype.connect = function(db) {
...
     dbHost = configuration.getConfig('dbHost');  
     dbPort = configuration.getConfig('dbPort'); 
 
     portString = dbPort ? ':' + dbPort : '';
 
-    this.conn = mongoose.createConnection('mongodb://' + authenticationString + dbHost + portString + '/' + dbName, options);
+    connectionString = 'mongodb://' + authenticationString + dbHost + portString + '/' + dbName;
   }
+
+  if(typeof configuration.getConfig('dbAuthSource') !== 'undefined' && configuration.getConfig('dbAuthSource') !== '' ){
```
tomgreenfield: Could we simplify this for readability?
```js
var authSource = configuration.getConfig('dbAuthSource');

if (authSource) {
  connectionString += '?authSource=' + authSource;
}
```

canstudios-louisem: I've kept the type checking but simplified it a bit.

tomgreenfield: `configuration.getConfig('dbAuthSource')` wouldn't ever return `undefined` because it falls back to an empty string: https://github.com/adaptlearning/adapt_authoring/blob/master/lib/configuration.js#L142.

canstudios-louisem: that's fine I've changed to just to check it is a string. It should have been that from the start.

PR #1683, opened 2017-07-24
Populate new courses so that they can be previewed

### LINKED ISSUE #1678: Newly created courses should be valid
When creating a new course it would be good if it contained a page, article, block and component.

It will then be possible to preview the course straight away, instead of working through a series of error messages.

canstudios-nicolaw:
Fixes #1678 

When you create a course the new course contains a page, article, block and text component. This means that the course can be previewed immediately.

I've put some text in the text component for now which I don't think will be final. We discussed some sort of welcome text in the stand up but didn't specify what it would be. Do we want just a straightforward 'This is an adapt course' type message or something a little more detailed that explains ABC?


frontend/src/core/project/models/projectContentObjectModel.js
```diff
@@ -0,0 +1,24 @@
+// LICENCE https://github.com/adaptlearning/adapt_authoring/blob/master/LICENSE
```
canstudios-nicolaw: I don't think this file will be necessary after the refactor PR

dancgray:
Block layoutOptions are not populated. Not sure if that matters? Otherwise looks good.

canstudios-nicolaw:
I wondered about that to. It didn't seem to do any harm when I took them out, so I left them out. Am happy to put them back in if that's the standard thing though

canstudios-nicolaw:
I have added the block layout options to be on the safe side. 

(I'm off next week so I don't want things waiting on me)

frontend/src/core/project/views/projectDetailView.js
```diff
@@ -60,6 +66,82 @@ define(function(require) {
...
+                  });
+
+                  componentTypes.fetch({
+                    error: _.bind(self.onSaveError, self),
+                    success: function() {
+                      var componentModel = new EditorComponentModel();
+                      componentModel.save({
+                        _courseId: model.get('_id'),
+                        _parentId: blockModel.get('_id'),
+                        body: 'This is a text component',
```
lc-thomasberger: It would ne nice if we can make this localizable ... or maybe use "Lorem ipsum ..." instead.

taylortom: +1, think that's a good idea @lc-thomasberger

taylortom:
Just noticed that various colorPicker assets have changed permissions, was this intentional?

canstudios-louisem:
No im not sure how that even happened must have been during the merge

frontend/src/modules/editor/course/views/editorCourseEditView.js
```diff
@@ -58,7 +63,84 @@ define(function(require) {
 
       var messageText = typeof response.responseJSON == 'object' && response.responseJSON.message;
       EditorOriginView.prototype.onSaveError.call(this, null, messageText);
+    },
+
+    /**
```
taylortom: Nesting reference

frontend/src/modules/editor/course/views/editorCourseEditView.js
```diff
@@ -58,7 +63,84 @@ define(function(require) {
...
+     * When a new course is created it gets populated with a page, article, block and blank component
+     * so that it can be previewed immediately.
+     * @param model
+     * @param response
+     * @param options
+     */
+    populateNewCourse: function(model, response, options) {
```
taylortom: Nesting reference

canstudios-nicolaw:
@taylortom I've refactored as per your suggestion (thanks for the code!). We have removed those colourPicker asset changes - sorry we didn't notice those as they didn't show on bitbucket.

PR #1684, opened 2017-07-25
Framework import feature

dancgray:
This PR adds the following functionality:

1. Export a course backup. Contains all plugins, assets and content with full metadata and tags for the course and each asset. This is in an authoring tool specific format.

2. Import a course backup, creates a new course from downloaded course backup. All course settings, assets and tags are imported with the exception of the hero image. Start page values are not corrected for the new page ID. Assets are not imported if a duplicate is found (matched on file name and size.). Plugins import is conservative, if the plugin is already installed or of a lower version the plugin will not be installed or updated. (prevents unintentional overwrite of data)

3. Framework import. Imports the adapt framework source code. There is an option to add tags. This will add tags to the imported course and assets on mass. Duplicate assets are not imported. Duplicate plugins are not imported or upgraded.

This PR is intended for testing of the code and UI suggestions. The UI is yet to be decided. Possibilities/tasks for this are:
- [ ] Renaming import/export buttons. 
- [ ] Combining framework and authoring tool import into one form
- [ ]  Combining authoring tool and framework into one export page with instructions for each option.
- [ ] Refactor and remove unzip module.
- [ ] Helpers [importPlugin](https://github.com/adaptlearning/adapt_authoring/blob/import-fixes-courseassets/plugins/output/adapt/helpers.js#L72) function needs better error handling


dancgray:
New PR for framework only import [here](https://github.com/adaptlearning/adapt_authoring/pull/1731/). 

It's better to close this and go for a simpler framework import. The AT export/import only added support for preserving tags. This could be worked into the framework export/import.

Issue #1580, opened 2017-03-08
Switch over to node v8


taylortom:
v4 goes into 'maintenance' from April, with v6 taking over as the active LTS. I've heard two reports that there's nothing to do for this, so may be a nice clean switch.

See https://github.com/nodejs/LTS for more.

canstudios-louisem:
Should we consider 8.x it is production ready and will go into LTS with the end of its maintenance being December 2019. It would give us even more of a performance boost with v8 5.8 and access to things like `util.promisify()`.

taylortom:
@canstudios-louisem I'm happy to take this approach, and aim to get a release out in october to support v8. 

Seems a bit pointless going for v6 at this point.

canstudios-louisem:
Should we update the [travis.yml](https://github.com/adaptlearning/adapt_authoring/blob/master/.travis.yml) file to start testing node 7 and 8

PR #1689, opened 2017-07-31
start to test node 7 and 8 

canstudios-louisem:
For #1580

Although we don't plan to move to v8 yet the tests pass and it would be good to keep them that way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ node_js:
   - "4"
   - "5"
   - "6"
+  - "7"
+  - "8" 
 
 sudo: required
 


canstudios-louisem:
I have updated this pr with the "fixed" tests and its now failing don't merge this until we have a passing build please.

taylortom:
The errors in 4 & 5 are caused by the length of the mocha timeout (fixed in #1699). 

I've just tried (and failed) to install adapt-cli with node v8, so this is likely a problem with the adapt-cli. There's something wrong with a URL for nodegit (returns a 403).

canstudios-louisem:
.Ill dig into it a bit when I get a chance as I was able to install adapt on node 8 with no issues but it doesn't have to go in the release.

canstudios-louisem:
It looks like the current version of node-git used in the adapt cli does not support node 7 and 8 and if we bump it we will lose v5 support. I will move this pr to the v8 milestone for now and we can revisit it then.

.travis.yml
```diff
@@ -4,6 +4,8 @@ node_js:
   - "4"
   - "5"
   - "6"
+  - "7"
+  - "8"
```
tomgreenfield: Please could this be changed to the below to align with the [CLI](https://github.com/adaptlearning/adapt-cli/blob/develop/.travis.yml#L3-L4)?
```yml
node_js:
  - "lts/*"
```

canstudios-louisem: I think we should explicitly define what versions we test otherwise when a lts version jumps all our tests may stop working.

tomgreenfield: I'd still be on board with having the dynamic LTS value.

I could see it as an advantage because by the time a version enters active LTS, it is pulled down by default for new Node downloads. The tests may help in flagging possible issues as they arise.

Happy to get a general consensus. Any thoughts @taylortom / @dancgray / @lc-thomasberger ?

taylortom: 👍

taylortom: It seems that just setting the value as `lts/*` will cause travis to use the latest active LTS, which at this point in time is v8 -- which I don't have a problem with (once we fix our v8 issues), because it'll force us to adopt current releases quicker... 

There's not much documentation on this, and can't find anything that matches our specific needs.

canstudios-louisem:
Closing this as its deprecated by https://github.com/adaptlearning/adapt_authoring/pull/1909

PR #1690, opened 2017-07-31
Use variables for font weights

### LINKED ISSUE #1638: Create LESS variables for font weights
Currently these are hard-coded, which causes a bit of a headache if you change the app font. We also don't seem to import the correct weights from Google fonts.

See https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/less/fonts.less

tomgreenfield:
* Resolves #1638
* Two named font weight variables:

  ```less
  @font-weight: 400;
  @font-weight-bold: 700;
  ```
* Revert header tags to bold weighting, removing the need for multiple overrides


tomgreenfield:
Any chance we could chuck this in for 0.4.0 @taylortom?

frontend/src/core/less/fonts.less
```diff
@@ -17,27 +18,10 @@ body {
...
-  font-weight:200;
-}
-
-h5 {
-  font-weight:200;
-}
-
-h6 {
-  font-weight:200;
+h1, h2, h3, h4, h5, h6 {
```
taylortom: Any reason we're changing these to use a single weight? (appreciate we weren't importing the specified weights)

tomgreenfield: > * Revert header tags to bold weighting, removing the need for multiple overrides

We have been overriding in several places just to get back to the (browser) default of bold weights for headings, making things a little convoluted. Also, the lighter weights have never been rendered from the offset, rather falling back to `normal`/`400`.

Issue #1502, opened 2017-02-06
Update mocha tests


taylortom:
In conjunction with #1440, the mocha tests need to be checked/updated.

See [issue/1502](https://github.com/adaptlearning/adapt_authoring/tree/issue/1502)

Issue #1649, opened 2017-06-07
Update package.json and version.json version numbers


canstudios-louisem:
Just a reminder so it doesn't get forgotten like it has in the past

tomgreenfield:
Note: with the develop branch, it seems that the version checking is a buggy regardless:
```console
You are currently running Adapt authoring tool v0.3.1 - v0.3.0 is now available
```

canstudios-louisem:
The current version checking is very simple it gets the latest tag from GitHub and does an if `!==` on that.

PR #1691, opened 2017-08-02
Issue/1502

canstudios-louisem:
* Bring in Kineo tests
* Tests now always start with a clean DB
* Upgrade should as Kineo tests needed it
* Fix tests broken from changes and should upgrade

#1502


lib/tenantmanager.js
```diff
@@ -489,18 +490,15 @@ exports = module.exports = {
...
         }
-
-        if (results && results.length === 1) {
-          return self.updateTenant({ '_id': results[0]._id }, { _isDeleted: true }, callback);
+        if (!result) {
+          return callback(new Error('No matching tenant record found'));
         }
-
-        return callback(new Error('No matching tenant record found'));
+        return self.updateTenant({ '_id': results[0]._id }, { _isDeleted: true }, callback);
```
tomgreenfield: `results` is no longer defined?

taylortom:
Tests should be working smoothly now

canstudios-nicolaw:
The tests are all passing for me, I have noticed I get the following output when I run them

`error: [07 Aug 2017 10:56:45 +01:00] test foo=bar`

I just wanted to check this is something we are expecting/aware of?

taylortom:
@canstudios-nicolaw that's because of [this test](https://github.com/adaptlearning/adapt_authoring/blob/issue/1502/test/lib.logger.js#L10-L17). We can set up the logger to mute all messages during the tests if we want?

canstudios-nicolaw:
Ok I am happy as long as we understand where it is coming from :) It feels a bit dangerous to mute all messages in case we miss something in future.

taylortom:
Have updated the test script to use the grunt 'test' task, thus removing the need for a makefile.

Gruntfile.js
```diff
@@ -349,6 +349,6 @@ module.exports = function(grunt) {
     }
   });
 
-  grunt.registerTask('test', ['mochaTest', 'casperjs']);
+  grunt.registerTask('test', ['mochaTest'/*, 'casperjs'*/]);
```
canstudios-louisem: Should we update the package.json's test command as it curently does make test not grunt test. Im not sure how much it matters really at Can we all use php storm's mocha test runner to run unit tests.

Gruntfile.js
```diff
@@ -357,7 +349,6 @@ module.exports = function(grunt) {
     }
   });
 
+  grunt.registerTask('test', ['mochaTest'/*, 'casperjs'*/]);
```
canstudios-louisem: Should we update the package.json's test command as it curently does make test not grunt test. Im not sure how much it matters really at Can we all use php storm's mocha test runner to run unit tests.

taylortom: The latest package.json should use `grunt test`

taylortom:
@canstudios-louisem you're right, I just knew if I waited to do a separate PR, I'd likely forget 😅

canstudios-louisem:
that's why i added a reminder months a go :) https://github.com/adaptlearning/adapt_authoring/issues/1649

Gruntfile.js
canstudios-louisem: Ah nice for some reason i missed you had changed it.

PR #1694, opened 2017-08-09
Remove unused files

### LINKED ISSUE #1503: Drop support for vagrant
This was discussed at the roadmap meeting.

Will involve the following:
- Removing relevant files from the repo
- Updating the wiki/removing references in the install instructions
- Potentially researching an alternative option (docker?)

taylortom:
Fixes #1503.

Also removes files associated with chef and heroku, which aren't supported anyway.


canstudios-louisem:
I had a quick look and it looks like you have removed all the Vagrant docs? is that right?

taylortom:
Yep, we were supposed to drop Vagrant support last release, but it was left. Thought now should be the time.

PR #1706, opened 2017-08-16
Remove grunt-casperjs from dependencies

taylortom:
As is currently broken...


canstudios-louisem:
Should we remove `test_frontend` as well? we can always get them back but they won't run at all now so it seems silly to leave them in.

taylortom:
Yeah, good idea

PR #1718, opened 2017-08-30
Basic building for ES lang file and rudimentary code to detect lang

canstudios-louisem:
This is a really basic pr to get the Spanish language file to be able to be used. 

Some more investigation needs to be done on the best way to detect the language the user wants or even just allow them to select it.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,6 +11,15 @@ module.exports = function(grunt) {
           'frontend/src/**/lang/en.json'
         ],
         dest: 'routes/lang/en.json'
+      },
+      es: {
+        src: [
+          'routes/lang/en-application.json',
+          'frontend/src/**/lang/en.json',
+          'routes/lang/es-application.json',
+          'frontend/src/**/lang/es.json'
+        ],
+        dest: 'routes/lang/es.json'
       }
     },
     copy: {
--- a/frontend/src/core/l10n.js
+++ b/frontend/src/core/l10n.js
@@ -13,7 +13,31 @@ define(['require', 'jquery', 'polyglot', 'core/origin'], function(require, $, Po
   /**
   * Initialise from language file
   */
-  var locale = localStorage.getItem('lang') || 'en';
+
+  //Rudimentary language detection https://stackoverflow.com/questions/673905/best-way-to-determine-users-locale-within-browser
+  var locale;
+  if (typeof navigator !== 'undefined'){
+    if(typeof navigator.languages !== 'undefined'){
+      locale = navigator.languages[0];
+    } else if(typeof navigator.language !== 'undefined'){
+      locale = navigator.language;
+    } else if(typeof navigator.browserLanguage !== 'undefined'){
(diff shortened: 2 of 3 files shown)


Gruntfile.js
```diff
@@ -11,6 +11,15 @@ module.exports = function(grunt) {
           'frontend/src/**/lang/en.json'
         ],
         dest: 'routes/lang/en.json'
+      },
+      es: {
+        src: [
+          'routes/lang/en-application.json',
```
canstudios-louisem: I have built on top of the English files that way if things are ever missing from the Spanish file it will at least display it in english.

frontend/src/core/l10n.js
```diff
@@ -13,7 +13,31 @@ define(['require', 'jquery', 'polyglot', 'core/origin'], function(require, $, Po
   /**
   * Initialise from language file
   */
-  var locale = localStorage.getItem('lang') || 'en';
+
+  //Rudimentary language detection https://stackoverflow.com/questions/673905/best-way-to-determine-users-locale-within-browser
+  var locale;
+  if (typeof navigator !== 'undefined'){
```
canstudios-louisem: I'm not happy with this code and I think we may be better allowing the user to explicitly request a language or looking at the locale headers passed in the api request.

taylortom:
Have just pushed a slightly different solution, which should take into account multiple languages.
See [compare/develop...issue/1715](https://github.com/adaptlearning/adapt_authoring/compare/develop...issue/1715?expand=1) for a comparison

I've also moved the build files into the temp folder, because it's always been confusing to have both the source and compiled files in the same folder. These should probably go into a new lang folder in the project's root.

Thoughts?

canstudios-louisem:
@taylortom  That would be a change to functionality slightly as it will now respond with a 404 rather than defaulting to English. If that's what we want to do it should probably respond with a 204 for the same reasons we did with theme previews.

In my version, it built everything off the English that way any missing strings would be in English, that way people could just translate a small area at first and there would be less of a push to update translations right away.

taylortom:
My thinking with the 404 was that we should be dealing with this via the UI, or more likely, to get a list of available languages from the server, and then only allow selection of these via the UI. An error would be more acceptable in this case I think.
The above is a more involved solution, so we may want to go with an alternative (quicker?) solution. 

I agree with you about automating the language -- think this should be a user selection (with a server default maybe set in the config.json, and again accessible from the server using either a lang-specific route, or from the exposed config settings)

canstudios-louisem:
Ah ok, I agree if the users are requesting a lang specifically, that does not exist that's a client side 4xx error it's only if we were doing it automatically that it should be a 204.

How do you want to move forward with this? shall we close my pr, for now, an open one from yours plus request some more input on the issue?

taylortom:
Have opened a PR #1721 

It probably makes sense to set up a feature development group for this. I know @lc-thomasberger has shown interest this previously, and I'm hoping we can work with @mariogalmonte to get this through quicker.

PR #1724, opened 2017-09-01
upgrade script now uses latest none draft or pre-release release

### LINKED ISSUE #1723: upgrade will just install the latest tag
We should change this to use the GitHub release API https://developer.github.com/v3/repos/releases/#get-the-latest-release.

This would also allow us and the framework people to push pre-release builds

canstudios-louisem:
fixes #1723

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -398,16 +398,16 @@ Origin.prototype.startServer = function (options) {
         headers: {
           'User-Agent' : 'Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.118 Safari/537.36'
         },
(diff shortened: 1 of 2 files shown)


oliverfoster:
would that pr exclude any releases with the "mark as pre-release" checkbox ticked? it could make it handy for terminal illiterates to publish releases which are excluded from the upgrade cycle

could do > https://api.github.com/repos/adaptlearning/adapt_authoring/releases + a filter for prerelease? make it a little more robush?

Louise McMahon @canstudios-louisem 11:51
the github releases/latest api only returns full releases not pre or draft releases

Oliver Foster @oliverfoster 11:52
lovely

taylortom:
Changes incorporated into https://github.com/adaptlearning/adapt_authoring/pull/1726

PR #1730, opened 2017-09-26
updates to the core schema files

moloko:
to:
* remove ARIA labels that aren't actually used
* add missing `incorrect` ARIA label ([#1732](https://github.com/adaptlearning/adapt_framework/issues/1732))
* add the core properties `_isVisible` and `_isHidden`
* add missing `instruction` property to article & block schemas
* improve usability through better titles/help text

see https://github.com/adaptlearning/adapt_framework/issues/1435

### DIFF
--- a/plugins/content/config/model.schema
+++ b/plugins/content/config/model.schema
@@ -237,8 +237,8 @@
       "isSetting": true,
       "inputType": "Checkbox",
       "validators": [],
-      "title": "Generate sourcemap",
-      "help": "Creates a JavaScript sourcemap for the output code - useful for debugging (in browsers that support it)."
+      "title": "Generate sourcemaps?",
+      "help": "Allows the course JavaScript & CSS to be debugged via the browser's developer tools."
     },
     "_forceRouteLocking": {
       "type": "boolean",
@@ -246,7 +246,7 @@
       "isSetting": true,
       "inputType": "Checkbox",
       "validators": [],
-      "title": "Enforce route locking",
+      "title": "Enforce route locking?",
       "help": "If locking is enabled, this setting prevents navigating to locked routes"
     },
     "_logging": {
--- a/plugins/content/config/model.schema
+++ b/plugins/content/config/model.schema
@@ -1,10 +1,10 @@
 {
-  "type":"object",
+  "type": "object",
   "$schema": "http://json-schema.org/draft-04/schema",
   "id": "http://jsonschema.net",
   "properties": {
     "_extensions": {
-      "type":"object"
+      "type": "object"
     },
     "_enabledExtensions": {
       "type":"object"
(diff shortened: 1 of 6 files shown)


moloko:
Note that most of these changes have already been reviewed, approved & merged in [this framework PR](https://github.com/adaptlearning/adapt_framework/pull/1770). 

The only changes that haven't been approved already are the amends to the help text for the `_onScreen` property

plugins/content/config/model.schema
```diff
@@ -51,17 +51,15 @@
           "isSetting": false,
           "inputType": "Checkbox",
           "validators": [],
-          "title": "Enable accessibility?",
-          "help": "If enabled, accessibility features will be enabled in this course"
```
canstudios-louisem: Is there any reason to remove the help?

plugins/content/config/model.schema
```diff
@@ -51,17 +51,15 @@
...
+          "title": "Enabled?"
         },
         "_isEnabledOnTouchDevices": {
           "type": "boolean",
           "default": false,
           "isSetting": false,
           "inputType": "Checkbox",
           "validators": [],
-          "title": "Enable accessibility on touch devices?",
-          "help": "If enabled, accessibility features will be enabled on touch devices - useful if you need accessibility on mobile browsers."
```
canstudios-louisem: Is there any reason to remove the help?

plugins/content/config/model.schema
moloko: Yes, because it's not particularly helpful in that it doesn't add any more information than you'd glean from reading the title!

plugins/content/config/model.schema
moloko: Again, it's not adding anything that you wouldn't already have gleaned from reading the title!

moloko: If feel it's equivalent to [doing this](https://github.com/adaptlearning/documentation/blob/master/01_cross_workstream/style_guide.md#comments)

plugins/content/config/model.schema
canstudios-louisem: It should probably be changed to be more useful rather than being removed such as. `Enables accessibility features in this course, such as extra tags for screen readers` 

Remember the users of the AT are often not as experienced as the framework users and being explicit even when it is super obvious to us can help.

moloko: Changing the title to 'Enable accessibility features?' would go some way towards making it clearer... but I'm not sure 'extra tags for screen readers' adds much in the way of useful info. 

I'm sure there is something better that can be written, I just don't feel qualified to write it!

How about I amend the title as suggested above and we raise another PR for 'improvements to help text'?

One other possibility would be to include a link to https://github.com/adaptlearning/adapt_framework/wiki/Accessibility in the help text

canstudios-louisem: Ok sure, that issue would be best on the framework repo I think.

PR #1731, opened 2017-09-29
Framework import

dancgray:
This PR adds framework source import functionality. A zip file downloaded from the AT "Export source code" or a zip containing a framework repository can be imported. Imports the full course and all assets and plugins. See below for asset and plugin import details.

### Multi language imports
The import will only import the first language folder it encounters. Other language folders will be ignored. The import takes the course language folder value and assigns it to config._defaultLanguage. (Note - This was done to deal with multi lang courses where the default language was not the same as the imported language)

### Plugins
Import will only import a plugin if it does not already exist. Newer versions of plugins are ignored. The AT can only deal with one theme and menu per course. The first theme and menu encountered will be assigned to the course in the AT.

### Assets
In order to reduce asset duplication imported asset title and size are compared to assets already in the AT. If a match on both is found the asset is not imported and the existing asset is used instead.

### Tags
Tags are added through the AT import screen. Tags assigned here will be added to all imported assets and to the course.

### Outstanding Tasks
- [ ]  Look at asset and plugin cleanup. Currently they are not cleaned up on failed imports
- [x]  Add tags. I would be nice to have the option to export tags when exporting source code from the AT so they can be preserved when re-imported.

### Known Issues
- [ ]  Menu position is not preserved in contentObjects.


canstudios-louisem:
Can a none super admin import and if so should they be allowed to add plugins? 

I think if a none super admin tries to import a course that would need to add a plugin it should error to the user telling them they need to ask a super admin to do this.

canstudios-louisem:
Ive merged in develop and fixed up the issues from the merge.

canstudios-louisem:
Everything seems to be working for me, im happy to approve do we want to make the change so other users can add courses but not plugins?

lc-thomasberger:
Regarding multi-lang imports, the [default language](https://github.com/adaptlearning/adapt_framework/blob/master/src/course/config.json#L2) of a course is stored in config.json
Can we use this value to load the language and fall back to the first one?

benhalverson:
@dancgray @canstudios-louisem 
I've tried this feature branch I'm getting an import failed error 
screen shot 2017-11-29 at 5 34 21 pm

dancgray:
@benhalverson thanks for testing, was there any extra information or stack trace in the terminal or server logs?

benhalverson:
@dancgray no the stacktrace had the same error as the UI about not finding a file or directory.

truonghongtrieu:
I've also tried to install a plugin with this feature branch and getting a Plugin upload failed error:

import plugin

Stacktrace on the server:
`E:\adapt\adapt_authoring>node server
(node:14444) fs: re-evaluating native module sources is not supported. If you are using the graceful
-fs module, please update it to a more recent version.
info: [14 Dec 2017 11:32:10 +08:00] configuration loaded from E:\adapt\adapt_authoring\conf\config.json
info: [14 Dec 2017 11:32:10 +08:00] Supporting the following authentication types:
info: [14 Dec 2017 11:32:10 +08:00] - local
info: [14 Dec 2017 11:32:12 +08:00] Adapt authoring tool 0.4.0.
info: [14 Dec 2017 11:32:12 +08:00] Adapt framework 2.2.3.
info: [14 Dec 2017 11:32:12 +08:00] Connection established: adapt-tenant-master
Cannot determine framework version
ReferenceError: addPackage is not defined`

I followed the stacktrace to "\plugins\content\bower\index.js", it appears that the method it not calling it on the plugin...:
`installHelpers.getInstalledFrameworkVersion(function(error, frameworkVersion) {
              if(error) {
                return next(error);
              }
              // Check if the framework has been defined on the plugin and that it's not compatible
              if (packageInfo.pkgMeta.framework && !semver.satisfies(semver.clean(frameworkVersion), packageInfo.pkgMeta.framework)) {
                return next(new PluginPackageError('This plugin is incompatible with version ' + frameworkVersion + ' of the Adapt framework'));
              }
              app.contentmanager.getContentPlugin(pluginType, function (error, contentPlugin) {
                if (error) {
                  return next(error);
                }
                addPackage(contentPlugin.bowerConfig, packageInfo, { strict: true }, function (error, results) {
                  if (error) {
                    return next(error);
                  }
                  res.statusCode = 200;
                  return res.json({ success: true, pluginType: pluginType, message: 'successfully added new plugin' });
                });
              });
            });`

canstudios-louisem:
I just had a quick test and tried to upload a download not an export and it give a unhelpfull error 
```{"success":false,"message":"ENOENT: no such file or directory, scandir '/app/temp/5a018885e49066b8019fcf2c/adapt_framework/courses/5a018885e49066b8019fcf2c/5a018885e49066b8019fcf2c_unzipped/src/course'"}```

Is there an easy check we can do to ensure its an export and not just any old zip file the user has uploaded the error message should probably say something about import source not built courses as well.

**Comment by canstudios-louisem** [comment]
The rest of the import worked fine for me FYI. Have we had any more thoughts on adding some kind of metadata file to the export of the form our content team use the custom less quite a bit and being able to copy that over easily would be great.

**Comment by canstudios-louisem** [comment]
I tested on a server where the plugin and framework versions don't match the server the course came from and I get the following error.

```
1|Communit | error: [02 Feb 2018 10:50:35 +00:00] Unable to import course: Cannot determine framework version
1|Communit | ReferenceError: pluginVersion is not defined
```

**Comment by lc-thomasberger** [comment]
When I import a course with a custom component, I do get an Error in [editorPageComponentView](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/modules/editor/contentObject/views/editorPageComponentView.js#L143):
Uncaught TypeError: Cannot read property 'get' of undefined

[_supportedLayout](https://github.com/LearnChamp/adapt-svgWrapper/blob/master/properties.schema#L9-L15) is defined though.

**Comment by lc-thomasberger** [comment]
When I do a Framework export, I do get the following Error in the node console: TypeError: fs.ensureDir is not a function

**Comment by lc-thomasberger** [comment]
Import of the assets seems to be fine, they appear in AssetManager. But when I build the course with imported images, the images are not available in the course.

**Comment by lc-thomasberger** [comment]
@dancgray and @canstudios-louisem I think the error `{"success":false,"message":"ENOENT: no such file or directory, scandir '/app/temp/5a018885e49066b8019fcf2c/adapt_framework/courses/5a018885e49066b8019fcf2c/5a018885e49066b8019fcf2c_unzipped/src/course'"}` is thrown when a Course with "custom" asset folder is imported. For instance I used `course/en/images` as in the Framework example code.

**Comment by canstudios-nicolaw** [comment]
Which version of the fs-extra module do you have locally @dancgray ? 

It seems the ensureDir function is part of that module but the fs-extra version in the authoring tool package.json is quite low. I wonder if bumping this up to a higher version would fix the issue Thomas mentioned above about the export not working.

**Comment by canstudios-nicolaw** [comment]
It looks like the ensureDir not a function is caused because the output/adapt/index.js file has

```
fs = require('fs'),
fse = require('fs-extra'),
```

And then later calls `fs.ensureDir` instead of `fse.ensureDir`

**Comment by taylortom** [comment]
Fixed conflicts and added glob to package.json (couldn't figure out if this was broken in the merge).

**Comment by taylortom** [comment]
Thanks for pushing those latest changes Dan, we've got some time booked for a bit of testing tomorrow.

A couple of requests from my initial test:
- Can we combine all the inputs into a single `form-group` div? The spacing between each section looks a bit odd
- I wondered if we could style the title/help text for the inputs to match the scaffold forms for consistency (not sure how straightforward this will be due to some of the weird nested LESS we have, so no worries if not) 

Screenshots for comparison: 

screen shot 2018-03-19 at 11 10 45

screen shot 2018-03-19 at 11 09 27

**Comment by taylortom** [comment]
Cheers for doing that Dan, looks :ok_hand:

**Comment by taylortom** [comment]
A quick thought on the labels we use: as we're combining the export packages into one, should we switch to more generic language rather than referencing the source code/framework?

**Global menu**<br/>_Framework Import_ > _Course Import_

**Editor sidebar**<br/>_Export source code_ > _Export course_<br/>(this one could perhaps be renamed simply _Export_, and styled more like the preview/download buttons now it's a 'proper' feature)

...

There may be other areas these are referenced

File: plugins/output/adapt/helpers.js [hunk 1]
```diff
@@ -0,0 +1,276 @@
...
+
+          var serverVersion = semver.clean(frameworkVersion);
+          var pluginRange = semver.validRange(bowerJson.framework);
+
+          if(semver.satisfies(serverVersion, pluginRange)) {
+            logger.log('info', 'Installing', pluginType, "'" + bowerJson.displayName + "'");
+            bowerJson.isLocalPackage = true;
+            app.bowermanager.addPackage(contentPlugin.bowerConfig, { canonicalDir: pluginDir, pkgMeta: bowerJson }, { strict: true }, cb);
+          } else {
+            logger.log('info', "Can't install " + bowerJson.displayName + ", it requires framework v" + pluginVersion + " (" + frameworkVersion + " installed)");
```
taylortom: Getting a `ReferenceError` related to `pluginVersion` here. Should this be `pluginRange` instead?

dancgray: That is correct, just changed 👍

taylortom:
I can't remember where this was discussed, but what do we think about the terms 'export' and 'download' -- we've had feedback that these are too similar, and could do with being more self-explanatory. I think 'export' is fine, but 'download' could be better.

canstudios-louisem:
Download source or similar would be a lot more explicit. If we did that changing import to import source would also help reduce confusion.

lc-thomasberger:
Just did a retest on Windows and it looks very good!
I just wonder why the export / import is limited to Admins. I do understand that the "Framework export" was sort of a power user feature. However I think import / export will also be used by a regular course creator. What do you guys think about this?

canstudios-nicolaw:
@lc-thomasberger I think part of the reason for limiting import to admins was because plugins can be uploaded to the tool as part of the import process. Normally course creators would not be able to install plugins.

canstudios-louisem:
@lc-thomasberger the issue is it can be used to add plugins hence why it was limited to super admins. Having a none super admin version that will not add plugins was discussed but it was not MVP.

lc-thomasberger:
Thanks @canstudios-louisem @canstudios-nicolaw yes that makes sense.

taylortom:
@dancgray have fixed the latest conflicts.

Is it possible to get the hero image working for exports? Part of the problem seems to be that we filter out any attributes with `editorOnly: true` as a part of the build process ([see here](https://github.com/adaptlearning/adapt_authoring/blob/master/lib/dml/mongoose/index.js#L482-L486)). We'll also need to do something about the value, as it's stored as just an ID, rather than tha `course/asset/blah` string that's looked up in the import.

I've also had some feedback to move the import button to the main projects page as it's not all that obvious to find. I've already got something working like the below and can commit this if people like it.

screen shot 2018-04-13 at 11 12 59

dancgray:
@taylortom, thanks for that. I like the idea of moving import button to the main projects page, keeps it simple. 
Were you thinking editorOnly functionality should remain for download but change selectively for export source?

taylortom:
Not sure really, I wouldn't have a problem with including all editorOnly attributes really, they shouldn't interfere with anything else. Either way, I think we should probably to the attribute 'trimming' in a more transparent way (although this could wait until the server refactor).

canstudios-louisem:
We have agreed to change the wording of export, download and import to Publish Course, Export Source and Import Source

plugins/output/adapt/importsource.js
```diff
@@ -0,0 +1,640 @@
...
+            if (!fs.existsSync(assetFolderPath)) {
+              assetFolderError = true;
+              missingFolders.push(assetFolders[index]);
+            }
+          }
+          // if a user input folder is missing log error and abort early
+          if (assetFolderError) {
+            var folderError = 'Cannot find asset folder/s ' + missingFolders.toString() + ' in framework import.';
+            return cb(folderError);
+          } else {
```
tomgreenfield: No need for `else` section due to the `return`.

plugins/output/adapt/importsource.js
```diff
@@ -0,0 +1,640 @@
...
+            if(error) {
+              logger.log('error', error)
+              return cb2(error);
+            }
+            cb2();
+          });
+        }, cb);
+      }],
+      checkAssetFolders: ['checkContentJson', function(results, cb) {
+        if (cleanFormAssetDirs.length) {
```
tomgreenfield: Would you mind reversing the logic for readability?
```js
if (!cleanFormAssetDirs.length) {
  assetFolders = Constants.Folders.ImportAssets;
  return cb();
}

// rest of function
```

plugins/output/adapt/importsource.js
```diff
@@ -0,0 +1,640 @@
...
+  /**
+  * Installs any custom plugins
+  */
+  function installPlugins(done) {
+    async.series([
+      function mapPluginIncludes(cb) {
+        async.each(plugindata.pluginTypes, function iterator(pluginType, doneMapIterator) {
+
+          var srcDir = path.join(COURSE_ROOT_FOLDER, Constants.Folders.Source, pluginType.folder);
+          if (fs.existsSync(srcDir)) {
```
tomgreenfield: Probably another good candidate for returning early? I get a bit lost in the 7/8 levels of indentation.
```js
if (!fs.existsSync(srcDir)) {
  logger.log('info', 'No plugins found.');
  return doneMapIterator();
}

// fs.readdDir...

File: plugins/output/adapt/index.js [hunk 5]
```diff
@@ -55,290 +56,617 @@ AdaptOutput.prototype.publish = function(courseId, mode, request, response, next
...
+  fs.readJson(path.join(frameworkDir, Constants.Filenames.Package), function onJsonRead(error, packageJson) {
+    gotPackageJson(null, _.pick(packageJson, 'version'));
+  });
+};
+
+// rips all course data from the DB
+function getCourseMetdata(courseId, metadata, gotCourseMetadata) {
+  database.getDatabase(function(error, db) {
+    if (error) {
+      return callback(error);
```
tomgreenfield: Is `callback()` defined anywhere?

plugins/output/adapt/index.js
```diff
@@ -55,290 +56,617 @@ AdaptOutput.prototype.publish = function(courseId, mode, request, response, next
...
-        var pluginsRE = new RegExp('\/components\/|\/extensions\/|\/menu\/|\/theme\/');
+/**
+* Copy functions
+*/
+
+// copies relevant files in adapt_framework
+function copyFrameworkFiles(results, filesCopied) {
+  self.generateIncludesForCourse(courseId, function(error, includes) {
+    if(error) {
+      return includesGenerated(error);
```
tomgreenfield: Is `includedGenerated()` defined anywhere?

frontend/src/modules/frameworkImport/templates/frameworkImport.hbs
```diff
@@ -0,0 +1,41 @@
...
+          <label for="file">
+              {{t 'app.assetfile'}} <span class="req">*</span>
+              <span class="error"></span>
+          </label>
+          <div class="field-help">
+            {{t 'app.importframeworkinstruction'}}
+          </div>
+          {{#if path}}
+          <div class="asset-management-new-asset-file">
+              File has been uploaded
```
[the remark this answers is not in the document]
lc-thomasberger: at least it should use polyglot.

frontend/src/modules/editor/global/views/editorView.js
```diff
@@ -115,6 +114,58 @@ define(function(require){
...
+           Origin.Notify.alert({
+             type: 'error',
+             title: Origin.l10n.t('app.exporterrortitle'),
+             text: messageText
+           });
+         }
+      });
+    },
+
+    showExportAnimation: function(show, $btn) {
```
tomgreenfield: This function also exists on line 252?

routes/lang/en-application.json
```diff
@@ -282,6 +282,17 @@
...
   "app.emaillogotext": "ADAPT",
   "app.emailfooter": "Sent by Adapt Learning.",
+  "app.importing": "Importing...",
+  "app.importframeworkfileuploaded": "File has been uploaded",
+  "app.importerrortitle": "Course import failed",
+  "app.importcoursesuccess": "Course imported successfully!",
+  "app.importcourse": "Import source",
+  "app.frameworkimporttitle": "Import framework source",
+  "app.importframeworkinstruction": "Select a framework zip for import. Tags will be added to the course and all new assets.",
+	"app.importassetfoldersinstruction": "Enter the names of folders that contain assets. If there are multiple folders enter a comma seperated list. Automatically checks for the following folders if no value is entered: assets, images, video, audio.",
```
tomgreenfield: Tab → spaces.

PR #1733, opened 2017-10-03
Fix invalid zip issue with plugin upload

### LINKED ISSUE #1732: Upload of Bower plugin fails with error msg "invalid signature: 0x20200a7b"
When trying to upload a Authoring Tool plugin packed by the grunt-contrib-compress task it fails with the rather unhelpful error message "invalid signature: 0x20200a7b".
This seems to be caused by unzip not handling the "store" attribute on the zip archive correctly.

Steps to reproduce:
 - Repack any Bower plugin using grunt-contrib-compress using these options:
     `options: {
                    mode: "zip",
                    archive: "./plugin_name.zip",
                    store: true
                }`

- Use the Authoring Tool to install the resulting archive via the "Upload Plugin" functionality.

Related issue: #920

UlfChristensen:
Fixes #1732

related: https://www.npmjs.com/package/unzip2


taylortom:
Thanks for this.

We actually moved over to unzip2 at some point in the past, but we ended up going back for reasons I can't remember -- the issues may have been resolved in later unzip2 releases, so definitely worth testing again.

Can you also switch the base to develop please?

UlfChristensen:
Base changed - I was sure that I did that when I made the request but I must have missed it....

Issue #1486, opened 2017-01-19
Menu and page editor should display the course/page title


dancgray:
It's not easy to tell which course/page you are editing from these screens, would be useful to orientate the user within the tool.

hmaffin:
Totally agree - was thinking this in the last few days especially after editing the wrong page - d'oh!

taylortom:
I've implemented a version of the breadcrumb trail as was included in the updated AD:

![screen shot 2017-01-21 at 17 10 36](https://cloud.githubusercontent.com/assets/1059083/22176200/c0827398-dffc-11e6-9a91-825cba3f1577.png)

The breadcrumb segments describe the page, while the page title is set to the title. Each segment is also a clickable link.

PR #1747, opened 2017-10-19
Add clickable breadcrumb trail

taylortom:
#1486


frontend/src/modules/projects/index.js
```diff
@@ -63,16 +61,14 @@ define(function(require) {
...
-        Origin.trigger('location:title:update', {title: 'Dashboard - viewing shared courses'});
-        Origin.contentPane.setView(ProjectsView, { collection: new SharedProjectCollection });
-        break;
-      case 'all':
-        Origin.trigger('location:title:update', {title: 'Dashboard - viewing my courses'});
-        Origin.contentPane.setView(ProjectsView, { collection: new MyProjectCollection });
-      default:
-        break;
+    if(options.type === 'shared') {
+      Origin.trigger('location:title:update', { breadcrumbs: ['dashboard'], title: Origin.l10n.t(langKey) });
```
tomgreenfield: Not sure where `langkey` has come from?

Is this meant to be `'app.sharedprojects'` and `'app.myprojects'` respectively?

frontend/src/modules/editor/contentObject/index.js
```diff
@@ -49,33 +49,34 @@ define(function(require) {
...
   function renderMenuStructure(data) {
-    Origin.trigger('location:title:update', { title: 'Menu editor' });
+    Helpers.setPageTitle(Origin.editor.data.course, true);
 
     Origin.editor.currentContentObjectId = data.id;
     Origin.editor.scrollTo = 0;
 
     Origin.sidebar.addView(new EditorMenuSidebarView().$el, {
-      "backButtonText": "Back to courses",
+      "backButtonText": Origin.l10n.t('app.backtoprojects'),
```
tomgreenfield: String not present in `en-application.json`.

frontend/src/modules/editor/global/templates/partials/part_editorCommon.hbs
```diff
@@ -33,39 +33,39 @@
...
+        <i class="fa fa-chevron-right primary-color"/>{{ t 'app.editorconfig'}}
       </span>
     </button>
   </div>
 
   <div class="sidebar-row">
     <button class="editor-common-sidebar-select-theme">
       <span class="editor-common-sidebar-select-theme-inner">
-        <i class="fa fa-chevron-right primary-color"/></i>{{ t 'app.themepicker'}}
+        <i class="fa fa-chevron-right primary-color"/></i>{{ t 'app.editorselecttheme'}}
```
tomgreenfield: String not present in `en-application.json`.

Should be left as `app.themepicker`?

frontend/src/modules/editor/global/helpers.js
```diff
@@ -1,44 +1,86 @@
...
+    var modelTitle = data.model.title || data.model.get && data.model.get('title');
+    return modelTitle || data.langString;
+  }
+
+  function getLangKey() {
+    var type = getType();
+    var action = getAction();
+    switch(type) {
+      case 'page':
+        if(action === 'edit') return 'editor'+type+'settings';
```
tomgreenfield: `app.editorpagesettings` is missing from `en-application.json`.

It's currently present as `app.editorpageedit`.

frontend/src/modules/assetManagement/index.js
```diff
@@ -61,4 +18,57 @@ define(function(require) {
...
+        });
+      }
+    });
+  }
+
+  function loadAssetView(id) {
+    var isNew = id === undefined;
+    var model = new AssetModel(isNew ? {} : { _id: id });
+
+    if(isNew) loadView();
```
tomgreenfield: Curly braces?

frontend/src/modules/location/less/location.less
```diff
@@ -3,14 +3,42 @@
...
+      }
+      .separator {
+        display: inline-block;
+        margin: 0 5px;
+        opacity: 0.35;
+        overflow: hidden;
+      }
+    }
+    .title {
+
```
tomgreenfield: This empty selector meant to stay in?

frontend/src/modules/projects/index.js
```diff
@@ -63,16 +61,14 @@ define(function(require) {
...
+      Origin.contentPane.setView(ProjectsView, { collection: new SharedProjectCollection });
+      return;
+    }
+    if(options.type === 'all') {
```
tomgreenfield: Reverse logic and return early?

```js
if (options.type !== 'all') return;
```

frontend/src/modules/editor/global/helpers.js
```diff
@@ -1,44 +1,86 @@
...
+        title: Origin.l10n.t('app.editorpage'),
+        url: '#/editor/' + page.get('_courseId') + '/page/' + page.get('_id')
+      });
+    }
+    crumbs.push({ title: data.langString });
+    return crumbs;
+  }
+
+  function getTitleForModel(data) {
+    var type = getType();
```
tomgreenfield: `type` not needed here.

frontend/src/modules/editor/contentObject/index.js
```diff
@@ -49,33 +49,34 @@ define(function(require) {
...
       currentCourseId: Origin.location.route1,
       currentView: 'page',
-      currentPageId: (data.id || null)
+      currentPageId: data.id
     });
   }
 
   function renderMenuStructure(data) {
-    Origin.trigger('location:title:update', { title: 'Menu editor' });
+    Helpers.setPageTitle(Origin.editor.data.course, true);
```
tomgreenfield: Second parameter `true` not needed?

lc-thomasberger:
I do get an error with `langKey` being undefined in: https://github.com/adaptlearning/adapt_authoring/blob/issue/1486/frontend/src/modules/projects/index.js#L65

frontend/src/modules/location/views/locationTitleView.js
```diff
@@ -1,41 +1,63 @@
...
+    initialize: function() {
+      this.listenTo(Origin, {
+        'location:title:update': this.render,
+        'location:title:hide': this.onHideTitle
+      });
+      EditorDataLoader.waitForLoad(_.bind(this.render, this));
+    },
+
+    render: function(data) {
+      var template = Handlebars.templates[this.constructor.template];
```
lc-thomasberger: Can we include the elements title if available?

image

If it is too long, we can also set it as a `title` attribute to show it on hover.

taylortom:
@tomgreenfield, @lc-thomasberger I tried to go with a solution that included both titles and page type, so you know where you are in terms of the current page _type_ (e.g. 'Block settings'), as well as the current instance (e.g. 'My block title'). I chose to swap out the course title to save screen space (@lc-thomasberger I think this also answers your suggestion?).

As you've noticed, this doesn't work in all scenarios, so would be grateful for any possible solutions. The boxes I think we need to tick are:
- Include the current page _route_ (i.e. Dashboard  > Course Structure  > Page structure  > Block settings)
- Include titles where useful < this one's up for grabs, I think at the very least we need the current course title, and the current page title
- Doesn't overrun on small monitors (not sure what's sensible to use as a limit here)
- Doesn't take up too much vertical space

A potential solution could be to _not_ include the course title in the breadcrumb (using 'Course structure' instead), and including the course title on the second line, something like: 

Dashboard  > Course Structure  > Page structure  > Block settings
**My First Course**: Block Title

lc-thomasberger:
We can limit the width of each breadcrumb Segment and use ellipses to take care with overflowing content. This would be a CSS only solution.
You might also add a title attribute to each breadcrumb item and set it to the elements title. If a user hover over the item, the title will be displayed.

tomgreenfield:
> Include titles where useful < this one's up for grabs, I think at the very least we need the current course title, and the current page title

Maybe it's easier to ditch this requirement in the trail itself. The breadcrumbs are still extremely useful for seeing how generically 'deep' you are in the course structure, and for skipping back to the desired editor.

taylortom:
That's fine with me!

I think it's important that we show the current course title somewhere at least -- this was something that came back from Kineo QA on various occasions, and have also heard similar things from the community.

Maybe the breadcrumb/title area isn't the place for the course title?

taylortom:
Have updated to use two separate lines:

screenshot 2019-02-08 at 12 02 16

We have the possibility of overlap between these titles and the options buttons, but I think this is something that needs to be fixed separately -- we'll have to mess around with the structure of the DOM elements to get this working properly.

frontend/src/modules/editor/contentObject/index.js
```diff
@@ -24,45 +24,46 @@ define(function(require) {
...
     Origin.contentPane.setView(EditorPageEditView, { model: data.model, form: form });
   }
 
   function renderPageStructure(data) {
-    Origin.trigger('location:title:update', { title: 'Page editor' });
+    Helpers.setPageTitle(data.model);
 
     Origin.sidebar.addView(new EditorPageSidebarView().$el, {
-      "backButtonText": "Back to course structure",
+      "backButtonText": Origin.l10n.t('app.backtomenu'),
```
tomgreenfield: No need for quotes on keys?

frontend/src/modules/editor/contentObject/index.js
```diff
@@ -24,45 +24,46 @@ define(function(require) {
...
 
   function renderMenuStructure(data) {
-    Origin.trigger('location:title:update', { title: 'Menu editor' });
+    Helpers.setPageTitle(Origin.editor.data.course);
 
     Origin.editor.scrollTo = 0;
 
     Origin.sidebar.addView(new EditorMenuSidebarView().$el, {
-      "backButtonText": "Back to courses",
+      "backButtonText": Origin.l10n.t('app.backtoprojects'),
```
tomgreenfield: No need for quotes on keys?

frontend/src/modules/location/templates/locationTitle.hbs
```diff
@@ -1,3 +1,31 @@
 <div class="location-title-inner">
-	{{title}}
+	{{#if breadcrumbs}}
```
tomgreenfield: As much as I ❤️ tabs, the style guide dictates that indentation should be two spaces in this file.

frontend/src/modules/location/views/locationTitleView.js
```diff
@@ -1,41 +1,66 @@
...
-		},
+    processData: function(data) {
+      // nothing to do
+      if(!data || !data.breadcrumbs) {
+        return data;
+      }
+      // add some shortcuts to common locations
+      var course = Origin.editor.data.course;
+      // Dashboard
+      var dashboardI = _.indexOf(data.breadcrumbs, 'dashboard');
```
tomgreenfield: Can use native `.indexOf()`s here.

routes/lang/en-application.json
```diff
@@ -148,7 +148,8 @@
   "app.projectsettings": "Project settings",
   "app.projectcontentbody": "This is a text component",
   "app.configurationsettings": "Configuration settings",
-  "app.themepicker": "Theme picker",
+  "app.themesettings": "Theme settings",
```
tomgreenfield: Probably need to revert this line:

picker

taylortom:
@tomgreenfield have actioned your amends, with the exception of your first bullet.

There's a bit of inconsistency with the way we reference the dash/courses (both in naming and how we navigate there...). I propose we put together a separate fix to rectify this which does the following:
- Renames all dashboard references to 'courses' or 'projects' (including the global menu button)
- Remove all references to the dashboard from projects module code
- Make sure all code uses the generic dashboard functions to get to the dashboard, as opposed to specifying the route manually*

*As part of the front-end refactor I added the ability for modules to register as the 'home' route so code can use generic `navigateToHome` functions rather than manually setting the home route.

frontend/src/modules/location/templates/locationTitle.hbs
```diff
@@ -1,3 +1,31 @@
...
+          {{else}}
+            {{#if url}}
+              <a href="{{url}}">{{title}}</a>
+            {{else}}
+              {{title}}
+            {{/if}}
+          {{/if}}
+        </div>
+        {{#if @last}}
+        {{else}}
```
tomgreenfield: Could switch to `{{#unless @last}}`?

PR #1750, opened 2017-10-25
Updates from master

taylortom:
Not sure why these were missed in the previous PR


canstudios-louisem:
@taylortom can this be merged now?

PR #1768, opened 2017-10-30
Stop throwing error if repo update check failed

### LINKED ISSUE #1763: Server should handle 'check for updates' errors more cleanly
## Expected Behaviour
Should just have a quiet 'cannot check for updates' message, rather than causing errors

## Actual Behaviour
Error thrown.

## Steps to Reproduce
1. Start the server without an internet connection

## Versions

taylortom:
Fixes #1763

### DIFF
--- a/lib/installHelpers.js
+++ b/lib/installHelpers.js
@@ -166,10 +166,10 @@ function getUpdateData(callback) {
       return callback(error);
     }
     var updateData = {};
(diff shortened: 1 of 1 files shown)


dancgray:
@taylortom This is still falling over with no internet. Is that expected?

taylortom:
Apologies @dancgray, have amended a broken if

PR #1771, opened 2017-11-01
Article view now generated after article model saved

### LINKED ISSUE #1770: Cannot delete empty articles
## Expected Behaviour
Removing an empty article (by clicking the trash icon) would remove the article if alert is confirmed.

## Actual Behaviour
Article is not removed. No console errors in browser or terminal. Delete using cog icon works.

## Steps to Reproduce
1. Create a new course
2. Add a new article that does not contain a component
3. Try to remove article using the trash icon.

## Versions
  - Authoring Tool Version: develop b ranch
  - Framework Version: v2.2.2
  - Operating System: MacOS 10.12.6
  - Browser: Chrome 62

canstudios-louisem:
Fixes #1770 

The View was being generated before the article was saved and so had no ID this caused events to be named incorrectly.

### DIFF
--- a/frontend/src/modules/editor/contentObject/views/editorPageView.js
+++ b/frontend/src/modules/editor/contentObject/views/editorPageView.js
@@ -127,8 +127,6 @@ define(function(require){
         _type:'article'
       });
 
-      var newArticleView = _this.addArticleView(newPageArticleModel);
-
       newPageArticleModel.save(null, {
         error: function() {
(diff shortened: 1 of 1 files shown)


canstudios-louisem:
The tests failed the first time they were ran on node 4 I'm not sure why I re-ran them and they worked. Travis was reporting some issues at the time so that may have been the cause.

PR #1777, opened 2017-11-08
Fix error when deleting a shared course that isn't yours

### LINKED ISSUE #1776: Deleting a shared course when you are not then owner gives an error
## Expected Behaviour
When deleting a shared course I should be warned right away I cant 

## Actual Behaviour
When you attempt to delete a shared course you have to click on the initial popup and wait the 5 seconds before being told you cant delete it.

## Versions

canstudios-louisem:
Fixes #1776


frontend/src/modules/projects/views/projectView.js
```diff
@@ -76,14 +76,23 @@ define(function(require) {
...
+            type: 'warning',
+            title: Origin.l10n.t('app.deletesharedproject'),
+            text: Origin.l10n.t('app.confirmdeleteproject') + '<br/><br/>' + Origin.l10n.t('app.confirmdeletesharedprojectwarning'),
+            destructive: true,
+            callback: _.bind(this.deleteProjectConfirm, this)
+          });
+        } else {
+          Origin.Notify.alert({
+            type: 'error',
+            title: Origin.l10n.t('app.error'),
```
taylortom: There shouldn't be any need for this, as a default title is added here:
- [modules/notify/plugins/alert/index.js#L32](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/modules/notify/plugins/alert/index.js#L32)
- [en-application.json#L248](https://github.com/adaptlearning/adapt_authoring/blob/master/routes/lang/en-application.json#L248)

PR #1794, opened 2017-11-24
Fix asset management pagination

taylortom:
Also gave the files I touched a bit of a tidy.

### DIFF
--- a/frontend/src/modules/assetManagement/index.js
+++ b/frontend/src/modules/assetManagement/index.js
@@ -10,39 +10,12 @@ define(function(require) {
   var TagsCollection = require('core/collections/tagsCollection');
 
   Origin.on('router:assetManagement', function(location, subLocation, action) {
-    Origin.assetManagement = {};
-    Origin.assetManagement.filterData = {};
-
-    if (!location) {
-      (new TagsCollection()).fetch({
-        success: function(tagsCollection) {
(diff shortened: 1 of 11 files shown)


dancgray:
@taylortom for me the filters are not working. Tested with a very large collection, no errors returned.

taylortom:
Ok cheers, will have a lookie

lc-thomasberger:
Pagination works for me, but filters are broken as well. Tested with 38 elements inside the assets collection.

lc-thomasberger:
edit seems to be a bit broken as well.

taylortom:
Thanks @lc-thomasberger, your changes work for me! :)

dancgray:
The asset picker modal that you get when choosing an asset for a plugin does not initially show all assets. Resizing causes all assets to render.

Tag and search filter works though :)

taylortom:
Thanks Dan, those should be sorted now

taylortom:
Just fixed another issue with the modal's scrollbar being shown when it shouldn't (the asset management modal has its own scrollbar)

dancgray:
@taylortom works for me. Does not seem to be refreshing the collection of assets if you upload through the modal window though. I'm sure it used to highlight the image you just uploaded.

taylortom:
@canstudios-louisem, @dancgray @lc-thomasberger This is ready to test again :)

PR #1798, opened 2017-11-27
Editor performance improvements

### LINKED ISSUE #1758: largely populated AT instance running slow after v0.4 upgrade
Since we upgraded to v0.4 the AT has been running really slow (like takes 15s to load a course when it would previously take 2s).

This was never an issue on my local installation or our dev installation, which runs of the same server and config as the live one.

## Versions
  - Authoring Tool Version: 0.4
  - Framework Version: 2.2.2
  - Operating System: Unix based server
  - Node version: 6.6.0

taylortom:
Have removed much of the shared data from the editor (which was preloaded at run-time), and replaced with a more ad-hoc method, which only fetches the relevant data as it's needed.

Fixes #1758.

### Main points
- Removes global course data in `Origin.editor.data`:
    - contentobjects
    - articles
    - blocks
    - components
    - courseassets
    - clipboards
- Model data is now accessed in a more ad-hoc way when it's needed, rather than pre-loading everything at once
- ContentModels no longer do a compulsory fetch in their initialize function, which was causing lots of unnecessary requests
- Course data models have been reworked to work asynchronously, pay particular attention to: `getParent`/`getChildren`/`getSiblings` which are now `fetchParent`/`fetchChildren`/`fetchSiblings` respectively (ContentModel.js)
- Have slightly reworked any files I've touched where I thought it was needed:
    - Fixed major whitespace issues (incorrect indentation/extra line-breaks etc.)
    - Some files were very confusing, so have moved some functions around to make everything more readable
    - Have removed unused/redundant code where I spotted it

### Known issues
- ~I don't like the way the page animates gradually/blocks pop into view, so going to look into rendering something (perhaps setting a min-height on the block) to show something's happening without the pop~
- ~Louise's preview popup blocker fix is broken due to async course validation~
- ~Pasting within a page causes a re-render of the entire page~
- ~Issue with menu editor not rendering layers correctly. For example:~
    - ~Add a page/sub-menu to a sub-menu~
    - ~Click on the new item~
    - ~Sub-menu items disappear~
    - ~Refreshing the page shows all missing items as expected~
- ~Getting a `TypeError` in EditorPageComponentView (when `getSupportedLayout` is called by `evaluateLayout`)~
- ~CKEditor errors when loading project settings for the second time (causing a broken body input)~
- Need to replace all notify calls using 'xxxxx'


taylortom:
Think I'm pretty much there with this, would appreciate some testing :)

frontend/src/modules/scaffold/views/scaffoldAssetView.js
```diff
@@ -1,342 +1,307 @@
...
+        console.log(courseassets.destroy);
+        // courseassets.destroy({
+        //   success: _.bind(function(success) {
+        //     this.saveModel();
+        //   }, this),
+        //   error: function(error) {
+        //     console.error('Failed to destroy courseasset record', courseasset.get('_id'));
+        //   }
+        // });
+      });
```
tomgreenfield: Need a `.bind(this)` here.

frontend/src/modules/scaffold/views/scaffoldAssetView.js
```diff
@@ -1,342 +1,307 @@
...
+      }, function(error, courseassets) {
+        if(error) {
+          return console.error(error);
+        }
+        if(courseassets.length === 0) {
+          this.setValue('');
+          this.saveModel();
+          return;
+        }
+        console.log(courseassets.destroy);
```
tomgreenfield: Console log?

frontend/src/modules/scaffold/views/scaffoldAssetView.js
```diff
@@ -1,342 +1,307 @@
...
+      this.fetchCourseAsset({
+        _fieldName: this.value.split('/').pop(),
+        _contentTypeId: Origin.scaffold.getCurrentModel().get('_id')
+      }, function(error, collection) {
+        if(error) {
+          console.error(error);
+          return _renderDelegate();
+        }
+        if(collection.length === 0) {
+          console.warn('No asset data found for', Origin.scaffold.getCurrentModel().get('_id'));
```
tomgreenfield: This warning fires a lot; maybe every time there is an asset field without one selected.

taylortom:
All know issues fixed, please retest when you have time :)

canstudios-louisem:
I've had a test on a copy of our big server and it seems a lot better navigating around can still feel a little "slow" but I'm talking 0.5-1 second and I'm happy enough for it to go into an RC.

taylortom:
Awesome, thanks @canstudios-louisem. 

I'd be interested to have a closer look at that and try and figure out where the remaining bottlenecks (not necessary for this release though if we're happy with the improvement...).

PR #1378, opened 2016-09-12
Persist dashboard prefs & add browser storage plugin

taylortom:
Prototype plugin to utilise WebStorage. Allows saving of data to both session and local storage (fixes #1337).

Includes functionality to save dashboard view to localStorage & sort/search/filtering to sessionStorage (fixes #1336).

Saves data in the WebStorage per user ID, so you get something like:

```
USER_ID: {
   "prefsKey": "value"
}
```

or in our dashboard example:

```
578e4dae41a8045c06f20ce0: {
   "dashboard": {
      "layout": "grid"
   } 
}
```

PR #1809, opened 2017-12-05
Minimise articles

### LINKED ISSUE #1483: Enhancement: Minimize articles
I have some pages which use block slider so the articles are many blocks long. If I have a few of these on the page the page can get very long and finding what I need is getting tricky.

I suggest that the article bar has an icon on it (+ and -) that allows you to minimize the article to a bar so you can work on other articles while leaving those at a minimum intrusion on the page.

Hope that makes sense!

lc-thomasberger:
Fixes #1483


taylortom:
Think we should push this one to the next release for testing reasons

lc-thomasberger:
@taylortom fine for me. Maybe with some more minor UI enhancements?

frontend/src/core/models/articleModel.js
```diff
@@ -6,7 +6,12 @@ define(function(require) {
     urlRoot: '/api/content/article',
     _parent: 'contentObjects',
     _siblings: 'articles',
-    _children: 'blocks'
+    _children: 'blocks',
+
+    defaults: {
+      _isCollapsed: false
```
canstudios-louisem: Can you confirm this is not saved to the DB and therefore is exported on build?

canstudios-louisem:
Are all blocked collapsed by default? if so could we have a button that expands and collapses them all? It would make it a bit easier if you are searching for a specific thing.

frontend/src/core/models/articleModel.js
lc-thomasberger: It is not saved in the DB and not exported. This is the article.json from an exported course:
```
{
    "_id": "5a7d5be53fbfedfc836c023d",
    "_type": "article",
    "_parentId": "5a7d5be53fbfedfc836c023c",
    "_onScreen": {
      "_isEnabled": false,
      "_classes": "",
      "_percentInviewVertical": 50
    },
    "_isVisible": true,
    "_isHidden": false,
    "_isAvailable": true,
    "_isOptional": false,
    "instruction": "",
    "_classes": "",
    "body": "",
    "displayTitle": "collapsed article ",
    "title": "collapsed article "
  }
```

lc-thomasberger:
@canstudios-louisem Default is not collapsed. When a user collapses an article, it only sets the height of the article and uses overflow hidden to hide the blocks.
The "collapsing" is only applied as long as you work in a course and not synced with other users. If you switch pages inside one course, the settings for collapsed is saved. If you leave the course, the settings are lost.

dancgray:
Nice feature, if the collapse toggle was a button outside the context menu it would halve the number of clicks required.

taylortom:
I agree with dan -- for usability, I think we should look at perhaps adding a specific minimise button

lc-thomasberger:
That's true @dancgray and @taylortom . Contextmenu will be pretty "crowded". Will add a button.

lc-thomasberger:
When I added the button I had to update the "toolbar". I've used flexbox for this and was surprised how well it worked in IE. This refactor should make future amends easier. 

image

lc-thomasberger:
Please note that I update options module to reset the selected element. `Origin.trigger('options:reset:ui', 'optionGroup');` I should have made a single PR for this ...

taylortom:
Wasn't convinced about this initially, but works well, and think people will find it really useful. A few suggestions:
- I'm not sure about the icons; these are usually used for full-screen/non-full-screen. What do you think of either a [plus](https://fontawesome.com/icons/plus?style=solid)/[minus](https://fontawesome.com/icons/window-minimize?style=solid) or [window-maximise](https://fontawesome.com/icons/window-maximize?style=regular)/minus?
- Can we change the tooltip on hover to show expand/collapse depending on the state? (it just shows collapse currently)
- I'm not sure about the collapsed styling (specifically being able to see some of the inner block). My initial thought is to just show the title bar -- I'm willing to be persuaded on this one, as I see arguments for both!

lc-thomasberger:
Thanks @taylortom 
I think window-maximise is too complicated. Thought about it initially but didn't like it. I also tried a [arrow](https://fontawesome.com/icons/angle-down?style=solid) and rotate it to indicate the states. Plus / Minus would make sense as it is used in Accordion.

Options would need some improvements. I think we should use flexbox to align the items. This will make layouts easier and we should be able to add a title to groups.

Tooltip based on state should be fine.

I get your point about the last one. I wanted to show that the blocks are still there and not "deleted". I may play with some styles and post screenshots in gitter.

taylortom:
Just noticed github have very similar functionality in their diff stuff (click the 'Files changed' tab up at the top of this page). They use up/down chevrons, and only show the title bar.

dancgray:
Works well for me, but I can imagine people will want the minimised articles preserved after editing an article/block/component on the screen.

lc-thomasberger:
@dancgray yes, there was an update with the editor data. The models are now refetched whenever something is edited. We could store the toggled state in the db, which will apply the selection to every other user or use a model to store the toggle state for each element and apply it once they are rendered.

For now, I added the global toggle buttons at the top. This makes it easier to collapse all articles at once.

taylortom:
What about storing the prefs in BrowserStorage like https://github.com/adaptlearning/adapt_authoring/pull/1378 ?

lc-thomasberger:
@taylortom this was sort of the idea with the model approach. Not sure if it should persist sessions, as this will likely spam your local storage and cleaning up will be hard. Hard to check if a article still exists if it is not deleted by the user.

lc-thomasberger:
Restore collapsed state and use plus / minus icons:
image

Have not updated the options view. This will need a proper update in a new PR.

canstudios-louisem:
Its a tiny thing but can we put these onto one line all icons in the AT are in line with their text. It should also probably say `Collapse Articles` `Expand Articles` so its a bit more clear what they do.

image

lc-thomasberger:
@canstudios-louisem alignment was fixed in this PR: https://github.com/adaptlearning/adapt_authoring/pull/1925

I updated the whole options layout to properly align the items using flexbox.

canstudios-louisem:
@lc-thomasberger why is that a sperate PR? Seems rather unnecessary to review it then review it again.

lc-thomasberger:
I was not sure how much needed changing so I started with a separate branch. I think it can be merged in this one directly so both can be reviewed at once.

taylortom:
Just found the screenshots I did for this:

Using the existing icons: 
screen shot 2018-03-23 at 11 05 12

Using the chevrons (my preference):
screen shot 2018-03-23 at 11 15 43

taylortom:
Looks good to me, thanks @lc-thomasberger! 

This makes dragging articles around a lot more intuitive. Once this PR has gone in, I wonder if we should look into minimising all articles on drag by default...

One minor thing -- I think the icons at the top need switching (i.e. the up chevron for 'collapse all', down for 'expand all')

lc-thomasberger:
@taylortom I've replicated githubs logic. down when expanded and up when collapsed.

taylortom:
Yeah I think it works on the articles themselves as it acts as a sort of status indicator, but seems the wrong way round in the buttons at the top (happy to go with the general consensus if people think otherwise though).

lc-thomasberger:
I think icon state for expand / collapse should always be the same.

taylortom:
Yeah that's fair, I think we maybe just need to invert them all then - which do you think is most intuitive?

lc-thomasberger:
Tbh. I never get the logic which chevron shows which state. 😄 That's why I decided to go for the plus and minus.
I am absolutely fine to go with chevrons though. It's only important that the icons are consistent for each state. If not it is confusing for the user.

taylortom:
Yeah I agree, maybe using [fa-plus](https://fontawesome.com/icons/plus?style=solid) and [fa-minus](https://fontawesome.com/icons/minus?style=solid) is the way to go 😅

lc-thomasberger:
I shall give it a go 😄

dancgray:
This works really well for me now. Noticed a couple of small things though, the trash icon on the menu editor is different 'purple' colour, and I can no longer double click on a component to edit it.

dancgray:
@lc-thomasberger, rebuilt but still cannot double click over the body of the component to edit which you used to be able to do:
screen shot 2018-04-11 at 13 52 28

I can double click when cursor is over component-detail-inner div:
screen shot 2018-04-11 at 13 51 46
Let me know if this is intended.

lc-thomasberger:
@dancgray thanks for posting the screenshots, I misunderstood your first query. Initially I updated the dbClick listener to only match the detail div. As @taylortom raised an issue with triggering the dbClick event when toggling article collapsed state. 
I updated the PR so now you can dbClick on the "body" of each element but excluded the buttons in the detail div. 
Also updated the styling of the menuItem's. It uses buttons instead of <a> and flexbox to align the content.

frontend/src/modules/editor/contentObject/views/editorPageArticleView.js
```diff
@@ -11,25 +11,37 @@ define(function(require){
...
 
     events: _.extend({}, EditorOriginView.prototype.events, {
-      'click a.add-block': 'addBlock',
-      'click a.article-delete': 'deleteArticlePrompt',
-      'click a.open-context-article': 'openContextMenu',
-      'dblclick': 'loadArticleEdit'
+      'click .add-block': 'addBlock',
+      'click .article-delete': 'deleteArticlePrompt',
+      'click .open-context-article': 'openContextMenu',
+      'dblclick .editor-page-element-detail-inner,.article-blocks': 'loadArticleEdit',
```
tomgreenfield: Please restore the original generic `dblclick` listeners so we may click anywhere to edit e.g. on an article's body text.

The code you've implemented in `editorOriginView.js` should be enough to prevent double clicking on buttons?

frontend/src/modules/editor/contentObject/less/editorPageArticle.less
```diff
@@ -20,13 +24,25 @@
...
+        padding: 20px 2%;
+        transition: opacity 200ms ease-in-out;
+    }
+    
+    &.collapsed-view {
+        overflow: hidden;
+        
+        .article-blocks {
+            transition: opacity 200ms ease-in-out;
+            opacity: 0.5;
```
tomgreenfield: Is the `opacity` property needed?

frontend/src/modules/editor/global/less/editor.less
```diff
@@ -100,35 +100,80 @@
...
+        box-sizing: content-box;
+        text-align: center;
+      }
 
-    .editor-delete-page-element {
-      float: right;
+      width: 100%;
+      box-sizing: border-box;
+      display: -ms-flexbox;
+      display: -webkit-flex;
```
tomgreenfield: Prefixed attributes not required in modern browsers.

frontend/src/modules/editor/contentObject/templates/editorPageArticle.hbs
```diff
@@ -1,16 +1,15 @@
...
 <div class="article-inner">
   <div class="article-detail unselectable">
-    <div class="article-detail-inner">
-      <h4 class="article-title title">{{#if displayTitle}}{{{displayTitle}}}{{else}}{{title}}{{/if}}</h4>
-      <div class="article-body">{{{body}}}</div>
-    </div>
+    {{>part_editorItemSidebar}}
   </div>
 
+  {{#if body}}
```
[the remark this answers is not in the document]
lc-thomasberger: Tried to not touch the dom structure as this might break existing theme amends.

frontend/src/modules/editor/global/less/editor.less
lc-thomasberger: but we can leave them in ... as it will help support older versions and not break modern browsers

tomgreenfield:
`overflow-y: scroll` will work fine with infinite scrolling – it's nearly always same behaviour as `auto`, with the addition of a disabled scrollbar when content doesn't overflow.

Therefore, happy for it to go on `contentPane` or `editor-view`.

frontend/src/modules/editor/global/less/editor.less
taylortom: I'd say we should probably try to avoid adding anything specifically to support legacy browsers (particularly as the framework no longer supports pre-v11 IE).

tomgreenfield:
I do see what you mean about `overflow-y` on the main content pane; the fixed position of Asset's Management preview container causes some whitespace in the layout.

Best to stick to `editor-view` as you suggest.

lc-thomasberger:
@canstudios-louisem may be related to markup update, ist this something that bothers us?

canstudios-louisem:
@lc-thomasberger i think its something that should be fixed as it doesn't look right now.

taylortom:
I've just made a few minor tweaks to the padding to look more consistent in the open/closed state. Most notable being the increased padding at the bottom of the article title when minimised (and increased page title to match). I can push this to **adaptlearning** if we want it (see commit [here](https://github.com/cgkineo/adapt_authoring/commit/c28bdd542943ef47ce06a0dd42666a5515a44716)).

See below for comparison.

### Before
screen shot 2018-04-25 at 13 07 18

### After
screen shot 2018-04-25 at 13 03 12

tomgreenfield:
Yep, please push @taylortom.

taylortom:
Todo:
- Padding on page body
- Initial animation

lc-thomasberger:
are we going to fix the initial animation? @taylortom

lc-thomasberger:
Have a fix for the padding of page-header, articIe-boy and block-body I am happy to push if we like the design.
before: https://files.gitter.im/lc-thomasberger/8de6/image.png

after: https://files.gitter.im/lc-thomasberger/sRRp/image.png

taylortom:
@lc-thomasberger yeah that looks great to me!

PR #1810, opened 2017-12-06
Asset filter fix

lc-thomasberger:
reset the collection when filter are set

### DIFF
--- a/frontend/src/modules/assetManagement/views/assetManagementCollectionView.js
+++ b/frontend/src/modules/assetManagement/views/assetManagementCollectionView.js
@@ -23,8 +23,8 @@ define(function(require){
       }
       this.initEventListeners();
 
-      this._doLazyScroll = _.bind(this.doLazyScroll, this);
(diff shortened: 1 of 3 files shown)


frontend/src/modules/assetManagement/views/assetManagementCollectionView.js
```diff
@@ -23,12 +22,16 @@ define(function(require){
         if(assetType) this.filters = assetType.$in;
       }
       this.initEventListeners();
+
+      this._doLazyScroll = _.bind(this.doLazyScroll, this);
```
taylortom: Should this be throttled?

frontend/src/modules/assetManagement/views/assetManagementCollectionView.js
```diff
@@ -23,12 +22,16 @@ define(function(require){
...
+      this._onResize = _.bind(_.throttle(this.onResize, 400), this);
```
taylortom: Think this needs to be debounced instead

PR #1812, opened 2017-12-06
throttle callbacks

lc-thomasberger:
use `.on` to properly remove scroll listener in remove function

### DIFF
--- a/frontend/src/modules/projects/views/projectsView.js
+++ b/frontend/src/modules/projects/views/projectsView.js
@@ -20,8 +20,11 @@ define(function(require){
     },
 
     initEventListeners: function() {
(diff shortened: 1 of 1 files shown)


frontend/src/modules/projects/views/projectsView.js
```diff
@@ -20,8 +20,11 @@ define(function(require){
     },
 
     initEventListeners: function() {
+      this._doLazyScroll = _.bind(_.throttle(this.doLazyScroll, 250), this);
+      this._onResize = _.bind(_.throttle(this.onResize, 250), this);
```
taylortom: I think we need a debounce here rather than a throttle?

taylortom:
Thanks for this 👍

PR #1820, opened 2017-12-13
Implemented basic docker environment

warslett:
A very simple docker environment

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,0 +1,34 @@
+# Running Adapt Authoring with Docker
+
+## Requirements
+
+You will need:
+* Docker https://docs.docker.com/engine/installation/
+* Docker Compose https://docs.docker.com/compose/install/
+
+
+## Installation
+
+* Run `docker-compose run --rm node npm install`
+* Run `docker-compose run --rm node node install`
+* During installation, hit enter to select the default option on all questions *except* "Database host" which should be
+"moodle"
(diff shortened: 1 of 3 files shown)


taylortom:
Thanks for this @warslett.

We recently discussed how we support this kind of thing going forward, and have agreed to not add these kinds of files to the main repo due to the extra expectations this adds in terms of maintainability (vagrant support was _officially_ removed in the last release for this reason).

I'm not sure the best approach for this would be -- maybe a new adaptlearning repo?

@canstudios-louisem, @dancgray @tomgreenfield any ideas?

canstudios-louisem:
Not sure if you saw the conversion in gitter @taylortom  but it was me that asked @warslett to share his docker stuff. I'm unsure of the best way to manage it in one repo or in its own both have their advantages. 

At Can we have our docker stuff in another repo and then pull in the AT repo to that, and that helps to keep things separate, you only pull it in if you need it etc.

warslett:
@canstudios-louisem @taylortom if the intention is to use Docker as a development environment for contributors it should stay in this repo. If the intention is to use it as a quick easy way of getting the tool up and running for users then a separate repo might be the way to go. Maybe even just a docker image with the code already checked out (a bit like how the wordpress image works: https://hub.docker.com/_/wordpress/). That way there would be no need to faff about cloning lots of different repos and mapping between containers for people who just want to run a single command and have it up and running. Would be especially useful if it could be configured by environment variables.

Issue #1791, opened 2017-11-17
Supporting HTML-based resources in Assets


fabiobeoni:
# Feature Request

## Affected Area

Assets library, assets import and export to course

## Requested Feature

Currently if you try to import an HTML file into the Assets Library you can complete the upload, but the HTML file doesn't appear in the assets list. Moreover, the HTML file uploaded in the assets doesn't get exported to the course build.

## Use Case

A user creates an HTML5-based content with an external editor, let's say Adobe Animate. Then he wants to include this content into the course.

User has a single HTML file or a ZIP including many files referring to a single entry point in the root

By the AT:

1. user uploads the HTML file into the Assets Library (HTML file holds an animation done with Adobe Animate or similar)
2. with-in the course page, the user includes an iframe in a block
3. the user can set the iframe SRC by opening the Assets Library and selecting the HTML file resource
4. the iframe shows the HTML file content while previewing the course (and of course when exported ;))

## Benefits
1. user can include any kind of HTML5-based content created with external editors, and fill some gaps of Adapt AT. Is the easiest way to bring custom and powerful vector animations into an Adapt course. 
2. HTML5-based content included in such way can work offline too, while they cannot when you embed them from external web site.
3. HTML5-based content can be a small client-side app or widget provided by the client company (some tool they have and they need to teach students about it in the course).

## Additional Information
Using the iframe as it is you may achieve the same result by referencing and external HTML file resource from another domain (like an embedding), or even an internal HTML file resource manually added to the course build export. Unfortunately this procedure is time consuming and has a really poor user experience.

paul-mediakitchen:
Yes please this sounds great!!

fabiobeoni:
I forgot to say that I'm available to work on it.

fabiobeoni:
I completed this feature, it also supports native Adobe Animate CC packages (.oam files). Waiting to open a pull request.

paul-mediakitchen:
Great I will give this a test. Thank you for creating this

PR #1821, opened 2017-12-13
Support HTML-based resources in Assets

fabiobeoni:
﻿­Hello guys, as described in #1791 the **idea behind this plugin is to build a basic integration between Adapt AT and Adobe Animate CC**, as well as any other external animation editor that can export animations to static HTML/JS files. Such editors normally export their animations to a ZIP file containing multiple static resources. The Adobe Animate .OAM package is an example of that (I can include one if needed). 

With this new “Animation Asset Unpack” (AAU) plugin, the  Adapt AT user can upload the animation OAM file into the assets library, and reference it into the course (as you normally do with images). Then the AAU plugin takes care of processing the OAM package during preview/publishing. As result, the animation is available to the output course in HTML format.

The AAU plugin works by adding a behavior to the default Preview/Publish features of Adapt AT. When installed, the AAU plugin starts watching the temp build directory of the course, and gets notified when a new file with OAM extension is available into the directory. Then it proceeds unpacking the OAM file into the “course/assets” folder, and removing the input package no longer needed.

This is an example of the output you get in course assets:

- /course
  - /assets
     - /myAnimation
       - /index.html
       - /anyOtherStaticResource

I tried to develop this new feature without making changes to the "core" code. I mean, I developed a back-end output plugin (_adapt_authoring/plugins/output/animationAssetUnpack_) with the intent of making the plugin 100% self contained. But I must say that I could not achieve that, and I had to change 1 line of code in the default output plugin too (_adapt_authoring/plugins/output/adapt_). I hope that it is not to be a problem.

To see this feature in action you also need to install the new "Adapt Animation Frame" component, that I have just released to the Adapt plugin repository (or https://github.com/fabiobeoni/adapt-animation-frame).

Some module dependencies are listed in the package.json of the plugin and need to be installed.

Two sample animation files are [available here](https://drive.google.com/drive/folders/1NAtgzZ-z0e3zmxrI_xycrbTMFor_DpSF?usp=sharing) to test the plugin.

Thanks! F.


plugins/output/animationAssetUnpack/index.js
```diff
@@ -0,0 +1,277 @@
...
+ * are moved one level up in the
+ * directory tree.
+ * @param packageDir {string}
+ * @param filesFromPackage {object[]}
+ */
+function removePackageRoot(packageDir, filesFromPackage) {
+  var entries = fs.readdirSync(packageDir);
+  // there is a root in decompressed package...
+  if (entries.length === 1) {
+    var packageRoot = entries[0] + '/';
```
truonghongtrieu: Hi @fabiobeoni , the separator does not cover Windows-based AT, [path.sep](https://nodejs.org/api/path.html#path_path_sep) would be needed here.
Recently I'm trying out your PR with [adapt-animation-frame](https://github.com/fabiobeoni/adapt-animation-frame), it works smoothly together!

plugins/output/animationAssetUnpack/index.js
```diff
@@ -0,0 +1,277 @@
...
+      for (var ii in changes.modifiedFiles) {
+        packagesList.push(path.join(watchingDir, changes.modifiedFiles[ii]));
+      }
+      //
+      // Monitor.on('change') also returns "changes.removedFiles"
+      // but we don't care about them since here we
+      // are working with the course preview only,
+      // and the output doesn't go into the course
+      // export downloaded by the user.
+      pluginInstance.unpack(packagesList, function onUnpacked(err) {
```
truonghongtrieu: Hi @fabiobeoni, 
Once the .oam file had been removed, fsmonitor.on('change') will trigger unpack() again. I think it's worth to consider removing that file from watchers, or perhaps in unpack() we have a checkpoint - something before running loopPackages() like:
`if (!fsextra.existsSync(packageFileName)&& (unpackedCount === total)) {
 onUnpackDone(null, filesFromPackage);
}`
Thanks again,
Cheers

plugins/output/animationAssetUnpack/index.js
fabiobeoni: Thanks for your note, sometime I would remove the "/" button on my keyboard ;)  Let's change it.

plugins/output/animationAssetUnpack/index.js
fabiobeoni: Totally agree. Thanks

taylortom:

PR #1822, opened 2017-12-14
Allow MongoDB authentication in install script

Annoraaq:
Previously it was not possible to pass authentication parameters for MongoDB to the install script, making it impossible to install the authoring tool on a secured MongoDB instance.


taylortom:
Have resubmitted this to avoid the weird history issues

PR #1829, opened 2017-12-19
Fix install script's storing of config.json values

### LINKED ISSUE #1828: Install script doesn't take into account all config.json values
## Expected Behaviour
Script should use all values in `config.json`, removing the need to input these values again.

## Actual Behaviour
Script prompts the user for inputs from the `inputData.features` inputs onwards.

## Steps to Reproduce
1. Run through install script, choosing `y` when asked to use `config.json` values

## Versions

taylortom:
Fixes #1828 and #1842.

### DIFF
--- a/install.js
+++ b/install.js
@@ -337,6 +337,15 @@ function configureMasterTenant(callback) {
       if(error) {
         return callback(error);
       }
+      if(USE_CONFIG && prompt.override.masterTenantName) {
+        /**
+        * remove the masterTenantDisplayName, as we can use the existing value
+        * (which isn't in config.json so can't be used as an auto override)
+        */
+        inputData.tenant = _.filter(inputData.tenant, function(item) {
+          return item.name !== 'masterTenantDisplayName';
+        });
+      }
       installHelpers.getInput(inputData.tenant, function(result) {
         console.log('');
         // add the input to our cached config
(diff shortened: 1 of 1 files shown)


dancgray:
Is useSmtp meant to be populated, seems to be missing in my tests.

install.js
```diff
@@ -231,7 +231,7 @@ installHelpers.getLatestFrameworkVersion(function(error, latestFrameworkTag) {
 function generatePromptOverrides() {
   if(USE_CONFIG) {
     var configJson = require('./conf/config.json');
-    var configData = JSON.parse(JSON.stringify(configJson).replace('true', '"y"').replace('false', '"n"'));
+    var configData = JSON.parse(JSON.stringify(configJson).replace(/true/g, '"y"').replace(/false/g, '"n"'));
```
canstudios-louisem: This feels very hacky.

taylortom: This was to deal with how we use y/n for boolean input values (which is the default for prompt). We could maybe look into also accepting true/false as values.

taylortom: I'm happy to look into it now, but as it's an enhancement, I'd prefer to push until 0.5.0 (or even the next patch if we have one)

taylortom:
What's the consensus with this one?

PR #1841, opened 2018-01-15
Fixes double-rendering of blocks on new article creation

lc-thomasberger:
not pretty but it solves it for me ...

do not render the blocks when a new article is created. This makes sure the block View is only rendered once the block Model is created.

### DIFF
--- a/frontend/src/modules/editor/contentObject/views/editorPageView.js
+++ b/frontend/src/modules/editor/contentObject/views/editorPageView.js
@@ -112,6 +112,7 @@ define(function(require){
       }, {
         success: _.bind(function(model, response, options) {
           var articleView = this.addArticleView(model);
+          articleView._skipRender = true; // prevent render of blocks in postRender
           articleView.addBlock();
         }, this),
         error: function() {
--- a/frontend/src/modules/editor/contentObject/views/editorPageArticleView.js
+++ b/frontend/src/modules/editor/contentObject/views/editorPageArticleView.js
@@ -22,7 +22,9 @@ define(function(require){
     },
 
     postRender: function() {
-      this.addBlockViews();
+      if (!this._skipRender) {
+        this.addBlockViews();
+      }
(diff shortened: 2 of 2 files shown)


frontend/src/modules/editor/contentObject/views/editorPageView.js
```diff
@@ -112,6 +112,7 @@ define(function(require){
       }, {
         success: _.bind(function(model, response, options) {
           var articleView = this.addArticleView(model);
+          articleView._skipRender = true; // prevent render of blocks in postRender
           articleView.addBlock();
         }, this),
```
taylortom: Does it make sense to remove that `addBlock` call completely, and just do this? 
(this seems to fix it for me, but I may have missed something)
```js
success: _.bind(function(model, response, options) {
  this.addArticleView(model);
}, this)
```

lc-thomasberger:
@taylortom can't really follow your suggestion. The function was already in the code before I added the fix. My goal really was to fix without messing to much with the code 😄

taylortom:
I understand your light-touch fix, but also wanted to ask the question as I'm not convinced I put the code in the right place -- my main concern is that sticking plaster fixes tend to stay in the codebase for years, so would prefer to fix things as we go along.

My suggestion above was to move the model creation code to editorPageView -- this way, editorPageArticleView just renders all its child models, without needing to worry about when it should/shouldn't be rendering.

lc-thomasberger:
I will have a look at this and see if restructuring the code will solve the issue.

taylortom:
If it looks like it could be tricky, happy to just log an issue and push this to the next release :)

lc-thomasberger:
I think we should refactor the Code later. Since it is a rather complex UI-Element, a thorough refactor would be good. Also, I do think that the PageEditor is missing some UI / UX features that might be integrated in the refactoring stage.

PR #1844, opened 2018-01-17
Fixes sub-menu rendering quirks

### LINKED ISSUE #1837: Deleting submenu leaves orphaned view and courseassets
## Expected Behaviour
In the menu editor deleting a submenu that contains pages should delete all articles, block, components and courseasset records for those pages. The menu editor view should also render correctly.

## Actual Behaviour
Records for contentObject, article, block and component are removed. However courseasset records for components containing media assets are not removed, leaving orphaned records.

Also the menu editor display is not correct:
screen shot 2018-01-15 at 09 39 02

## Steps to Reproduce
1. Create a course with a submenu
2. Add a page to that submenu that contains a media/graphic component with corresponding asset selected.
3. Delete the submenu using either the trash button or cog/delete.

## Versions
  - Authoring Tool Version: tag/v0.4.1-rc.1
  - Framework Version: v2.2.3
  - Operating System: Mac OS 10.13.2
  - Browser: Chrome 63

lc-thomasberger:
Partially fixes #1837

### DIFF
--- a/frontend/src/modules/editor/contentObject/views/editorMenuView.js
+++ b/frontend/src/modules/editor/contentObject/views/editorMenuView.js
@@ -145,11 +145,8 @@ define(function(require){
     },
 
     onSelectedItemChanged: function(model) {
-      if (model) {
-        if (model.get('_id') === Origin.editor.currentContentObject && Origin.editor.currentContentObject.get('_id')) {
-          return;
-        }
-      }
+      if (!model || !Origin.editor.currentContentObject) return;
+      if (model.get('_id') === Origin.editor.currentContentObject.get('_id')) return;
 
       Origin.editor.currentContentObject = model;
       this.renderLayers();
--- a/frontend/src/modules/editor/contentObject/views/editorMenuView.js
+++ b/frontend/src/modules/editor/contentObject/views/editorMenuView.js
@@ -55,13 +55,11 @@ define(function(require){
 
         // remove all unused layerviews 
         for (var id in this.layerViews) {
-          if (this.layerViews.hasOwnProperty(id)) {
-            var layerView = this.layerViews[id];
-            if (ids.indexOf(id) === -1) {
-              layerView.remove();
-              delete this.layerViews[id];
-            }
+          if (this.layerViews.hasOwnProperty(id) || ids.indexOf(id) === -1) {
+            continue;
           }
+          this.layerViews[id].remove();
+          delete this.layerViews[id];
         }
 
         _.defer(_.bind(function() {
@@ -75,14 +73,13 @@ define(function(require){
     /**
      * Renders a single menu layer
      */
-    renderLayer: function(model, callback) {
+    renderLayer: function(model) {
       var menuLayerView = new EditorMenuLayerView({
         _parentId: model.get('_id'),
         models: this.contentobjects.where({ _parentId: model.get('_id') })
       });
       this.layerViews[model.get('_id')] = menuLayerView;
       $('.editor-menu-inner').append(menuLayerView.$el);
-      if (typeof callback === 'function') callback();
     },
     
     updateItemViews: function(previousParent, model) {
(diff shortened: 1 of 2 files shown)


frontend/src/modules/editor/contentObject/views/editorMenuView.js
```diff
@@ -82,8 +80,25 @@ define(function(require){
         _parentId: model.get('_id'),
         models: this.contentobjects.where({ _parentId: model.get('_id') })
       });
+      this.layerViews[model.get('_id')] = menuLayerView;
       $('.editor-menu-inner').append(menuLayerView.$el);
-      if(typeof callback === 'function') callback();
```
taylortom: I'm not sure we still need this async stuff? I may have left it in accidentally. Will also need to remove func param

frontend/src/modules/editor/contentObject/views/editorMenuView.js
```diff
@@ -39,38 +40,35 @@ define(function(require){
...
         }
-        // we can reuse layers up to 'index', remove the rest
-        if(index !== undefined) {
-          hierarchy = hierarchy.slice(index);
-          var layersToRemove = renderedLayers.slice(index);
-          for(var i = 0, count = layersToRemove.length; i < count; i++) {
-            layersToRemove[i].remove();
+
+        // remove all unused layerviews 
+        for (var id in this.layerViews) {
```
taylortom: Can this be refactored to be a bit more readable? Something like:
```
// remove all unused layerviews 
for (var id in this.layerViews) {
  if (!this.layerViews.hasOwnProperty(id) || ids.indexOf(id) > -1) {
    continue;
  }
  this.layerViews[id].remove();
  delete this.layerViews[id];
}
```
*think* that does the same...

frontend/src/modules/editor/contentObject/views/editorMenuView.js
```diff
@@ -114,28 +129,31 @@ define(function(require){
...
       _getParent(model, function() {
-        if(typeof done === 'function') done.call(__this, hierarchy.reverse());
+        if (typeof done === 'function') done.call(__this, hierarchy.reverse());
       });
     },
 
     onSelectedItemChanged: function(model) {
-      if(model.get('_id') === Origin.editor.currentContentObject && Origin.editor.currentContentObject.get('_id')) {
-        return;
+      if (model) {
```
taylortom: This can probably just be `if(model && model.get('_id') ...`, or
```js
if (!model || !Origin.editor.currentContentObject) {
  return;
}
if(model.get('_id') === Origin.editor.currentContentObject.get('_id')) {
   return;
}
```
Prefer the first for brevity, although is a bit long

frontend/src/modules/editor/contentObject/views/editorMenuView.js
lc-thomasberger: thanks, fixed it

frontend/src/modules/editor/contentObject/views/editorMenuView.js
lc-thomasberger: thanks, fixed it

frontend/src/modules/editor/contentObject/views/editorMenuView.js
```diff
@@ -114,28 +126,28 @@ define(function(require){
...
-        if(typeof done === 'function') done.call(__this, hierarchy.reverse());
+        if (typeof done === 'function') done.call(__this, hierarchy.reverse());
       });
     },
 
     onSelectedItemChanged: function(model) {
-      if(model.get('_id') === Origin.editor.currentContentObject && Origin.editor.currentContentObject.get('_id')) {
-        return;
-      }
+      if (!model || !Origin.editor.currentContentObject) return;
```
taylortom: Apologies, this is causing the code to block items being selected for the first time something's clicked (as Origin.editor.currentContentObject will always be undefined)

dancgray:
Just tested and cannot reproduce issues with nested menus. Tested in chrome 63.

frontend/src/modules/editor/contentObject/views/editorMenuView.js
taylortom: Yeah, think the loop was fine, just the if condition

lc-thomasberger: updated

PR #1852, opened 2018-01-24
Allow upgrade if missing frameworkRepository or authoringToolRepository

### LINKED ISSUE #1851: Upgrade script fails if missing `authoringToolRepository` and `frameworkRepository` from config.json.
## Expected Behaviour
Should be able to perform an automatic upgrade if `authoringToolRepository` and `frameworkRepository` attributes are missing from config.json.

## Actual Behaviour
Upgrade process falls over with the message:
```
Cannot perform an automatic upgrade when custom repositories are used.
```

## Steps to Reproduce
1. Remove `authoringToolRepository` and `frameworkRepository` from config.json
1. Run `node upgrade`
1. Input `y` and `y` to the two questions

## Versions

taylortom:
Fixes #1851.

### DIFF
--- a/upgrade.js
+++ b/upgrade.js
@@ -25,7 +25,19 @@ function start() {
   logger.level('console','error');
   // start the server first
   app.run({ skipVersionCheck: true, skipStartLog: true });
-  app.on('serverStarted', getUserInput);
+  app.on('serverStarted', function() {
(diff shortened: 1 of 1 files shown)


canstudios-nicolaw:
I think this branch now has the wrong destination?

PR #1854, opened 2018-01-24
Fix route error logging

### LINKED ISSUE #1853: Router logs 'no index.js found' for any MODULE_NOT_FOUND when loading routes
When Router attempts to load a route (by require-ing `index.js`), if a `MODULE_NOT_FOUND` error is thrown, a generic 'no index.js found' message is logged. Depending on where the error's thrown, this could be caused by something else (e.g. missing local file, missing node module).

## Steps to Reproduce
1. Require a node module that doesn't exist in a `routes/index.js` file.

## Versions

taylortom:
Fixes #1853


canstudios-louisem:
This was also targeting the wrong branch and will need to be updated

PR #1856, opened 2018-01-24
Handle Database#addModel errors appropriately

### LINKED ISSUE #1855: Database doesn't handle errors when adding new models
The Database#addSchema function is async, and can throw an error if there's something wrong with a schema. This should be handled appropriately in Database#addModel

taylortom:
Fixes #1855


canstudios-louisem:
This isn't going to the correct branch

canstudios-louisem:
I've changed the target but you will need to rebase the changes

PR #1864, opened 2018-02-04
Merge master

taylortom:


taylortom:
Left this PR open in case someone can point out what I did wrong....

PR #1868, opened 2018-02-08
Scroll position

### LINKED ISSUE #1814: scroll position in page editor
## Expected Behaviour
scroll position in page editor should be persistent 

## Actual Behaviour
When you edit an element in the page editor and return back, scrollTop is set to 0.

## Steps to Reproduce
1. create course
2. add enough components so that scrollbars appear
3. edit a component
4. return to page editor

## Versions
  - Operating System: All
  - Browser: All

lc-thomasberger:
fixes #1814


canstudios-nicolaw:
Just a minor thing I've noticed, if I have a course with lots of components and a scrollbar, and I scroll to the very bottom of the page and edit the last component, when I press save the page is opened at the bottom and then it scrolls up a little.

The last component is still in view, but it is only just in view.

Is this known behaviour? I don't think it's necessarily broken or needs fixing, it just struck me as a little odd at first.

lc-thomasberger:
Thanks @canstudios-nicolaw . I have not noticed this. 
This may due to the entry animation of the elements. So the scroll position is set to the very bottom and then the content of the page changes and makes the overall page longer.
I think there isn't an easy way to hook into once all elements are rendered on screen with the right content and height to set the scroll position again or only then.

lc-thomasberger:
I think the animation was required? @dancgray 

If it is a transition we can also use the `transitionend` [event](https://developer.mozilla.org/de/docs/Web/Events/transitionend) to hook into the animation and set the scroll position when all elements are rendered.

taylortom:
https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/modules/editor/global/less/editor.less#L152-L166

PR #1870, opened 2018-02-09
Add ability to Set a Default View and/or Default to "Recent"

### LINKED ISSUE #1857: Ability to Set a Default View and/or Default to "Recent"
# Course Ordering -- Preference for "Recent" courses
I notice that the authoring environment's home screen seems to default to ordering courses by the time they were created[?] or maybe alphabetically[?] 

I usually want to see the courses I'm currently working on first, so defaulting to "Recent" first would be my preferred default. Of course, other people may disagree, so it would be awesome to have a toggle or preference that allows us to set our own preferred view. 

## Steps to Reproduce
1. Go to authoring environment
1. See how courses are ordered
1. Scratch head and wonder why...

lc-thomasberger:
fixes #1857 

`setUserPreference()` from origonView can store values in localStorage. These settings are synced with the sessionModel attributes.

In the ProjectsView the settings for course display and sorting are stored for a user in local storage.

### DIFF
--- a/frontend/src/core/app.js
+++ b/frontend/src/core/app.js
@@ -59,6 +59,7 @@
       'modules/filters/index',
       'modules/globalMenu/index',
       'modules/help/index',
+      'modules/localStorage/index',
       'modules/location/index',
       'modules/modal/index',
       'modules/navigation/index',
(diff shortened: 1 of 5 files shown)


taylortom:
@lc-thomasberger I'd assigned this one to me as there's already an old PR open for this

lc-thomasberger:
Did not see this one. Should I close this one then?

taylortom:
I think let's leave them both for now, and discuss what we want from this in the next catch-up :)

PR #1881, opened 2018-02-20
Add colour labels to content

### LINKED ISSUE #1785: color labels
## Affected Area
Eidtor Interface

## Requested Feature
Use colors to label individual elements (Articles, Blocks, Components). Similar to Finder in MacOSX.
Color might be selected by the user, or from a list of predefined colors.

## Use Case
Color labels would be useful to highlight sections when editing long pages.

## Current Workaround
None

lc-thomasberger:
resolves #1785 

Adds ability to set background colors for elements in page and menuEditor. Right now the colors are stored in a model. Therefore, user defined colors are not yet supported. 

I took the colors from google Material-Design Docs. Feel free to suggest colors :)

image

If you set a color-label for a Page, it also applies this color to the background-color in the pageEditor. That's because bot view's extend from the same base view. To me this is not an issue, as it actually improves UX by showing "I am currently working on the redish page I just entered.". I do however understand if the color is a bit to much and I am happy to change this.

image
image

### DIFF
--- a/frontend/src/core/app.js
+++ b/frontend/src/core/app.js
@@ -70,6 +70,7 @@
       'modules/sidebar/index',
       'modules/user/index',
       'modules/userManagement/index',
+      'modules/colorLabel/index',
     ], callback);
   }
 
--- a/frontend/src/core/models/blockModel.js
+++ b/frontend/src/core/models/blockModel.js
@@ -27,8 +27,10 @@ define(function(require) {
       'title',
       '_extensions',
       'themeSettings',
-      '_onScreen'
+      '_onScreen',
+      '_colorLabel'
     ]
+
   });
 
   return BlockModel;
--- a/frontend/src/core/models/componentModel.js
+++ b/frontend/src/core/models/componentModel.js
@@ -26,7 +26,8 @@ define(function(require) {
       'title',
       'version',
       'themeSettings',
-      '_onScreen'
+      '_onScreen',
+      '_colorLabel'
     ]
   });
 
--- a/frontend/src/modules/colorLabel/index.js
+++ b/frontend/src/modules/colorLabel/index.js
@@ -1,0 +1,48 @@
+define([
+    'core/origin',
+    './models/colors',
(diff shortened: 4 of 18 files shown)


lc-thomasberger:
I tried to add the color-labels also to courses in the dashboard. I think it would be nice to group courses by color, however, the `colorLabel` object was not stored in the DB. If anyone knows what prevents this a hint is very much welcome 😃

canstudios-louisem:
Rather than colouring the whole page background why not colour the page title div? It would make it a lot less offensive on the eyes :)

lc-thomasberger:
@canstudios-louisem sounds sensible. Will have a look at this! 
My goal is to touch as less files as possible. Maybe just give the page a colored border if your suggestion doesn't work.

taylortom:
I'm slightly concerned that this will confuse technophobes, as it visually gives the impression that you've changed the rendered colour of an element.

I very much like the idea, but think we could do with looking at this a bit more closely to figure out exactly what usability problem we're trying to fix.

lc-thomasberger:
@taylortom I get your point. I think we can add a Text in the color selection popup that informs the user that the colors are only applied in the authoring tool and will not affect the course output.

We might also change the styling, for instance only use a colored border.

I have updated the way the colors are applied. This required a overwrite in every editView of the  `applyColorLabels` method. If you don't like this, I can create a new View from `EditorOriginView` that handles the styling. This View would then be the base for the Views that need colorLabels.

canstudios-louisem:
The more I think about this the more I'm not sure how useful this would be over tag's or references and I do think it could be confusing to a user that it only applies to the AT.

I personally have always found if you have some kind of colouring system you need a key to go with it and at that point, you may as well just put the key's labels on them rather than the color's.

lc-thomasberger:
have screwed up the branch, will creat a new PR

lc-thomasberger:
new PR https://github.com/adaptlearning/adapt_authoring/pull/1921

PR #1886, opened 2018-02-23
minor framework-import-fixes

lc-thomasberger:
fix issue with path separator on Windows
update button loading animation after form validation


canstudios-nicolaw:
There is a separate issue for making the reference to fs-extra more consistent across the whole tool https://github.com/adaptlearning/adapt_authoring/issues/1901

Issue #1875, opened 2018-02-14
Add ability to enable framework plugins by default on new courses


kineopete:
## Affected Area
Plugin Management section
## Requested Feature
It would be handy to be able to specify in the plugin management section those plugins which you would like to enable in all new courses by default. UI-wise I'm thinking an extra column of checkboxes headed up with 'Default' next to the  'Enabled' column.
## Use Case
This particularly relates to the new AATJSON extension available from here - https://github.com/cgkineo/adapt-AATJSON. It's a prefilter which cleans up AAT json so it works better with screenreaders. It would be very handy for any organisation that needs their content to be as accessible as possible to be able to ensure that this extension is added to all new courses and enabled by default.
## Current Workaround
Manually enabling your favourite extensions when setting up a new course.
## Additional Information

PR #1891, opened 2018-03-02
Experimental: automatically add extensions to new courses

taylortom:
#1875


plugins/content/extension/index.js
```diff
@@ -461,52 +485,35 @@ function initialize () {
...
-      }
-
-      toggleExtensions(courseId, 'disable', extensions, function(error, result) {
-        if (error) {
+      disableExtensions(req.params.courseid, req.body.extensions, function(error) {
+        if(error) {
+          logger.log('error', error);
           res.statusCode = error instanceof ContentTypeError ? 400 : 500;
-          return res.json({ success: false, message: error.message });
+          return res.status(400).json({ success: false, message: error });
```
lc-thomasberger: isn't statusCode already set above?

lc-thomasberger:
Have tested and works for me. Happy to +1 once the conflict is resolved.

canstudios-louisem:
Can we clarify what this is solving? The linked issue does not match the PR title.

taylortom:
That is the correct issue, but the issue's title is a bit misleading. 

The aim is to be able to flag certain extensions so that they're automatically enabled on new courses.

lc-thomasberger:
I think the feature makes sense. For instance most projects will require Tutor for displaying Feedback.

I also think the opposite would be handy, "disable extensions for publish". For instance, you would use dev-tools in the preview, but it will automatically excluded once you publish a course when the checkbox is enabled.

PR #1894, opened 2018-03-02
Grunt task to flag HACK/TODO/FIXME comments

### LINKED ISSUE #1867: HACK | TODO | FIXME
The whole authoring tool is riddled with these as comments most of them don't have related issues, so they are not being tracked. 

Some of them are minor, should be removed and have an issue raised for them. Others are somewhat more serious and need to be looked at.

It does concern me that some have "sneaked in" quite recently and no one thought to raise them when writing them, creating the pr that pushed them or reviewing the PR. I think people need to stop and think if they write a `//HACK` or `//FIXME` in the code base "should I be fixing this" or "has the code I've written that creates this issue put the code base in a better state than when I left it.

I suggest we start by listing these out and looking first at the `//HACK` then `//FIXME` and '//TODO` comments. Along with checking all PR's more thoroughly for these especially the big pr's I'm sure that's how most have got in and I know I've reviewed those so I must have missed them as well.

taylortom:
Fixes #1867

Run as part of grunt test, so will be a part of the Travis CI process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@
 /nbproject/
 /master/
 
+package-lock.json
+
 /plugins/content/component/versions/
 /plugins/content/component/componentcache/
 /plugins/content/extension/versions/
@@ -31,3 +33,4 @@
 
 .idea
 .project
+.vscode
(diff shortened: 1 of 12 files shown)


canstudios-louisem:
Its a great idea but we cant merge it untill all the TODO's HACK's are fixed otherwise all builds will fail.

taylortom:
👍 I was thinking we could also do with only showing errors for new changes too

canstudios-louisem:
I have been thinking the same but can't really find a good way to do it. I think it would need a GitHub CI integration.

PR #1896, opened 2018-03-05
Fix pluginDependencies

### LINKED ISSUE #1858: pluginDependencies in plugin bower files should work
## Affected Area
It should be possible to add a `pluginDependencies` object to the bower.json of any plugin and have the plugins listed in it pulled into the build. Eg. The hotgraphic component should be able to ensure the narrative component gets included by listing it in its pluginDependencies.

## Requested Feature
At the moment the pluginDependencies are ignored by the authoring tool. Instead in the outputmanager.js file there is a hardcoded hack which adds the narrative component if the hotgraphic is present, and adds the mcq if the gmcq component is present.

## Use Case
New plugins with dependencies can be developed.

## Additional Information
The hack was added in these PRs:
https://github.com/adaptlearning/adapt_authoring/issues/1033
https://github.com/adaptlearning/adapt_authoring/issues/1038

canstudios-nicolaw:
Fixes #1858


canstudios-nicolaw:
This PR removes the hacks that were added in the linked issue. There were two hacks, one to make sure narrative was included if hot graphic was present, and one to make sure mcq was included if gmcq was present.

With this change the narrative component is still included if hotgraphic is present, because hotgraphic mentions narrative in it's pluginDependencies.

However mcq is no longer included when gmcq is present. I have tested a gmcq component locally and can't see a problem (on mobile or full screen). If anybody knows why the mcq needs to be included when gmcq is present then I will make a PR to add it to gmcq's pluginDependencies.

taylortom:
Haven't tested this, so don't know what effect it has, but the code's required here: https://github.com/adaptlearning/adapt-contrib-gmcq/blob/master/js/adapt-contrib-gmcq.js#L3

canstudios-nicolaw:
Hi @tomgreenfield, MCQ doesn't get included when exporting the source because the GMCQ component doesn't list the MCQ component in it's plugin dependencies.

As I mentioned in the comment above, I have tested a course with a GMCQ component locally and I couldn't see a problem with this. Are you aware of a specific problem that might be caused by MCQ not being included in the export? If so I will need to make a PR to the GMCQ component to update it's pluginDependencies.

tomgreenfield:
Hi @canstudios-nicolaw. GMCQ _definitely_ extends MCQ as @taylortom [mentions](https://github.com/adaptlearning/adapt_authoring/pull/1896#issuecomment-370378888). It will not function without it.

I haven't investigated how MCQ's code is being bundled in for the download and preview tasks. This will need clarifying before I can +1 the PR.

canstudios-nicolaw:
Thanks @tomgreenfield I'm happy to add it to the gmcq pluginDependencies if needed, because the old hack was in place in the tool for a while it's hard to tell if it is still needed or if it is no longer needed and just never got removed as the GMCQ component was updated.

I really appreciate your help testing this as I've tested all I could think of.

canstudios-nicolaw:
Was anything broken in your exported course?

tomgreenfield:
The **export** will simply not compile because there is no MCQ code:
```console
Running "javascript:compile" (javascript) task
Fatal error: Error: ENOENT: no such file or directory, open 'C:\Users\tom.greenfield\Desktop\mcq-test-export-2018-03-27-090947\src\components\adapt-contrib-mcq\js\adapt-contrib-mcq.js'
In module tree:
    core/js/app
      components/adapt-contrib-gmcq/js/adapt-contrib-gmcq
```
Whereas the **download** has the compiled MCQ code, even though it isn't in the build includes:
```json
"build": {
  "includes": [
    "adapt-contrib-vanilla",
    "adapt-contrib-boxMenu",
    "adapt-contrib-text",
    "adapt-contrib-gmcq"
  ]
}
```
download

tomgreenfield:
For the preview and download tasks, MCQ code is compiled by RequireJS in the [`javascript`](https://github.com/adaptlearning/adapt_framework/blob/master/grunt/tasks/javascript.js#L50) Grunt task. MCQ's Less isn't included but this isn't needed by GMCQ.

As discussed in the weekly call, we probably should retain the existing hardcoded plugin check as a stopgap between releases.

taylortom:
[outputmanager.js#L384-L393](https://github.com/adaptlearning/adapt_authoring/blob/issue/1858/lib/outputmanager.js#L384-L393)

@tomgreenfield is this what we were planning to do?

canstudios-nicolaw:
Sorry I've not been around much the past few days. @taylortom we were planning to just re-add the hack for the mcq/gmcq relationship, the hotgraphic/narrative relationship works fine.

So lines 391-393 from your link.

taylortom:
Have just updated my branch (see [outputmanager.js#L384-L390](https://github.com/adaptlearning/adapt_authoring/blob/issue/1858/lib/outputmanager.js#L384-L390)). Feel free to merge this into yours if it's what we need :)

canstudios-nicolaw:
@tomgreenfield @taylortom I have (finally) updated this PR now so you should find the export contains the mcq dependency. @taylortom I didn't merge your branch because I re-added the hack in a slightly different place so that it was only called once.

PR #1900, opened 2018-03-06
Fix modules define

### LINKED ISSUE #1866: Adding front end plugins requires editing the app.js
## Expected Behaviour
Adding a plugin to the `frontend/plugins` folder should just work as `plugins.js` file can be built and used to require the plugin into the tool

## Actual Behaviour
The `plugins.js` is ignored and you need to edit the `core/app.js` [here](https://github.com/adaptlearning/adapt_authoring/blob/dc04351192d62c6e6f6520e33eee1478a01d672d/frontend/src/core/app.js#L48) 

## Versions
  - Authoring Tool Version: 0.4.1

taylortom:
Fixes #1866

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -260,23 +260,21 @@ module.exports = function(grunt) {
 
   // Compiles frontend plugins
   grunt.registerMultiTask('requireBundle', 'Generates a .js file with a bunch of imports for the path files', function() {
-    var requirePaths = '';
+    var modulePaths = '';
     // Go through each subfolder in the plugins directory
     var foldersArray = grunt.file.expand({ filter: "isDirectory" }, this.data.src);
     // Check if any plugins are available
     if (foldersArray.length === 0) {
-      requirePaths += "'";
+      modulePaths += "'";
     }
     foldersArray.forEach(function(path, index, folders) {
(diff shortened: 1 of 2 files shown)


canstudios-nicolaw:
At Can we have some plugins that are dependent on things in the modules being loaded first. (There is a plugin which changes the elements on the dashboard display for example.) So for us I have had to tweak this slightly to ensure modules are loaded before plugins (included below).

I'm not sure how much of a problem this would be for others. It could be that we need to re-think how some of our plugins are written, I'm just pointing this out in case it hadn't been considered. Would there be any interest in me making a PR with this tweak?

```
  function loadModules(callback) {
    require([
      'modules/modules'
    ], callback);
  }

  function loadPlugins(callback) {
    require([
        'plugins/plugins'
    ], callback);
  }

  /**
  * Start app load
  */
  loadLibraries(function() {
    loadCore(function() {
      loadModules(function() {
        loadPlugins(function() {
          // start session
          // FIXME required here to avoid errors
          require(['modules/user/models/sessionModel'], function(SessionModel) {
            origin.startSession(new SessionModel());
          });
        });
      });
    });
  });
})();
```

taylortom:
Happy for this to go ahead 👍

PR #1916, opened 2018-03-14
Export import assets

### LINKED ISSUE #1915: Import/export assets round trip creates duplicates
Proposal: https://docs.google.com/document/d/1c1OL6_tUnn-jS8lDWu1TR8XGez1t08OzV_WUfzOdDDE/edit#

If a course contains assets and it is exported and imported into the same tool, the assets are likely to be duplicated. When an asset is exported it has the db attribute 'filename' as it's name. During the import the import process checks if the filename matches the title of an existing asset. As it doesn't the asset is incorrectly identified as being new.

The following changes are outlined in the proposal:
- When exporting a course from the tool, the export contains an additional file, assets.json, which contains the filename, title, description and tags for every asset in the course.
- When importing a course, compare the filename of assets to be imported with the filename of assets in the database (rather than comparing filename to title)
- When importing a course, if there is not a filename match, and an assets.json file exists, create a new asset with the same title, description, tags and filename as provided in assets.json
- assets.json is not mandatory to import and is ignored by the framework

canstudios-nicolaw:
Fixes #1915


canstudios-nicolaw:
I have tested the following locally (I hope this will help others to test):

For an exported course that contains assets.json:
- Export and re-import, check assets are not duplicated in the asset manager
- Export and delete original course and assets from db, import and make sure new assets are created with correct title, description and tags
- Export course that contains image with tags, delete the course and the image, on the framework import page enter some additional tags and import, check the image has been re-added to the server with the same title and description as before, and both the tags the image had already and the tags that were added in the import form
- Export a course with 2 of the same image, delete original course and asset, import, make sure only one asset is imported and it has correct title, description and tags

For an exported course that does not contain assets.json:
- Export and re-import, check assets are not duplicated
- Export, delete original course and assets, import, check new assets just have filename

dancgray:
Manually merged [here](https://github.com/adaptlearning/adapt_authoring/commit/a006235a56b7a41bb2b85462095223a50442fd09) and conflicts resolved. Done to allow testing.

canstudios-louisem:
Closing as it was merged in by dan manually

PR #1921, opened 2018-03-19
Add colour Labels to elements 

### LINKED ISSUE #1785: color labels

lc-thomasberger:
fixes #1785


canstudios-louisem:
If I set a colour label on an article or block and then refresh the label stays if I do the same on a component it does not.

lc-thomasberger:
@canstudios-louisem `_colorLabel` was not defined in the schema for components ... thought it would also inherit from content. Should work now.

taylortom:
Works for well me :) 

A few minor things:
- Could you switch the context label to use the british english variation 'colour' (not sure if we have this documented in a language style guide anywhere, but that's what we use)
- I think it would also be useful to add some kind of description text/title in the colour select popup, as it looks a bit blank/not obvious what's going on (see screen below)
- One last thing: should have some way to configure these colours other than in the LESS? Seems like something people might want to change

screen shot 2018-03-23 at 15 31 56

lc-thomasberger:
Cool!
The re-design of the labels required switching to css rather than inline styling. So the colors must be set in the theme of the AT. The easiest way would be a file that stores all the variables that can than be customized. I will updated the colors to use generic names instead of the colors.

canstudios-louisem:
If you're going to put them in a file can they just go in the config? I would rather we didn't create more config files.

lc-thomasberger:
@canstudios-louisem it would be a less file. If it should be set via configuration we would need to inject the file in the grunt task.

dancgray:
Colour labels work really well. Is anyone else getting an error "constructorFn is not a constructor" when trying to edit a component? Could be my install.

lc-thomasberger:
@dancgray good catch ... I will have a look at this. It only shows a popup without an error so not sure what causes this

frontend/src/core/less/colourLabels.less
```diff
@@ -0,0 +1,14 @@
+@colorlabel-1: #616161;
```
tomgreenfield: Name of this file is inconsistent with the other Less files.

frontend/src/modules/colorLabel/views/colorLabelPopupView.js
```diff
@@ -0,0 +1,100 @@
...
+            if (!this.selected) return;
+            this.addItem();
+        },
+
+        preRender: function() {
+            this.selected = -1;
+        },
+
+        postRender: function() {
+            if (!this.parentView.model.has('_colorLabel')) return;
```
tomgreenfield: No need for `.has()`.
```js
var color = this.parentView.model.get('_colorLabel');

if (!color) return;

this.selected = color;
this.updateClasses();
```

frontend/src/modules/colorLabel/views/colorLabelPopupView.js
```diff
@@ -0,0 +1,100 @@
...
+            });
+        },
+
+        onApply: function(event) {
+            if (!this.selected) return;
+            this.addItem();
+        },
+
+        preRender: function() {
+            this.selected = -1;
```
tomgreenfield: `this.selected` could be `undefined` (on initialise), [`null`](https://github.com/adaptlearning/adapt_authoring/pull/1921/files#diff-4082fa8f4c93488e4a000c7ea57ed2dfR35) or [`-1`](https://github.com/adaptlearning/adapt_authoring/pull/1921/files#diff-4082fa8f4c93488e4a000c7ea57ed2dfR50), when there's no colour chosen.

Is there a reason for the three possible values?

Might be easier to have `selected: null` on the view and set / return to `null` when appropriate.

frontend/src/modules/editor/global/views/editorOriginView.js
```diff
@@ -10,6 +10,14 @@ define(function(require){
       'click .field-object .legend': 'onFieldObjectClicked'
     },
 
+    attributes: function() {
+      if (this.model && this.model.has('_colorLabel') && this.model.get('_colorLabel')) {
```
tomgreenfield: No need for `.has()`.
```js
var colorLabel = this.model && this.model.get('_colorLabel');

if (colorLabel) return { 'data-colorlabel': colorLabel };
```

frontend/src/modules/colorLabel/views/colorLabelPopupView.js
lc-thomasberger: I thought about allowing custom colors first ... will have to revisit.

lc-thomasberger:
Thanks @tomgreenfield added your suggestions as they improve the code.

frontend/src/modules/colorLabel/views/colorLabelPopupView.js
```diff
@@ -0,0 +1,97 @@
...
+            if (!color) return;
+            this.selected = color;
+            this.updateClasses();
+        },
+
+        applyOnParent: function() {
+            var color = this.selected;
+
+            if (this.selected) {
+                this.parentView.$el.attr('data-colorlabel', color);
```
tomgreenfield: Now you chosen `null` as the falsey value, this whole function be condensed into one line:
```js
this.parentView.$el.attr('data-colorlabel', this.selected);
```
Reference: http://api.jquery.com/attr/#attr-attributeName-value.

frontend/src/modules/colorLabel/views/colorLabelPopupView.js
```diff
@@ -0,0 +1,97 @@
...
+            });
+        },
+
+        onApply: function(event) {
+            if (!this.selected) return;
+            this.addItem();
+        },
+
+        preRender: function() {
+            this.selected = null;
```
tomgreenfield: `this.selected` can be declared on instantiation if you preferred to not use this function for it, like [`className`](https://github.com/adaptlearning/adapt_authoring/pull/1921/files#diff-4082fa8f4c93488e4a000c7ea57ed2dfR8) is:
```js
selected: null
```

tomgreenfield:
@lc-thomasberger, have you pushed your recent commit(s)?

routes/lang/en-application.json
```diff
@@ -311,5 +313,8 @@
   "app.article": "article",
   "app.block": "block",
   "app.component": "component",
-  "app.errorloadconfig": "Failed to load configuration settings for %{course}"
+  "app.errorloadconfig": "Failed to load configuration settings for %{course}",
+  "app.colourLabel": "Colour Label",
+  "app.colourLabelTitle": "Colour Label",
+  "app.colourLabelIntro": "The colours are <b>only</b> applied in the authoring tool and will <b>not affect the genereated course</b>. The settings are synced with all users in the authoring tool."
```
tomgreenfield: Should be `generated`.

PR #1923, opened 2018-03-20
Fixes #1901 reference fs-extra as fs

### LINKED ISSUE #1901: Tool contains references to both 'fs' and 'fse'
There are discrepancies in the way that the fs-extra module is referenced. It gets referenced in one of the following two ways.

`var fs = require('fs-extra');`
`var fse = require('fs-extra');`

We should change all references to this module to be `fs` for consistency.

dancgray:
Only seems to be referenced as fse in the one file.

### DIFF
--- a/plugins/output/adapt/index.js
+++ b/plugins/output/adapt/index.js
@@ -11,8 +11,7 @@ var origin = require('../../../'),
     database = require('../../../lib/database'),
     util = require('util'),
     path = require('path'),
-    fs = require('fs'),
-    fse = require('fs-extra'),
+    fs = require('fs-extra'),
     async = require('async'),
     archiver = require('archiver'),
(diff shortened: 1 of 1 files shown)


dancgray:
This has already been done in the latest changes to [framework-import](https://github.com/adaptlearning/adapt_authoring/blob/feature/framework-import/plugins/output/adapt/index.js#L14) branch. Do we need to bother with this PR?

canstudios-louisem:
If that's so shall we just close this? it will just cause extra conflicts.

dancgray:
Closing as changes included in [framework-import](https://github.com/adaptlearning/adapt_authoring/blob/feature/framework-import/plugins/output/adapt/index.js#L14)

Issue #1430, opened 2016-10-18
Add ability to reorder list items


mldaq:
I have a number of people on my team that are using the resources extension and one issue they are having is once they have created a number of resources (like 10) they is no easy way to reorder those items without deleting an doing a number items over. The suggestion would be to allow the designer to drag to reorder or have an arrow to move the item up or down in the order.

Thanks,
Mark

taylortom:
+1

dancgray:
+1 same goes for accordion and narrative items

taylortom:
There's a community plugin that make work for this: https://gist.github.com/tamagokun/95e76ee50f59a97696ce

mldaq:
@taylortom - Thanks, I'll check it out.

mldaq:
@taylortom - can you provide info on how I would use this or apply it?

taylortom:
@mldaq it was as much a note for the dev that picks this up as something directed at you. If you have the dev capacity though, I'm happy to guide.

SpacegoatHQ:
+1 same goes for accordion and narrative items

PR #1924, opened 2018-03-20
Scaffold refactor

tomgreenfield:
The principal purpose of this scaffold refactor is to provide a more consistent base layer to use going forward with Backbone Forms, or indeed any other means of rendering fieldsets.

### What’s changed?
* Fields from the properties schema are now properly passed through to Backbone Forms
  <details><summary>This includes:</summary>
  <ul>
  <li><code>confirmDelete</code></li>
  <li><code>default</code></li>
  <li><code>editorAttrs</code></li>
  <li><code>editorClass</code></li>
  <li><code>fieldAttrs</code></li>
  <li><code>fieldClass</code></li>
  <li><code>help</code></li>
  <li><code>inputType</code></li>
  <li><code>legend</code></li>
  <li><code>title</code></li>
  <li><code>titleHTML</code></li>
  <li><code>validators</code></li>
  </ul>
  </details>
* All editors given a spring clean for readability and code consistency
* ‘Backbone overrides’ split off into Handlebars templates where possible over hard-coded strings
* Switched to Selectize library for tags to improve the buggy editor

### What’s new?
The UI tweaks in this PR were designed not to be too far reaching since this requires a wider discussion for the next stage of the refactor. [This issue](https://github.com/adaptlearning/adapt_authoring/issues/1430) is an example that should probably be deferred to the next release once we decide on the mechanism.
* Spacing improved for consistency
* Tooltips added with help text, where available, and the underlying attribute name for alignment with documentation and to aid troubleshooting
* Reset button added to revert a field, where possible, to its default state

scaffold</details>

### What’s been removed?
* The custom Question Button editor has been replaced with the standard Text editor out since this was only compatible with version 1 of the framework
* The custom Boolean true/false dropdown has been replaced with the Checkbox editor for consistency. All core plugin schema have already been switched over
* The undocumented & unused custom `TextArea:blank` editor has been retired


frontend/src/modules/scaffold/backboneFormsOverrides.js
```diff
@@ -1,292 +1,325 @@
...
+  Backbone.Form.Field.prototype.templateData = function() {
+    return _.extend(templateData.call(this), this.schema);
+  };
+
+  // process nested items
+  Backbone.Form.editors.List.Modal.prototype.itemToString = function(value) {
+    var createTitle = function(key) {
+      var context = { key: key };
+
+      return Form.Field.prototype.createTitle.call(context);
```
lc-thomasberger: I think this should be: `Backbone.Form.Field.prototype.createTitle.call(context);`

lc-thomasberger:
Nice one! I really like the tooltips. Makes the UI much cleaner. 

I like the reset Buttons, but they are quiet a few. Is it possible to only show the button when the input is different to the default value? We also use title attributes to add tooltips for ui elements in the AT. I think this will help to understand the icon.

frontend/src/modules/scaffold/views/scaffoldAssetView.js
```diff
@@ -209,96 +182,100 @@ define(function(require) {
...
+            contentTypeId: model.get('_id') || '',
+            contentType: model.get('_type'),
+            contentTypeParentId: model.get('_parentId') || Origin.editor.data.course.get('_id'),
             fieldname: data.assetFilename,
             assetId: data.assetId
           };
+
           // all ScaffoldAssetViews listen to the autofill event, so we trigger
           // that rather than call code directly
           // FIXME only works with graphic components
```
canstudios-louisem: Any chance of looking at this FIXME while your here?

tomgreenfield: The associated code it's referring to is actually part of [Asset Management](https://github.com/adaptlearning/adapt_authoring/blob/develop/frontend/src/modules/assetManagement/views/assetManagementModalView.js#L19). Would you like me to remove the comment?

canstudios-louisem: May as well if its not relevant anymore.

lc-thomasberger:
The tags input is wider then the rest of the input fields in project settings:
image

lc-thomasberger:
@tomgreenfield just tested adding `editorAttrs` to the schema and it's added to the views schema attribute. 👍 

So is this used to pass attributes to editors? For instance if I create a range input type I can pass the value for `minimum`, `maximum` and `step` value?

```json
"_percentInviewVertical": {
          "type": "number",
          "required": false,
          "default": 50,
          "editorAttrs": {
            "minimum": 1,
            "maximum": 100,
            "step": 1
          },
          "isSetting": true,
          "title": "Percent in view",
          "inputType": "RangeSlider",
          "validators": ["number"],
          "help": "Controls what percentage of the component's height needs to be in the viewport in order for the CSS class(es) to be added."
}
```

tomgreenfield:
Yep, looks like you answered your own question, @lc-thomasberger!

Full list of fields is in my original [comment](https://github.com/adaptlearning/adapt_authoring/pull/1924#issue-176158035).

tomgreenfield:
Just so we don't forget, your latest commits, @taylortom, have killed the `Extensions` label in Plugin Management.

taylortom:
Re-added missing label in https://github.com/adaptlearning/adapt_authoring/pull/1940.

dancgray:
Get the following error if the course has no extensions enabled:

Uncaught TypeError: Cannot convert undefined or null to object schemas.js:23

taylortom:
@lc-thomasberger do you have any time in the near future to do a retest? Only waiting on one more approval 🎉

PR #1925, opened 2018-03-20
restyle options 

lc-thomasberger:
to improve alignment for collapse and expand articles


lc-thomasberger:
merge this one in directly as it is reviewed in the PR for color labels

PR #1926, opened 2018-03-21
Allow super admins to delete any project

### LINKED ISSUE #1922: Super admins should be able to remove any course in the system
## Affected area
We're currently unable to clean up old courses unless either the course is shared, or the owner of the course logs in and deletes it themselves.

## Requested Feature
A Super Admin should be able to see a list of all courses in the system, and have permission to update/delete these.

## Use Case
This is an issue we frequently face when employees leave an organisation for example.

## Current Workaround
A Super Admin must change the password of the course owner's account, log in, and delete the course.

taylortom:
Fixes #1922.


plugins/content/course/index.js
```diff
@@ -39,103 +79,26 @@ var DASHBOARD_COURSE_FIELDS = [
...
-            res.statusCode = 500;
-            return res.json(err);
-          }
-          return res.json(results);
-        });
+  app.once('serverStarted', function (server) {
+    rest.get('/all/course', function(req, res, next) {
+      console.log(usermanager.getCurrentUser()._id, 'delete', permissions.buildResourceString('*', '/*'));
+      permissions.hasPermission(usermanager.getCurrentUser()._id, 'delete', permissions.buildResourceString('*', '/*'), function(error, hasPermission) {
+        console.log(error, hasPermission);
```
lc-thomasberger: can you please remove the `console.log`'s

lc-thomasberger:
Tested and works for the "shared courses" as well as "all projects" view.

I think it would be useful if we can group courses by the author for the shared and all projects view.

taylortom:
@lc-thomasberger, could we do this as a sort, or are you thinking of something else? 
Should probably become a new issue, and maybe also apply to shared courses.

lc-thomasberger:
I thought about a group action. So I guess this would be sorting. 

Will create a new issue.

lib/helpers.js
```diff
@@ -135,39 +137,56 @@ function isMasterPreviewAccessible(courseId, userId, next) {
...
+        }
+        if(hasSuperPermission) { // user with full CRUD permissions, so allow
+          return next(null, true);
+        }
+        if(course.createdBy.toString() === userId) { // user's own course
+          return next(null, true);
+        }
+        if (course._isShared) { // shared universally
+          return next(null, true);
+        }
```
lc-thomasberger: The code is pretty descriptive so I think the comments are not required.

taylortom:
Nah, they should probably be the same

lc-thomasberger:
It is a relict from SharedProjectView

canstudios-nicolaw:
If I go to All courses view and I click the context menu on a course with is owned by another super admin and is not shared, then nothing happens. The context menu seems to appear for my courses and shared courses.

Also 'Dashboard' in the breadcrumbs always goes back to the 'My courses' dashboard rather than the dashboard I was most recently on. If it's an easy fix I think it's worth making it go to the most recent dashboard?

canstudios-nicolaw:
Thanks @taylortom I can see the context menu now. 

I've just noticed that on the 'All courses' dashboard, if I attempt to scroll down and load the next batch of courses, I get a content permission error - "You are not permitted to do that".

The initial load of all courses works fine, it's only the second batch that has the problem.

taylortom:
@canstudios-nicolaw I'm not able to replicate that one, is there any more debug info?

canstudios-nicolaw:
@taylortom Nothing is appearing in the logs, it seems to happen when it tries to load the last batch of courses. I will try and investigate this further, I have done an npm install, grunt build and a restart so I'm not sure why this would be different for me.

geterror
error2

It would be interesting to know if this happens for other people or if it's just me.

taylortom:
Weird. It's probably an edge case-type problem with a specific course; will try harder to see if I can replicate it.

lc-thomasberger:
I cant really replicate the issue. I created 37 courses and had no issues fetching them.

canstudios-nicolaw:
@taylortom I've worked out the problem - it's because there was an old course in my db which had a different tenant id. The error was coming from line 172 in lib/helpers.js.

I think this is acceptable for now as tenants are an unreleased feature and I doubt this will be the only check of this kind.

taylortom:
Coolio, I shall log an issue, and fix it for the next release, just to be safe.

PR #1934, opened 2018-04-05
Don't fail on plugin install error

### LINKED ISSUE #1890: Install should not fail if a plugins framework requirement does not match the framework installed
## Expected Behaviour
Installing the AT should not fail if one of the plugins installed from adapt install is not compatible with the framework version installed.

It should just not install the plugin and log an error.

## Actual Behaviour
AT install fails with the following error.
```
ERROR: Unable to install adapt-contrib-spoor (undefined) as it is not supported in the current version of the Adapt framework (2.2.5)

Failed to create master tenant. Please check the console output.
```
Spoor v2.1.4 version number matches the adapt.json version defined in adapt.json in FW v2.2.5 but the [required Framework version](https://github.com/adaptlearning/adapt-contrib-spoor/blob/v2.1.4/bower.json#L4) defined in spoor bower.json is FWv3

## Steps to Reproduce
1. Download AT v0.4.1 and install using a framework tag/branch (v2.2.5) that installs a plugin (spoor v2.1.4)

## Versions
  - Authoring Tool Version: v0.4.1
  - Framework Version: v2.2.5

taylortom:
Fixes #1890

### DIFF
--- a/lib/installHelpers.js
+++ b/lib/installHelpers.js
@@ -409,10 +409,12 @@ function updateFrameworkPlugins(opts, callback) {
     }
     var plugins = Object.keys(json.dependencies);
     async.eachSeries(plugins, function(plugin, pluginCallback) {
-      var _done = function() {
+      var _done = function(error) {
         hideSpinner();
-        // don't fail on error, just continue with next plugin (see #1890)
-        pluginCallback.apply(this);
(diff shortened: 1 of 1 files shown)


taylortom:
Have added a warning log on error

PR #1766, opened 2017-10-30
Add test data cache

taylortom:
Fixes #1761

PR #1938, opened 2018-04-06
Migrations can be created and are ran during upgrade

### LINKED ISSUE #1937: Migrations
## Affected Area
Upgrade

## Requested Feature
The ability to write scripts that will run on upgrade.

## Use Case
This will allow us to make changes to the DB structure and folder structure of the app without breaking existing installs.

## Current Workaround
None

## Additional Information
It would be useful to get this into the next version even if there are no migrations as it will mean we can use in the version after.

canstudios-louisem:
## Migrations
Migration files are simple they export a down and an up function. Up should make the changes you want and return the callback. Down should revert changes made by up if possible, this allows easier testing and downgrading of a release if needed.

The migration library works by looking in a folder for migration files and comparing the files there to a DB table where it stores the ran state. Any migrations that have not been run will have a state of down and it will run them. 

The migration lib needs a DB connection string because every install can be different its easiest to generate this into a config file for it. I've done this on grunt build that way if someone updates the DB details they just need to run grunt build.

Migrations are not run on install but any new ones will run on upgrade. On install, we need to add any existing ones and mark them as done.

Unfortunately, the release where we add this we cant run migrations because each release uses the last release's upgrade file so it would be good to get this in asap.

### To create a migration
To create a migration you can run `npm run migration create <name>`. This will create a new file from the template in `migrations/libs`. 

### To run a migration from the command line (to test)
To create a migration you can run `npm run migration up <name>`,  `npm run migration down <name>` will revert it.

Fixes #1937

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 /conf/config.json
-
+/conf/migrate.json
 /data
 
 /frontend/build
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -234,6 +234,38 @@ module.exports = function(grunt) {
     return config.serverPort;
   };
 
+  grunt.registerTask('migration-conf', 'Creating migration Conf', function() {
+    var config = grunt.file.readJSON('conf/config.json');
+    var connectionString = '';
+
+    // Construct the authentication part of the connection string.
+    var authenticationString = config.dbUser && config.dbPass ? config.dbUser + ':' + config.dbPass + '@' : '';
+
+    // Check if a MongoDB replicaset array has been specified.
+    if (config.dbReplicaset && Array.isArray(config.dbReplicaset) && config.dbReplicaset.length !== 0) {
+      // The replicaset should contain an array of hosts and ports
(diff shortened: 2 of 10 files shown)


canstudios-louisem:
Looks like migrate-mongoose is not compatable with node 4 hence th failing tests are we happy to stop testing that now? We havent supported it for a while.

dancgray:
Seems to work pretty well with the commands to create a template: `npm run migrate create <name>`, and to run a migration: `npm run migrate up <name>`.

taylortom:
Works well for me. Can we add an `fs.ensureDir` to make sure there's a `lib` folder in migrations? This threw an error for me.

Also a side-note for others testing: you may need to run `grunt migration-conf` if you're just checking out the branch in a current install.

taylortom:
Note: I remove Node 4 and 5 in https://github.com/adaptlearning/adapt_authoring/pull/1766, so the broken test shouldn't stop us merging.

Issue #1495, opened 2017-02-02
Ability to edit an applied theme.


taylortom:
Doesn't need any introduction!

taylortom:
Fixed as a part of #1116

PR #1940, opened 2018-04-12
Expose theme and menu pluginLocations to editor

### LINKED ISSUE #1116: Authoring tool does not surface attributes defined in properties.schema for menu plugins
Some menus will have editable attributes e.g https://github.com/cgkineo/adapt-cover-menu

taylortom:
### I've merged Tom's changes from #1924 into this PR, so that needs to be merged prior to this

Fixes #1116 and #1495.

This is a halfway-house on the way to full theming, and allows the editing of any property specified on the pluginLocations object for any theme or menu (as we already have for extensions). 

**Testing**<br/>
This can be tested with my forks of [boxMenu](https://github.com/taylortom/adapt-contrib-boxmenu) and [vanilla](https://github.com/taylortom/adapt-contrib-vanilla) (please also test the plugin management area, as that was affected by my changes).

**To-do**
- Nest theme settings under theme name as with extensions/menu settings

**Notes**
- We had duplicate copies of the plugin models and collections, so these have been removed, and plugin management changed to use those in core
- We now load all plugin types on load, rather than just components and extensions
- _**Side-note**: I'm not a fan of the **Themesettings** and **Menusettings** labels, but this is an issue with the way that the titles are [generated from object keys](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/core/helpers.js#L35-L42) rather than using an arbitrary language string, and should be fixed separately._

screen shot 2018-04-12 at 16 50 22


taylortom:
@tomgreenfield could you have a look over the [`schemas.js`](https://github.com/adaptlearning/adapt_authoring/pull/1940/files#diff-07cccea11d98f825679934e639a08f16) and [`ScaffoldColorPickerView.js`](https://github.com/adaptlearning/adapt_authoring/pull/1940/files#diff-c1c9a96f1be8cffeb41929cd53609a4b) files, as these will likely conflict with your scaffold branch (can't remember exactly what I needed to change for the latter, but will look it up if needed).

taylortom:
So far we have one known issue when creating a new course, which I'm looking into

lc-thomasberger:
I think so far there is no way for the color picker to reset the color, right?

Also I think the used color-picker is a bit hard to use. It took a while that I understood that I must click on the "colored sphere" at the bottom right to select my color.
I did a quick research on other more user friendly plugins:
https://www.cssscript.com/demo/color-picker-component-pickr/
https://www.cssscript.com/demo/color-picker-alpha-selection/
http://bgrins.github.io/spectrum/

lc-thomasberger:
Feature request: 😃 
it would be great to have course specific color swatches. Ideally in Project-settings a author could add colors that can then be reused from a color-picker. If you change the color of that swatch item, it get's automatically updated. This makes sure that courses use the exact brand-colors.

taylortom:
Have opened a new issue for the colour picker upgrade: #2014

frontend/src/modules/pluginManagement/views/pluginManagementView.js
```diff
@@ -3,50 +3,43 @@ define(function(require){
...
-      this.collection = new (this.pluginCollections[this.pluginType])();
-      this.listenToOnce(this.collection, 'sync', this.renderPluginTypeViews);
-      this.collection.fetch();
+      return OriginView.prototype.initialize.apply(this, arguments);
     },
 
     preRender: function() {
       Origin.trigger('location:title:update', { title: Origin.l10n.t('app.' + this.pluginType + 'management') });
-      this.initialiseCollection();
+      this.getColl().fetch({ success: _.bind(this.renderPluginTypeViews, this) });
```
tomgreenfield: Can use native `.bind()`.

Issue #1704, opened 2017-08-15
Update npm dependencies


taylortom:
Need to make sure we support current/supported versions of our npm deps.

Particular issue with:
- grunt-casperjs v2 which errors (v1 looks to work ok).
- nodemailer no longer supports node v4

It's been suggested that we cut out as much dead wood as we can to keep these kind of issues to a minimum.

taylortom:
Related to #1440

canstudios-louisem:
Their are a few tools that we can use to help upgrade npm deps easier https://www.npmjs.com/package/next-update and greenkeeper.io both look useful.

canstudios-louisem:
Related to #1121

canstudios-louisem:
Something to be aware of, in with npm 5 (standard in node 8) npm now creates a `package-lock.json` similar to `composer.lock` and is intended to be committed into VCS.

I have not used it yet and am unsure of its advantages of YARN but if we did not wish to require another global package it may be a good option.

canstudios-louisem:
Could we suppress the depreciation warnings from NPM they seem to cause some users undue worry?

taylortom:
@tomgreenfield assigned this to you for visibility -- this was done as part of your PR, right?

tomgreenfield:
@taylortom, naturally.

PR #1969, opened 2018-05-29
Add `package-lock.json` and so remove it from `.gitignore`

LoicMahieu:
Package-lock to ensure a consistent install and compatible dependencies. Its purpose is to avoid situation where installing modules from the same package.json results in two different installs.
This change is not a breaking change due that older versions of NPM (< 5.x.x) will not use it.
IMO, this change is essential for going forward and make `adapt_authoring` code base and dependencies up-to-date in the future.

Related #1704
Related #1873

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9408 @@
+{
+  "name": "adapt_authoring",
+  "version": "0.4.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@ladjs/i18n": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -0,0 +1,9408 @@
+{
```
tomgreenfield: Please refresh thie file for the current dependencies.

LoicMahieu:
Hi!
I am not using adapt learning any more. Feel free to pick this PR on your own.

PR #1971, opened 2018-05-30
Add new config `dbConnectionUri`

LoicMahieu:
It will allow developer to specific himself the whole "connection string". Indeed, there are some case that are not covered by the config, simple example: `ssl=true` that could be append to the query of the string.
I don't think adding a new config key for each case is a good idea.

A new dependency has been introduced `mongodb-uri`. Indeed, in case where `connect(db)` with a db, we need to change the database name in the connection string.

Related #1978

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,3 +1,6 @@
+
+var mongoUri = require('mongodb-uri');
+
 // LICENCE https://github.com/adaptlearning/adapt_authoring/blob/master/LICENSE
 module.exports = function(grunt) {
   require('matchdep').filterAll('grunt-*').forEach(grunt.loadNpmTasks);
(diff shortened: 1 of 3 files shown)


dancgray:
Agree that the current implementation for connection string is too rigid. Plus some of the database values held in the tenant collection are not used. 

Only problems: 
1 - that dbName is usually part of the mongodb-uri but is also used throughout the app. So dbName would have to entered on install in addition to the name in the nongodb-uri. 
2 - install input for dbHost, dbPort etc becomes redundant, but config values for these are still checked in parts of the install process when the app starts and does a [checkConnection](https://github.com/adaptlearning/adapt_authoring/blob/develop/lib/database.js#L506)

LoicMahieu:
Indeed this PR does not work well with the install script. And also, the `dbName` and `dbConnectionUri` seems to be redundant. In fact, the database name in the `dbConnectionUri` is never used since it will always be overwritten by the `dbName`.

I will be happy to discuss with you on issue #1978 about possibilities.

taylortom:
@dancgray was this added in v0.6.1? Think it can probably be closed along with #1978

dancgray:
@taylortom, yes this was added in PR #2015.

PR #1992, opened 2018-06-19
use prompt override

### LINKED ISSUE #1991: upgrade should allow cli arguments
# Feature Request 

## Affected Area
upgrade.js

## Requested Feature
Since optimist is required in the upgrade script it should be used like in the install script to pass default settings as cli arguments.

## Use Case
better integration of the upgrade script into automation or other tools.

## Current Workaround
none, only manual upgrade

lc-thomasberger:
fixes #1991

### DIFF
--- a/lib/installHelpers.js
+++ b/lib/installHelpers.js
@@ -27,7 +27,8 @@ var inputHelpers = {
   numberValidator: /^[0-9]+\W*$/,
   alphanumValidator: /^[A-Za-z0-9_-]+\W*$/,
   toBoolean: function(v) {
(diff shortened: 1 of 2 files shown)


tomgreenfield:
I cannot get past the first question in Git Bash on Windows 10 running Node v8.11.2:
```shellsession
$ node upgrade

This script will update the Adapt authoring tool and/or Adapt Framework. Would you like to continue?
> Continue? Y/n (Y)

Bye!
```

lc-thomasberger:
@tomgreenfield same setup, same issue ... have only tested with the cli arguments. Will have a look at it.

PR #1994, opened 2018-06-19
Strip unused grunt tasks

### LINKED ISSUE #1993: Strip Gruntfile
There's lots of stuff in there that we don't use, and various node modules that are very very out of date.

taylortom:
Fixes #1993


package.json
```diff
@@ -109,8 +101,7 @@
     "mocha": "^5.2.0",
     "mocha-multi": "^1.0.1",
     "mocha-simple-html-reporter": "^1.1.0",
-    "mongodb": "^3.0.8",
-    "phantom": "^4.0.12",
+    "mongodb": "3.0.10",
```
tomgreenfield: Caret `^` shouldn’t have been removed.

PR #1997, opened 2018-06-20
orphaned element 

### LINKED ISSUE #1986: import can create orphaned elements
# BUG 

## Expected Behaviour
Import should only accept valid courses.

## Actual Behaviour
When id's can't be mapped, the element is still imported.
This may result in elements being not removed when there course is deleted.

## Steps to Reproduce
1. import a course with orphaned elements
2. search for "Cannot update 5a6aff09b382e6db06bd8c68._parentId, 5a6afeeeb382e6db06bd8c67 not found in idMap"
3. Delete the course, the orphaned element should still be stored in the database.

## Versions
  - Authoring Tool Version: v0.5.0-rc1
  - Framework Version: --
  - Operating System: --
  - Browser: --

lc-thomasberger:
fixes #1986 

## to test:
Import the demo course and change an elements _parentId or _id.
For instance if you change an articles _id attribute, it gets imported but all it's blocks and components are not imported. The id's of this elements should be listed in the frontends server log.
Inspect the imported course in the authoring tool. It should have created the page and the article without the blocks and components.


dancgray:
This works for me, however I'm not keen on the error messages being at the bottom. With a smaller window they are off screen and the user does not see any message. For the time being I prefer the alert.

taylortom:
👍 for keeping an alert.

Maybe a good solution would be to summarise the issue in the alert, with some text letting users know more information is at the bottom (or wherever).

plugins/output/adapt/importsource.js
```diff
@@ -86,9 +87,13 @@ function ImportSource(req, done) {
       async.apply(installPlugins),
       async.apply(addAssets, formTags),
       async.apply(cacheMetadata),
-      async.apply(importContent, formTags),
-      async.apply(helpers.cleanUpImport, cleanupDirs)
-    ], done);
+      async.apply(importContent, formTags)
+    ], function(importError, result) {
+      helpers.cleanUpImport(cleanupDirs, function(cleanUpError) {
```
taylortom: Can we get a comment here stating that `cleanUpImport` should be called regardless of whether the control function passes or fails - my brain couldn't process this 😂

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -75,23 +75,24 @@ define(function(require){
...
+      var title = data.responseJSON.title || Origin.l10n.t('app.importerrortitle');
+      var msg = data.responseJSON.body.replace(/\n/g, "<br />") || error;
+      this.promptUser(title, msg, true);
+    },
+    
+    promptUser: function(title, message, isError) {
       Origin.trigger('sidebar:resetButtons');
-      var text = (data.responseText) ? data.responseText.replace(/\n/g, "<br />") : error;
-
-      Origin.Notify.alert({
```
taylortom: Can we keep an alert, but just summarise the error, and instruct the user to view the full details in the `.server-log`.

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -75,22 +75,22 @@ define(function(require){
       Origin.router.navigateToHome();
     },
 
-    onFormSubmitSuccess: function(importData, impoprtStatus, importXhr) {
+    onFormSubmitSuccess: function(data, impoprtStatus, importXhr) {
```
tomgreenfield: Could you fix that `impoprtStatus` typo while you're here?

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -75,22 +75,22 @@ define(function(require){
...
       Origin.router.navigateToHome();
     },
 
-    onAjaxSuccess: function() {
-      this.goBack();
-    },
-
     onAjaxError: function(data, status, error) {
+      var title = data.responseJSON.title || Origin.l10n.t('app.importerrortitle');
+      var msg = data.responseJSON.body.replace(/\n/g, "<br />") || error;
```
tomgreenfield: Query: the original code had an additional truthy check for the response before doing a `.replace()`. Is it fine that this has been removed?

taylortom: I think this is only present in the case of a JSON response (duh?), or when `json` is set as the `$.ajax` `dataType`. As we also use [`Express.send`](https://expressjs.com/en/api.html#res.send), we're not guaranteed to **send** a JSON response back, so the return type may vary.

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -75,22 +75,23 @@ define(function(require){
...
+    onFormSubmitSuccess: function(data, importStatus, importXhr) {
       Origin.router.navigateToHome();
     },
 
-    onAjaxSuccess: function() {
-      this.goBack();
+    onAjaxError: function(data, status, error) {
+      var resJson = data.responseJSON || {};
+      var title = resJson.title || Origin.l10n.t('app.importerrortitle');
+      var msg = resJson.body && resJson.body.replace(/\n/g, "<br />") || error;
```
lc-thomasberger: should this be:
`var msg = (resJson.body && resJson.body.replace(/\n/g, "<br />")) || error;`

tomgreenfield:
Please squash when merging.

PR #1999, opened 2018-06-21
double scrollbar fix

lc-thomasberger:
handle scrollbar and content height in ContentPaneView

To help with collapsible articles in pageEditor, overflow-y is set to scroll to always show the scrollbars.

fixes #1998

PR #2000, opened 2018-06-21
fix scroll-position in pageEditor

### LINKED ISSUE #1814: scroll position in page editor

lc-thomasberger:
fixes #1814 

css transitions not fired when articles are collapsed (display-none).
Use velocity to animate the blocks and trigger the events to detect that all children are ready.


taylortom:
Doesn't seem to be restoring the position for me

taylortom:
Changing the scroll div from `.contentPane` to `.editor-view` should fix the problem

taylortom:
@tomgreenfield haven't touched the velocity code you commented on, as your suggested fix stopped the inner block content from showing, and wasn't sure of its initial intent.

tomgreenfield:
Sorry @taylortom, that's a typo – I meant use a `begin` function. The inner context text doesn't animate properly without this amend.

taylortom:
Should be good to go

lc-thomasberger:
@taylortom have you tested this PR without https://github.com/adaptlearning/adapt_authoring/pull/1999 being merged?
`.contentPane` should be the scrollcontainer. If I test this PR (with the latest changes from develop) the scroll position is not restored. If I change it to `.contentPane` it works.

taylortom:
Cool, no, had no idea it was dependent on anything else.

PR #2002, opened 2018-06-22
Escape tag content in helper function, fixes #2001

### LINKED ISSUE #2001: Tags in asset preview are escaped (develop branch)
## Expected Behaviour
Tags should render correctly in the asset manager preview. Content within tags should be escaped.

## Actual Behaviour
The whole generated tag HTML is escaped:
screen shot 2018-06-22 at 16 27 12

## Steps to Reproduce
1. add a tag to an image and preview in the asset manager

## Versions
  - Framework Version: any
  - Operating System: any
  - Browser: any

dancgray:


frontend/src/core/helpers.js
```diff
@@ -154,10 +154,10 @@ define(function(require){
       }
       var html = '<ul class="tag-container">';
       for (var i = 0; i < list.length; ++i) {
-        var tag = (key && list[i][key]) ? list[i][key] : list[i];
+        var tag = Handlebars.Utils.escapeExpression((key && list[i][key]) ? list[i][key] : list[i]);
```
tomgreenfield: Would you mind switching out the ternary to help a bit with readability?
```js
var item = list[i];
var tag = Handlebars.Utils.escapeExpression(key && item[key] || item);
```

taylortom: I'd prefer even moving all logic to the item declaration, and just have the tag use escapeExpression
```
var item = list[i][key] || list[i];
var tag = Handlebars.Utils.escapeExpression(item);
```

PR #2011, opened 2018-07-03
Expose CKEditor‘s extraAllowedContent

### LINKED ISSUE #1619: CKEditor is stripping out attributes and styles
this issue is easy to replicate by adding the data attribute to a span in the source editor of a component, for example:
`<span class="tidbit-trigger tidbit-color" data-bitnum="0">your text here</span>`

if you edit anything in the component a second time, the data attribute will get stripped

based on the [CKEditor documentation](http://sdk.ckeditor.com/samples/acf.html), CKEditor filters incoming HTML content by transforming and deleting disallowed elements, attributes, classes and styles.

taking what is already in the code and adding two things, i have a solution. change this line: https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/core/scaffold/scaffoldOverrides.js#L125

to this:
`extraAllowedContent: 'span(*)[*]{*}',`

please note that this might only fix the span tag ([see this post](https://community.adaptlearning.org/mod/forum/discuss.php?d=1857#p9032)). there's [another post in the forum](https://community.adaptlearning.org/mod/forum/discuss.php?d=1849) that appears to be doing this to ordered list tags.

the editor documentation shows that specific things can be allowed or disallowed. there is also a way to allow everything... so maybe changing line 125 to this might also fix the ol tags? i haven't tested this theory, just putting it out there:
`allowedContent = true`

lc-thomasberger:
Use `ckEditorAllowedContent` config option to allow custom ckeditor filtering.
The config option is not set during installation, let me know if you think it makes sense to add it.

Resolves #1619.

### DIFF
--- a/frontend/src/modules/scaffold/backboneFormsOverrides.js
+++ b/frontend/src/modules/scaffold/backboneFormsOverrides.js
@@ -196,7 +196,7 @@ define([
         dataIndentationChars: '',
         disableNativeSpellChecker: false,
         entities: false,
-        extraAllowedContent: Origin.constants.ckExtraAllowedContent,
+        extraAllowedContent: Origin.constants.ckEditorExtraAllowedContent,
         on: {
           change: function() {
             this.trigger('change', this);
--- a/frontend/src/modules/scaffold/backboneFormsOverrides.js
+++ b/frontend/src/modules/scaffold/backboneFormsOverrides.js
@@ -196,7 +196,7 @@ define([
         dataIndentationChars: '',
         disableNativeSpellChecker: false,
         entities: false,
-        extraAllowedContent: 'span(*)',
+        extraAllowedContent: Origin.constants.ckEditorAllowedContent || 'span(*)',
         on: {
           change: function() {
             this.trigger('change', this);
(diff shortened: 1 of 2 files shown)


lib/configuration.js
```diff
@@ -31,7 +31,8 @@ var ALLOWED_CLIENT_SIDE_KEYS = [
   'useSmtp',
   'useffmpeg',
   'isProduction',
-  'maxLoginAttempts'
+  'maxLoginAttempts',
+  'ckEditorAllowedContent'
```
tomgreenfield: `ckEditorAllowedContent` is misleading as there is a [corresponding option](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-allowedContent) in CKEditor.

frontend/src/modules/scaffold/backboneFormsOverrides.js
```diff
@@ -196,7 +196,7 @@ define([
         dataIndentationChars: '',
         disableNativeSpellChecker: false,
         entities: false,
-        extraAllowedContent: 'span(*)',
+        extraAllowedContent: Origin.constants.ckEditorAllowedContent || 'span(*)',
```
tomgreenfield: Just to say that If the user wanted to allow _no_ extra content `''`, they wouldn't be able because it would fall back to `'span(*)'`.

PR #2012, opened 2018-07-04
make modal screen wider

lc-thomasberger:


lc-thomasberger:
close this one, we will have a styling refactor in the future

PR #2013, opened 2018-07-09
Localise fieldset filter labels

### LINKED ISSUE #1376: Remove dependency between form fieldset titles and the data key
The code's set up at the moment like so:
- Fieldset `legend` is taken from the related database key (e.g. `_buttons`), and essentially just capitalised. This is probably fine as a default fall-through solution, but breaks for keys with multiple words (e.g. `_settingsName` gives you **Settingsname**). Think we should allow custom values here.
- The sidebar fieldset filter buttons hide the fieldsets using their `legend` value. This seems wrong to me, as the legend is a human-readable value -- we should probably use the key (i.e. `_settingsName`).

taylortom:
Fixes #1376.


lc-thomasberger:
I think Themesettings (-> Theme settings or Theme-settings) and Menusettings (-> Menu settings or Menu-settings) should also be added to the lang file.

PR #2015, opened 2018-07-10
Add dbConnectionUri for database connection

### LINKED ISSUE #1978: Database config: unable to specify query options
We would like to use MongoDB Atlas as mongodb backend. There are some issues with it:

- Use `SSL` as transport layer. So `ssl=true` should be passed as querystring of the connection string.
- Use of replica set. Currently, the config `dbReplicaset` is supported but Atlas also needs a `replicaSet=<shard>` in the querystring. (I don't know if it is stricly necessary).

The service recommends these kind of URI:
```
mongodb://<USER>:<PASSWORD>@project-name-shard-00-00-yw72i.mongodb.net:27017,project-name-shard-00-01-yw72i.mongodb.net:27017,project-name-shard-00-02-yw72i.mongodb.net:27017/test?ssl=true&replicaSet=project-name-shard-0&authSource=admin&retryWrites=true
```

There is already a few config about the database:
- `config.dbUser`
- `config.dbPass`
- `config.dbReplicaset`
- `config.dbName`
- `config.dbPort`
- `config.dbHost`

IMO there is 2 possibilities:

1. Add new config `dbQueryString` for being able to specify the querystring which will be appended to the connection string.
2. Add new config `dbConnectionUri` for being able to specify the whole connection string manually. Since `adapt_authoring` could use different database in runtime, we will have to parse the connection string for change the db name at runtime.

dancgray:
Fixes #1978 
## Proposed changes

Add a single connection URI to install, config.json and and required changes in mongoose and lib/database.js.

This will enable advanced users to construct a connection URI and add any connection [options](https://docs.mongodb.com/manual/reference/connection-string/#connection-string-options) they require.

On install it defaults to the current connection config values. The user is offered the option to add a connection URI, if response if yes only dbConnectionUri is populated along with dbName. 

Allows the user of replica sets. Allows connections to MongoDB Atlas.

There are a number or related outstanding issues:
1. It is possible to set a config value for dbReplicaset which is used in lib/dml/mongoose.js. This is now redundant and could be removed?
2. Same goes for dbOptions
3. The tenant collections holds details of the database connection. However only dbName appears to be used. This should probably be resolved as part of the multi tenancy release.


install.js
```diff
@@ -326,6 +348,29 @@ function configureServer(callback) {
   });
 }
 
+function configureDatabase(callback) {
+  async.series([
```
lc-thomasberger: Is it required to use `async.series` for a single function call?

install.js
```diff
@@ -326,6 +348,29 @@ function configureServer(callback) {
...
+    function database(cb) {
+      installHelpers.getInput(inputData.database.dbConfig, function(result) {
+        addConfig(result);
+        if(!result.useConnectionUri || USE_CONFIG && configResults.useConnectionUri !== 'y') {
+          installHelpers.getInput(inputData.database.configureStandard, function(result) {
```
tomgreenfield: Would this be slightly quicker to digest with a couple of variables?
```js
var isStandard = !result.useConnectionUri || USE_CONFIG && configResults.useConnectionUri !== 'y';
var config = inputData.database[isStandard ? 'configureStandard' : 'configureUri'];

installHelpers.getInput(config, function(result) {
  addConfig(result);
  callback();
});
```

PR #2020, opened 2018-07-16
sequential locking on sub menus

lc-thomasberger:
uses `editorAttrs` to customize the behavior of the inputType.

### DIFF
--- a/frontend/src/modules/scaffold/backboneFormsOverrides.js
+++ b/frontend/src/modules/scaffold/backboneFormsOverrides.js
@@ -79,7 +79,7 @@ define([
 
     _.each(selectedFields, function(key) {
       var fieldSchema = schema[key];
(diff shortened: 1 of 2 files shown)


taylortom:
Does this supersede https://github.com/adaptlearning/adapt_authoring/pull/1647?

lc-thomasberger:
If you are happy with the `editorAttr` then it does.

tomgreenfield:
For a future release, suggest adding a `data-adapt-type` or similar to the parent view, by checking `this.model.get('_type')` [here](https://github.com/adaptlearning/adapt_authoring/blob/develop/frontend/src/modules/editor/global/views/editorOriginView.js#L14-L19). Could then maybe hide / grey out fields with classes.

lc-thomasberger:
@tomgreenfield yes we need some sort of standardized way about how to deal with customizing editors.
Showing editors only for certain types, as well as collapsing editors (= reduces page size) would be great for the start.
I think we should do this already at the start and maybe defer this PR to the next release.

I would not grey out items. If a attribute is not available for a certain type, it should not be displayed at all.

Issue #2021, opened 2018-07-17
Change your own email via the profile view 


canstudios-joes:
## Expected Behaviour

The email input on the profile page should be editable.

## Actual Behaviour

The email input only displays your current email address and can not be changed.

## Steps to Reproduce
1. Access the profile view by selecting your email at the top right of the dashboard.
1. Try to change the email displayed.

## Versions
  - Authoring Tool Version: 0.5.0
  - Framework Version: N/A
  - Operating System: Linux Ubuntu 14.04
  - Browser: Google Chrome

lc-thomasberger:
@canstudios-joes any progress on this?

canstudios-nicolaw:
@lc-thomasberger I will inherit this work as Joe no longer works here

lc-thomasberger:
Thanks for the update @canstudios-nicolaw .

PR #2022, opened 2018-07-18
Change email via the profile view

canstudios-joes:
## Proposed changes

For #2021 - users can now change their own email address via their profile page.

Their email is editable from their profile, and if their email is changed they are automatically logged out of the system to login as their new email.

### DIFF
--- a/frontend/src/modules/user/views/userProfileView.js
+++ b/frontend/src/modules/user/views/userProfileView.js
@@ -105,12 +105,14 @@ define(function(require){
 
     saveUser: function() {
       var self = this;
+      var prev_email = self.model.get('email');
 
       this.$('.error-text').addClass('display-none');
 
       var toChange = {
         'firstName': self.$('#firstName').val().trim(),
-        'lastName': self.$('#lastName').val().trim()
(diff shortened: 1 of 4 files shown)


frontend/src/modules/user/views/userProfileView.js
```diff
@@ -134,7 +136,12 @@ define(function(require){
           });
         },
         success: function(model, response, options) {
-          Backbone.history.history.back();
+          if (prev_email !== self.$('#email').val().trim()) {
```
tomgreenfield: Suggest `model.get('email')` here instead of another form lookup?

frontend/src/modules/user/views/userProfileView.js
```diff
@@ -105,12 +105,14 @@ define(function(require){
 
     saveUser: function() {
       var self = this;
+      var prev_email = self.model.get('email');
```
lc-thomasberger: camelCase please `prevEmail`

canstudios-nicolaw:
Replaced by https://github.com/adaptlearning/adapt_authoring/pull/2208

Issue #1527, opened 2017-02-17
Custom CSS box should be resizeable


hmaffin:
On project Settings it would be a nice touch to have the Custom CSS/LESS input box growable, as sometimes I have a fair bit in that and I can't see it all at once, so have to copy it out into notepad, work with it then paste it all back in.

PR #2031, opened 2018-07-30
Added autoresize feature into ACE editor

simondate:
## Proposed changes

https://github.com/adaptlearning/adapt_authoring/issues/1527

Currently editing CSS/LESS directly in the AT is quite a bad experience as you are limited to a 200px high editor. This change enables the autoresize feature that makes the editor grow as more code it added. https://ace.c9.io/demo/autoresize.html

I set `minLines: 14` as this is the same as 200px. `maxLines: 45` is 700px, this seems to be a reasonable height for someone who wants to make changes here.


taylortom:
Will be fixed in v0.7.0: https://github.com/adaptlearning/adapt_authoring/pull/2181

PR #2036, opened 2018-08-02
match menu-item styling

lc-thomasberger:
update menu-items to match styling of list-item styling when dragging


taylortom:
Cool, that looks really good, thanks!

Issue #880, opened 2015-10-01
Verify installation pre-requisites


taylortom:
I've had a few occasions where my database connection has fallen over unbeknownst to me, causing the install to hang when it tries to initialise the db. Depending on hardware and internet connection, this can be several minutes into the install.

Would be good to have some kind of check for this very early on (before clone/npm install of the framework preferably).

brian-learningpool:
I agree this would be good to have.  Perhaps we should add a function that tests all the pre-requisites?

brian-learningpool:
Actually, we definitely should have a function that verifies all the pre-requisites.

brian-learningpool:
Pre-requisites:
* Node.js version
* MongoDB is installed and can be connected to
* Dependencies are installed (Grunt, etc.)
* GitHub is accessibile

lc-thomasberger:
node version check is related to https://github.com/adaptlearning/adapt_authoring/pull/2038

lc-thomasberger:
With the current way how we handle configuration settings and do the setup we are only able to check the following prerequisites at the start of the installation (with no detailed config settings available):
- installed node version ( #2038 )
- git installed
- grunt-cli installed
- connection to github-api possible

As users can specify a custom connection-url for mongo-db, a check if mongodb is installed locally does not make sense.
Also during installation a custom git-url for the repositories can be specified. Previously we only checked for the github-api being available.

Do you guys think the check's add value and improve the installation experience? 

To check if **git** and **grunt-cli** are installed we might also update the documentation and provide a snippet to check if both are properly installed.
The same would work for validation the node version before installing the tool or upgrading the tool.

taylortom:
Another to check: ffmpeg

taylortom:
Yep, but we can check if it's turned on in the config (agree it should be a warning though, as the app can run just fine without it)

PR #2038, opened 2018-08-07
Add Node.js version on start-up

### LINKED ISSUE #1917: Check for compatible node version on start
## Affected Area
Backend node app. 

## Requested Feature
The app should report on start if an unsupported version of node is installed.
Could use: https://nodejs.org/api/process.html#process_process_release

taylortom:
Fixes #1917.

### DIFF
--- a/install.js
+++ b/install.js
@@ -25,245 +25,250 @@ var superUser = false;
 // from user input
 var configResults;
 
-// we need the framework version for the config items, so let's go
-installHelpers.getLatestFrameworkVersion(function(error, latestFrameworkTag) {
+installHelpers.checkNodeVersion(function(error) {
   if(error) {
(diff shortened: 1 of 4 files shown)


tomgreenfield:
If the check is going to be this stringent, please can you change package.json's `engines.node` to `">=8"` for future-proofing? I have the authoring tool working swimmingly in Node v10.7.0.

lc-thomasberger:
Suggest we replace `installHelpers.checkNodeVersion` with `installHelpers.checkpreRequisites`. In this function we can use `asyc.every` to check the prerequisites defined in #880 . Each check would be a function defined in installHelpers.

taylortom:
TODO: Add warning for versions higher than pkg

taylortom:
Have merged in @lc-thomasberger's changes, but these need to be rebased on the 0.5.1 branch and not develop.

PR #2046, opened 2018-08-14
check prerequisites 1 / 2

### LINKED ISSUE #880: Verify installation pre-requisites

lc-thomasberger:
fixes #880 

Checks known prerequisites in an early stage and exits if not met:
- grunt-cli installed 
- git installed 
- github-api not accessible
Warns if not met:
- local mongodb not found

Part 2 will be to check if a connection to database can be established once it is configured. This will be based on https://github.com/adaptlearning/adapt_authoring/pull/2015

### DIFF
--- a/install.js
+++ b/install.js
@@ -26,7 +26,10 @@ var superUser = false;
 var configResults;
 
 installHelpers.checkPrerequisites(function(error) {
-  exit(0);
+  if(error) {
+    return handleError(error, 1, 'Not all required pre-requisites fulfilled. Please check install guides.');
+  }
+
   installHelpers.getLatestFrameworkVersion(function(error, latestFrameworkTag) {
     if(error) {
       return handleError(error, 1, 'Failed to get the latest framework version. Check package.json.');
--- a/install.js
+++ b/install.js
@@ -25,247 +25,251 @@ var superUser = false;
 // from user input
 var configResults;
 
-// we need the framework version for the config items, so let's go
-installHelpers.getLatestFrameworkVersion(function(error, latestFrameworkTag) {
-  if(error) {
-    return handleError(error, 1, 'Failed to get the latest framework version. Check package.json.');
-  }
-  inputData = {
-    useConfigJSON: {
-      name: 'useJSON',
-      description: 'Use existing config values? y/N',
-      type: 'string',
-      before: installHelpers.inputHelpers.toBoolean,
-      default: 'N'
-    },
-    startInstall: {
-      name: 'install',
-      description: 'Continue? Y/n',
-      type: 'string',
-      before: installHelpers.inputHelpers.toBoolean,
-      default: 'Y'
-    },
-    server: [
-      {
-        name: 'serverPort',
-        type: 'number',
-        description: 'Server port',
-        pattern: installHelpers.inputHelpers.numberValidator,
-        default: 5000
-      },
-      {
-        name: 'serverName',
+installHelpers.checkPrerequisites(function(error) {
+  exit(0);
+  installHelpers.getLatestFrameworkVersion(function(error, latestFrameworkTag) {
+    if(error) {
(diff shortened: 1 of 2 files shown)


lib/installHelpers.js
```diff
@@ -578,3 +579,71 @@ function error(msg) {
...
+  const url = 'https://api.github.com/';
+  request({
+    headers: {
+      'User-Agent': DEFAULT_USER_AGENT
+    },
+    uri: url,
+    method: 'GET'
+  }, function(error, response) {
+    if (error) {
+      result.errors.push(`Failed to´connect to ${url}`);
```
tomgreenfield: Random character `´` in error string.

install.js
```diff
@@ -25,248 +25,255 @@ var superUser = false;
 // from user input
 var configResults;
 
-// we need the framework version for the config items, so let's go
-installHelpers.getLatestFrameworkVersion(function(error, latestFrameworkTag) {
+installHelpers.checkPrerequisites(function(error) {
   if(error) {
-    return handleError(error, 1, 'Failed to get the latest framework version. Check package.json.');
+    return handleError(error, 1, 'Not all required pre-requisites fulfilled. Please check install guides.');
```
tomgreenfield: Suggest removing `required`. A prerequisite is required by nature.

lib/installHelpers.js
```diff
@@ -578,3 +579,71 @@ function error(msg) {
 function isSilent() {
   return process.env.SILENT;
 }
+
+/**
+* Manually collect all misisng prerequisites
```
tomgreenfield: Typo: `misisng`.

lib/installHelpers.js
```diff
@@ -578,3 +579,71 @@ function error(msg) {
...
+
+  async.applyEach([
+    checkGitInstalled,
+    checkGruntCliInstalled,
+    checkGithubConnection,
+    checkLocalMongoDbConnection
+  ], result, function(error) {
+    if (error) return callback(error);
+    if (result.errors.length) {
+      return callback(new Error(result.errors.join(', ')));
```
tomgreenfield: Suggest a line break between errors.

lc-thomasberger: have tried this, will format it oddly. I can do however if you like.

tomgreenfield: This is the output for me in Windows if changed to `.join('\n')`:

taylortom: Looks good.

Agree that it would look nicer with each error on a separate line.
```
Missing the following pre-requisites, please check install guides:
- git
- grunt
- ffmpeg
- failed to connect to github.com
```

lib/installHelpers.js
```diff
@@ -578,3 +580,89 @@ function error(msg) {
...
+    method: 'GET'
+  }, function(error, response) {
+    if (error) {
+      result.push(`- failed to connect to ${url}`);
+    }
+    githubCb();
+  });
+}
+
+function checkLocalMongoDbConnection(result, callback) {
```
tomgreenfield: This doesn't actually check the connection does it? Maybe needs a rename to avoid any confusion.

lib/installHelpers.js
```diff
@@ -578,3 +580,89 @@ function error(msg) {
...
+  ],
+  'Missing the following pre-requisites, please check install guides:',
+  callback);
+}
+function checkOptionalDependencies(callback) {
+  checkDependencies([
+    checkLocalMongoDbConnection,
+    checkFfmpegInstalled
+  ],
+  'Missing the following optional dependecies, please check install guides:',
```
tomgreenfield: Typo: `dependecies`.

PR #2047, opened 2018-08-14
log DB-connection error

### LINKED ISSUE #880: Verify installation pre-requisites

lc-thomasberger:
fixes #880 

Unfortunately we can only use `lib/database` once the app is started (preload the configuration module). Therefore we can't check potential connection issues earlier in the install script.

This PR emit's an error event when the app can't be started. This is mostly related to not being able to connect to the DB. The install script displays this error message to the user. 
This makes the overall installation experience a bit better, as it gives useful information about what caused the error.

### DIFF
--- a/install.js
+++ b/install.js
@@ -417,6 +417,9 @@ function configureMasterTenant(callback) {
   installHelpers.showSpinner('Starting server');
   // run the app
   app.run({ skipVersionCheck: true });
+  app.on('serverStartError', function(error) {
+    return exit(1, error.message || error);
+  });
   app.on('serverStarted', function() {
     installHelpers.hideSpinner();
     database.checkConnection(function(error) {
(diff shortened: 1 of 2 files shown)


install.js
```diff
@@ -417,6 +417,9 @@ function configureMasterTenant(callback) {
   installHelpers.showSpinner('Starting server');
   // run the app
   app.run({ skipVersionCheck: true });
+  app.on('serverStartError', function(error) {
+    return exit(1, error.message || error);
+  });
```
taylortom: It looks like this just replicates the error handling in `lib/application.js`

lc-thomasberger: It also logs the error message that connection failed.
Without that I don't get this message.

lc-thomasberger:
Is this still needed in regards to https://github.com/adaptlearning/adapt_authoring/pull/2059

PR #2049, opened 2018-08-14
auto-generate session secret

### LINKED ISSUE #2007: Session secret should be auto-generated
# Feature 

## Affected Area
https://github.com/adaptlearning/adapt_authoring/blob/master/conf/config-sample.json#L5
install script
https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

## Requested Feature
Instead of setting a (weak) default session secret, we should auto generate a secure random default session secret string when the install script is executed.
A user defined session secret string may be passed as cli argument or configuration option.

## Current Workaround
Users must type a secure sessions secret string whenever the install script is executed.

## Additional Information
Should we always use a auto generated session secret? 
Is it only used to encode sessions?

lc-thomasberger:
Session-secret is autogenereated by default. If a session-secret is defined in config.json, this is used.
Also removes the prompt to enter the session-secret.

fixes #2007

### DIFF
--- a/install.js
+++ b/install.js
@@ -109,13 +109,6 @@ installHelpers.getLatestFrameworkVersion(function(error, latestFrameworkTag) {
         default: 'data'
       },
       {
-        name: 'sessionSecret',
-        type: 'string',
-        description: 'Session secret (value used when saving session cookie data)',
-        pattern: /^.+$/,
-        default: 'your-session-secret'
-      },
-      {
         name: 'authoringToolRepository',
         type: 'string',
         description: "Git repository URL to be used for the authoring tool source code",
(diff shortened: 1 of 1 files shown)


taylortom:
I think we should enable this by default, unless a value has been set in the config.json (can probably remove the input from install.js and leave that as an advanced setting)

install.js
```diff
@@ -273,6 +267,10 @@ function generatePromptOverrides() {
     var configData = JSON.parse(JSON.stringify(configJson).replace(/true/g, '"y"').replace(/false/g, '"n"'));
     configData.install = 'y';
   }
+
+  const isSessionSecretDefined = USE_CONFIG && configData.sessionSecret;
+  const sessionSecret = (isSessionSecretDefined) ? configData.sessionSecret : crypto.randomBytes(64).toString('hex');
```
tomgreenfield: `isSessionSecretDefined` will end up being the secret itself if defined, so suggest just the one var:
```js
const sessionSecret = USE_CONFIG && configData.sessionSecret ||
  crypto.randomBytes(64).toString('hex');
```

taylortom:
worksforme

PR #681, opened 2015-05-21
Fixes around asset selection

brian-learningpool:
Resolves:
- 'Remove Asset' button with old courses and bad data
- The issue with 'tags' being pushed up

PR #2051, opened 2018-08-20
Delete unnecessary removal of tags attributes, Fixes #2050

### LINKED ISSUE #2050: Course tags are removed when selecting an asset (scaffoldAssetView) in project settings extensions modal window
## Expected Behaviour
Adding assets or external assets should not delete other fields.

## Actual Behaviour
Adding an asset to an extension through scaffoldAssetView removes any tags associated with the course.

## Steps to Reproduce
1. add tags to a course and save
2. add an extension that the user can select an asset or external asset from project setting (e.g resources)
3. add a resource that uses an asset e.g. add a link and input and save the "_link" value for the external asset.
4. click save on project settings
5. tags are removed from the course

Probably cause by [this](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/modules/scaffold/views/scaffoldAssetView.js#L190) line. Not sure why you would want to remove tags from the model at this point?

## Versions
  - Framework Version: n/a
  - Operating System: n/a
  - Browser: n/a

dancgray:


tomgreenfield:
@brian-learningpool, would you mind casting your eye over this one? Just tagging you because of https://github.com/adaptlearning/adapt_authoring/pull/681.

Issue #2053, opened 2018-08-24
Deprecation warning when using MongoDB versions 3.2 or above 


dancgray:
This issue is for reference. [Waiting](https://github.com/Automattic/mongoose/issues/6880#issuecomment-414151421) for mongoose update or docs.

## Actual Behavior
The following warnings are given when installing or running the develop branch of the authoring tool with mongoDB versions 3.2 or above. (Tested in MongoDB v3.6, Node v8.11.2)

(node:3144) DeprecationWarning: collection.find option [fields] is deprecated and will be removed in a later version.
(node:3124) DeprecationWarning: collection.ensureIndex is deprecated. Use createIndexes instead.

## Versions
  - Authoring Tool Version: develop branch
  - Framework Version: n/a
  - Operating System: n/a
  - Browser:

## Additional Information

See Mongoose issue [#6880](https://github.com/Automattic/mongoose/issues/6880).

Issue #2028, opened 2018-07-26
Update Mongo URL parser


tomgreenfield:
Upon `node server`:
> DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect.

Probably should sort this out at some point to avoid worrying novice users.

dancgray:
Looks like a result of an [update](https://stackoverflow.com/questions/50448272/avoid-current-url-string-parser-is-deprecated-warning-by-setting-usenewurlpars#) to the mongodb driver.

We can either use an older mongoose version or add the following [here](https://github.com/adaptlearning/adapt_authoring/blob/develop/lib/dml/mongoose/index.js#L70):
`options.useNewUrlParser = true;`

Have tried with useNewUrlParser successfully (MongoDB 3.6).

benoit-girard:
I am a novice user, and I had precisely this error message upon `node server`.
Is this really fixed?
I am using node v8.12.0

taylortom:
@benoit-girard this fix will be available in the next release (v0.6.1)

PR #2062, opened 2018-09-04
Pin mongoose to v5.1.* to stop deprication warnings

dancgray:
See issues #2028 and #2053 . [This post](https://stackoverflow.com/questions/50448272/avoid-current-url-string-parser-is-deprecated-warning-by-setting-usenewurlpars#) gives a good explanation.

I think pinning mongoose is a safer option than implementing the changes to mongoose connection options unless we need any fixes in mongoose v5.2.* or support for mongoDB v4.0. 

We should specify which mongoDB versions we support in the docs as the mongoose driver we have specified only supports MongoDB >=3.*. See [version compatibility](https://mongoosejs.com/docs/compatibility.html).

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "migrate-mongoose": "^3.2.2",
     "mime": "^2.3.1",
     "moment": "^2.22.1",
-    "mongoose": "^5.1.1",
+    "mongoose": "~5.1.1",
     "morgan": "^1.9.0",
     "multer": "^1.3.0",
     "needle": "^2.2.1",


taylortom:
Happy for us to support the current stable major version (i.e. currently 4).

tomgreenfield:
Note: `mongoose@5.2.13` has since been released which alleviates most of the issues.

All our production code requires is this [patch](https://github.com/adaptlearning/adapt_authoring/files/2356466/mongoose.patch.txt).

The only outstanding warning arises from our use of [connect-mongo](https://github.com/jdesboeufs/connect-mongo/issues/297):
```shellsession
DeprecationWarning: collection.update is deprecated. Use updateOne, updateMany, or bulkWrite instead.
```

taylortom:
Thanks Tom. 

Let's approve/merge the linked fixes, and revisit this prior to the release of 0.5.1. `connect-mongo` doesn't look to be updated very regularly, so we may need to check alternative modules and use one of those.

wootwoot1234:
@taylortom did you find a good alternative?  I'm in the same boat.

taylortom:
Superseded by https://github.com/adaptlearning/adapt_authoring/pull/2112

PR #2073, opened 2018-09-10
Expose CKEditor's enterMode

### LINKED ISSUE #1898: Nested `<p>` tags in JSON output
# BUG Template

## Expected Behaviour
Line breaks should be represented by `<br>` rather than a wrapping `<p>`

## Actual Behaviour
All string attributes are wrapped with `<p>` tag.
Most JSON string attributes are now also handlebars templates. 
Nested templates create nested `<p>` tags. [see here](https://ckeditor.com/old/forums/Support/Nested-paragraph-tags.)

## Steps to Reproduce
1. Edit some text
1. Download the sco
1. Run the course
1. Issue with any component using nested attributes (see assessment results, feedback and completion body)

## Versions
  - Authoring Tool Version: Current
  - Framework Version: Current
  - Operating System: All
  - Browser: All

tomgreenfield:
Adds a configuration option to resolve #1898.

Have left the default as wrapping in `<p>` tags for legacy purposes and the fact it's not [universally recommended](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_enterkey.html).

### DIFF
--- a/frontend/src/modules/scaffold/backboneFormsOverrides.js
+++ b/frontend/src/modules/scaffold/backboneFormsOverrides.js
@@ -70,6 +70,7 @@ define([
       this.editor = CKEDITOR.replace(this.$el[0], {
         dataIndentationChars: '',
         disableNativeSpellChecker: false,
+        enterMode: CKEDITOR[Origin.constants.ckEditorEnterMode],
         entities: false,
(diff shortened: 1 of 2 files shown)


lc-thomasberger:
I have update the wiki with the new settings https://github.com/adaptlearning/adapt_authoring/wiki/Configuring-the-Authoring-Tool

Also updated SMTP section as it was out of date.

PR #2074, opened 2018-09-12
Refactor mailer and add connection test

### LINKED ISSUE #2072: Mailer should verify settings on startup
For reference, from http://nodemailer.com/smtp

> You can verify your SMTP configuration with verify(callback) call (also works as a Promise). If it returns an error, then something is not correct, otherwise the server is ready to accept messages.

```javascript
// verify connection configuration
transporter.verify(function(error, success) {
   if (error) {
        console.log(error);
   } else {
        console.log('Server is ready to take our messages');
   }
});
```

taylortom:
Fixes #2072.

**TODO**: needs to be integrated with the new dependency checks once they're in.


lc-thomasberger:
Will this allow us to send different templates with the mailer?

taylortom:
That _should_ be possible already (but I've added a few related fixes to this).

lib/mailer.js
```diff
@@ -6,130 +6,166 @@ const helpers = require('./helpers');
...
-const configure = function (options, callback) {
-  // TODO localise these errors
-  if (!helpers.isValidEmail(options.to)) {
-    return callback(new Error("Can't send email, '" + options.to + "' is not a valid email address"));
+/**
+* Sends an email
+* @param {String} toAddress Address to send mail to
+* @param {String} subject Email subject line
+* @param {String} text Email body
+* @param {Object} templateData
```
tomgreenfield: Optional?

PR #2076, opened 2018-09-12
Fix LocalAuth.prototype.resetPassword response

### LINKED ISSUE #2075: Password reset displays error alert on success
The original calling code does a `Backbone.Model.save`, which causes the underlying `$.ajax` call to expect a JSON response, but we just do `res.status(200).end()` on the server side. Changing the response code to `res.status(200).json({})` fixes this (could also return the user).

See [plugins/auth/local/index.js#L199-L222](https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/auth/local/index.js#L199-L222).

taylortom:
Fixes #2075.

### DIFF
--- a/frontend/src/modules/user/views/resetPasswordView.js
+++ b/frontend/src/modules/user/views/resetPasswordView.js
@@ -61,12 +61,9 @@ define(function(require) {
       this.model.save(toChange, {
         patch: true,
         success: _.bind(function(model, response, options) {
-          if (response.success) {
-            this.$('.form-reset-password').addClass('display-none');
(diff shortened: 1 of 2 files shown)


tomgreenfield:
After entering the new password on the page from the link in the email, clicking the 'Change password' button does nothing for me.

lc-thomasberger:
`response.success` is `undefined`

https://github.com/adaptlearning/adapt_authoring/blob/develop/frontend/src/modules/user/views/resetPasswordView.js#L64

taylortom:
@tomgreenfield @lc-thomasberger have removed that if block

PR #2083, opened 2018-09-24
build course only once validated correctly

### LINKED ISSUE #2067: No useful error when attempting to preview course with empty children
On `develop` branch, attempt to preview a course with no content objects, or a block with no components etc.

### Erroring function:
https://github.com/adaptlearning/adapt_authoring/blob/b86cac9c0f2e0fd942ab81ee2a1122dd3ab361b2/lib/outputmanager.js#L199-L272

### Backend output:
```shellsession
info: [06 Sep 2018 12:25:38 +01:00] Retrieving JSON
error: [06 Sep 2018 12:27:27 +01:00] TypeError: Cannot read property 'length' of undefined
```

lc-thomasberger:
fixes #2067 

**Note:**
Error is passed as second argument [here](https://github.com/adaptlearning/adapt_authoring/blob/release/0.5.1/frontend/src/modules/editor/global/views/editorView.js#L80). That is why the cb does not follow node's error first convention.

### DIFF
--- a/frontend/src/modules/editor/global/views/editorView.js
+++ b/frontend/src/modules/editor/global/views/editorView.js
@@ -45,11 +45,11 @@ define(function(require) {
         'editorView:copyID': this.copyIdToClipboard,
         'editorView:paste': this.pasteFromClipboard,
         'editorCommon:download': function(event) {
-          this.validateProject(event, this.downloadProject);
+          this.validateProject(this.downloadProject);
         },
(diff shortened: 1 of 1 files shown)


taylortom:
Thanks for looking at this :) Would be happy for the event param to be removed, as it doesn't seem to be used anywhere anyway

lc-thomasberger:
I've rebased this PR on the release branch. This removed your review.

PR #2094, opened 2018-10-04
Stop export excluding course.model.schema

### LINKED ISSUE #1512: exported course can't be used for translation with manual modifications

dancgray:
Fixes #1512. Currently one of the core framework schemas is excluded from export source due to [this](https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/output/adapt/index.js#L316) line.  This prevents the grunt translation task on the exported framework.

## Proposed changes
Re-add the negative look ahead that [previously](https://github.com/adaptlearning/adapt_authoring/pull/1601/files#diff-df68ee67d4821734092e7178fd97957aR309) fixed this. 

Do we need to add this for node_module and courses as well?

### DIFF
--- a/plugins/output/adapt/index.js
+++ b/plugins/output/adapt/index.js
@@ -313,7 +313,7 @@ function copyFrameworkFiles(results, filesCopied) {
       includes[i] = '\/' + includes[i] + '(\/|$)';
 
     var includesRE = new RegExp(includes.join('|'));
-    var excludesRE = new RegExp(/\.git\b|\.DS_Store|\/node_modules|\/courses\b|\/course\b|\/exports\b/);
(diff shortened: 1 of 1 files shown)


taylortom:
I guess it depends on how likely we are to have anything named similarly in the future. _Could_ happen with courses, very unlikely with node_modules 😄 

Either way, happy to +1 as is.

PR #2100, opened 2018-10-08
maxFileUploadSize config option

### LINKED ISSUE #2095: Formidable limits file uploads to 200mb
This isn't something that's caused a problem in the past, but may become an issue going forward, now we have course import. Rather than just upping this arbitrarily, we could surface this in config.json (or both...). 

Docs here: https://github.com/felixge/node-formidable

We'd need to add something like the below to any code that creates a new IncomingForm (e.g. [assetmanager.js#L386](https://github.com/adaptlearning/adapt_authoring/blob/master/lib/assetmanager.js#L386)).

```javascript
var form = new IncomingForm();
form.maxFileSize = 200 * 1024 * 1024;
```

lc-thomasberger:
fixes #2095 

Exposes `maxFileUploadSize` to config.json. Defaults to 200mb. Uses [bytes](https://github.com/visionmedia/bytes.js#bytesparsestringnumber-value-numbernull) to parses a human redable string e.g. `200kb|50mb|1gb`.

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -79,7 +79,8 @@ function Configuration() {
   this.conf = {
     root: this.serverRoot,
     maxLoginAttempts: 3,
-    ckEditorExtraAllowedContent: 'span(*)'
+    ckEditorExtraAllowedContent: 'span(*)',
+    maxFileUploadSize: '200mb'
   };
   this.mconf = {};
   return this;
@@ -150,7 +151,7 @@ Configuration.prototype.setConfig = function (name, value) {
  */
 Configuration.prototype.getConfig = function (name) {
   if (name === 'maxFileUploadSize') {
-    return bytes.parse(this.conf.maxFileUploadSize) || (200 * 1024 * 1024);
+    return bytes.parse(this.conf.maxFileUploadSize);
   }
 
   if (name && typeof name === 'string') {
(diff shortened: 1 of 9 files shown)


taylortom:
Cool, thanks :)

A couple of questions on standards/consistency: 
- Do we need to bother with client-side validation here, rather than just returning an error from the api/server, as it adds quite a lot of extra code (more stuff to maintain).
- Should we abstract the form stuff into a helper, to remove the repetition of setting `maxFileUploadSize` (may not be worth it).

lc-thomasberger:
I added the client-side validation because the Error feedback is not consistent on the different modules (Plugin upload, Course import, Asset upload). For instance, only course import shows the error that describes that the file size is too large. Plugin upload only throws a generic error. Implementing the client side checks was rather easy and it improves the UX at least a bit.

I thought about writing the server side code in a reusable manner, but I think this would have required changing more on the existing code. For the server refactor I think we should abstract the form / upload handling in a reusable component. This will also make sure that error handling will become consistent and easier to maintain.

lc-thomasberger:
Reverted the clientside validation.

Created a hbs helper `{{maxUploadSize}}` to show the max allowed size from config.

taylortom:
Can we do something with the error handling? I was thinking something like this: [errorhandling.txt](https://github.com/adaptlearning/adapt_authoring/files/2561338/errorhandling.txt)

lc-thomasberger:
@taylortom are you happy with the changes?

lib/configuration.js
```diff
@@ -78,8 +80,9 @@ function Configuration() {
   this.conf = {
     root: this.serverRoot,
     maxLoginAttempts: 3,
-    ckEditorEnterMode: 'ENTER_P',
-    ckEditorExtraAllowedContent: 'span(*)'
+    ckEditorExtraAllowedContent: 'span(*)',
+    maxFileUploadSize: '200mb',
```
tomgreenfield: Please uppercase the default: `200MB`.

PR #2109, opened 2018-10-23
Include Custom LESS in import

### LINKED ISSUE #2108: Custom LESS should be included on import

taylortom:
Includes a refactor of content type iteration to make this happen (course & config are now included in contentMap).

Fixes #2108

### DIFF
--- a/plugins/output/adapt/importsource.js
+++ b/plugins/output/adapt/importsource.js
@@ -20,12 +20,15 @@ function ImportSource(req, done) {
   var dbInstance;
 
   var contentMap = {
+    'course': 'course',
+    'config': 'config',
     'contentobject': 'contentObjects',
     'article': 'articles',
     'block': 'blocks',
     'component': 'components'
   };
 
+  var cachedJson = {};
   var plugindata = {
     pluginTypes: [
       { type: 'component', folder: 'components' },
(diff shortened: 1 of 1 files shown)


plugins/output/adapt/importsource.js
```diff
@@ -462,114 +452,132 @@ function ImportSource(req, done) {
...
-            } else {
-              detachedElementsMap[originalData._id] = type;
-              logger.log('warn', 'Cannot update ' + originalData._id + '._parentId, ' +  originalData._parentId + ' not found in idMap');
-              return cb();
-            }
-          }
-          // define the custom properties and and pluginLocations
-          var genericPropKeys = Object.keys(dbInstance.getModel(type).schema.paths);
-          var customProps = _.pick(data, _.difference(Object.keys(data), genericPropKeys));
+      if(_.isEmpty(customProps)) resolve(data);
```
tomgreenfield: Just checking you intended to remove the `return` here?

taylortom: Probably should still return to be safe.
```suggestion
      if(_.isEmpty(customProps)) return resolve(data);
```

plugins/output/adapt/importsource.js
```diff
@@ -462,114 +452,132 @@ function ImportSource(req, done) {
...
             logger.log('warn', 'Failed to import ' + type + ' ' + (originalData._id || '') + ' ' + error);
             return done();
-          }
-          // Create a courseAssets record if needed
-          createCourseAssets(type, record, function(error){
-            if(error) return done(error, record);
-            done(null, record);
-          });
+          } // Create a courseAssets record if needed
+          createCourseAssets(type, record, () => done(null, record));
```
lc-thomasberger: cb will never return an error. Is this intended? 
```
createCourseAssets(type, record, error => done(error, record));
```

taylortom: Yeah, the original code didn't do this -- will gladly add this if needed. 
@dancgray do you have any opinions?

dancgray: @taylortom I think it should return the error.

taylortom:
```suggestion
          createCourseAssets(type, record, done);
```

taylortom: Just seen the diff with the original, I'm clearly misremembering things!

PR #2112, opened 2018-10-25
 Replace connect-mongo with connect-mongodb-session

taylortom:
Have replaced the offending node module, which solves the mongo warnings. However, due to a change in the way the modules store the session data, we're now getting errors at various other points. 

**Detail**:
We use resource URIs as keys for storing cached user permissions, which mongo doesn't like. We don't get these issues with the old package as it stringifys everything. The new package stores as a regular object.

Old data:
```
"{\"cookie\":{\"originalMaxAge\":false,\"expires\":false,\"httpOnly\":true,\"path\":\"/\"},\"passport\":{\"user\":{\"_id\":\"5b4e06ca37804c9aec64adcc\",\"email\":\"admin\",\"tenant\":{\"_id\":\"5b4e063b37804c9aec64adaf\",\"name\":\"master\",\"isMaster\":true}}},\"cachedPermissions\":{\"urn:x-adapt:5b4e063b37804c9aec64adaf/api/role\":true}}"
```

New data: 
```
       "cookie" : {
            "originalMaxAge" : false,
            "expires" : false,
            "secure" : null,
            "httpOnly" : true,
            "domain" : null,
            "path" : "/",
            "sameSite" : null
        },
        "passport" : {
            "user" : {
                "_id" : ObjectId("5b4e06ca37804c9aec64adcc"),
                "email" : "admin",
                "tenant" : {
                    "_id" : ObjectId("5b4e063b37804c9aec64adaf"),
                    "name" : "master",
                    "isMaster" : true
                }
            }
        },
        "cachedPermissions" : {
            "urn:x-adapt:5b4e063b37804c9aec64adaf/api/tenant" : true,
            "urn:x-adapt:5b4e063b37804c9aec64adaf/api/role" : true,
            "urn:x-adapt:5b4e063b37804c9aec64adaf/api/upload/contentplugin" : true,
            "urn:x-adapt:5b4e063b37804c9aec64adaf/api/content/course/*" : true,
            "urn:x-adapt:5b4e063b37804c9aec64adaf/api/content/course/5bd185fe0540a600a68b676f" : true
        }
```


taylortom:
Field names with dots are allowed in MongoDB version 3.6 and above. See [Restrictions on Field Names](https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names) for more info.

taylortom:
Have updated the PR to store cached permissions in two arrays as below:
```
      "cachedPermissions" : {
            "allowed" : [ 
                "urn:x-adapt:5b4e063b37804c9aec64adaf/api/asset/undefined", 
                "urn:x-adapt:5b4e063b37804c9aec64adaf/api/shared/course?search%5Btitle%5D=.*.*&operators%5Bskip%5D=0&operators%5Blimit%5D=1&operators%5Bsort%5D%5BupdatedAt%5D=-1", 
                "urn:x-adapt:5b4e063b37804c9aec64adaf/api/shared/course?search%5Btitle%5D=.*.*&operators%5Bskip%5D=0&operators%5Blimit%5D=8&operators%5Bsort%5D%5BupdatedAt%5D=-1"
            ],
            "disallowed" : []
        }
```
This is a lot less performant than accessing object values. I'm also in two minds about how useful cached permissions even are - we also seem to be storing routes with query strings (see above example), which just seems wrong.

How does everyone feel about removing cached permissions altogether?

lc-thomasberger:
When I test this with node v 10.11.0 I get several errors related to ObjectId's being passed into path functions.
Errors occur during: preview, export, editor save, course import ...

lc-thomasberger:
Do we need a migration for this to cleanup existing sessions in the DB?

If you try to login with removing existing documents the Server crashes due to an Error:
TypeError: Cannot read property 'expires' of undefined

taylortom:
@adaptlearning/authoring-tool-push have added the migration as discussed on yesterday's call. Please have a test and let me know what happens (back-up DB/sessions collection first).

taylortom:
@adaptlearning/authoring-tool-push should have fixed the install/upgrade issues now, both scripts working for me.

lib/permissions.js
```diff
@@ -597,38 +484,26 @@ exports = module.exports = {
...
+      this.hasPermission(user._id, action, resource, function (error, allowed) {
+        if(error) {
+          return res.status(500).json(error);
         }
-
-        if (allowed) {
-          // if we have permission, just continue
-          return next();
+        if(!allowed) {
+          res.status(403).json({ statusCode: STATUS.ACCESS_DENIED });
```
dancgray: Should this return here?

I'm getting the error `Error: Can't set headers after they are sent.`pointing to this line if I try and preview as a course creator.

taylortom:
```suggestion
          return res.status(403).json({ statusCode: STATUS.ACCESS_DENIED });
```

taylortom: Good spot, have added

dancgray: Cool, is it parsing the statements for actions correctly [here](https://github.com/adaptlearning/adapt_authoring/blob/issue/2028/lib/permissions.js#L139)? When I login as a course creator and try and preview my own course it cannot find read action in any statements and an error is returned.

package.json
```diff
@@ -47,7 +47,7 @@
     "bower": "^1.8.4",
     "chalk": "^2.4.1",
     "compression": "^1.7.2",
-    "connect-mongo": "^2.0.1",
+    "connect-mongodb-session": "2.0.3",
```
tomgreenfield:
```suggestion
    "connect-mongodb-session": "^2.0.5",
```
Else we're missing out on what look to be useful [releases](https://github.com/mongodb-js/connect-mongodb-session/blob/master/History.md).

lib/permissions.js
```diff
@@ -128,61 +128,33 @@ function PolicyStatementParser(action, resource, statements, callback) {
...
-      for (index = 0; index < statements.length; ++index) {
-        el = statements[index];
-        if (!el) { // save us!
-          continue;
-        }
-
-        // check if action matches
-        if (-1 === el.action.indexOf(action)) {
-          // no point in further parsing of the statement
+      for(let i = 0, statement = statements[i]; i < statements.length; statement = statements[++i]) {
```
tomgreenfield: As discussed, might be more readable keeping the `statement` assignment in the body.

lib/contentmanager.js
```diff
@@ -881,7 +880,13 @@ ContentManager.prototype.setupRoutes = function () {
...
       that.retrieve(type, query, options, function (error, results) {
         if (error) {
-          res.statusCode = error instanceof ContentTypeError ? 400 : 500;
+          if(error instanceof ContentTypeError) {
+            res.status(400);
+          } else if(error instanceof ContentPermissionError) {
+            res.status(403);
+          } else {
+            res.status(500);
+          }
```
tomgreenfield: If you detest nested `if`s as much as the rest, would be good to reverse the logic here.
```js
if (!error) {
  // @TODO - we need to filter resources according to permissions!
  res.setHeader('Cache-Control', 'no-cache, no-store, must-revalidate');
  ...
}

if(error instance of ContentTypeError) {
  res.status(400);
} else if(error instanceof ContentPermissionError) {
...
```

taylortom: I prefer to keep error handling at the top, hence the nesting

PR #2119, opened 2018-11-04
Swedish translation of *-application.json

KristofferBroberg:
Might come in handy for some!

I understand that at this point some strings are still hard coded and thus won't be affected by this JSON. This is my frist PR ever so if I've done any mistakes, please do tell!


lc-thomasberger:
Thanks @KristofferBroberg , looks very good to me.

Unfortunately there are 2 keys missing:
`app.maxfileuploadsize`
`app.uploadsizeerror`

taylortom:

PR #2122, opened 2018-11-05
User management filter / search 

### LINKED ISSUE #1509: Add ability to search/filter user list
Thinking specifically:
- Search box to search by email
- Sort by ... (all attributes)
- Maybe also filters for relevant fields (role, tenant?)

lc-thomasberger:
fixes #1509


frontend/src/modules/userManagement/collections/userCollection.js
```diff
@@ -2,13 +2,134 @@
...
+    filterFailedLoginCount: function(model) {
+      if (this.filterGroups.failedLoginCount
+        && this.filterGroups.failedLoginCount.indexOf(model.get('failedLoginCount').toString()) < 0) {
+        model.set('_isHidden', true);
+      }
+    },
+
+    filterTenants: function() {
+      if (this.filterGroups.tenantName
+        && this.filterGroups.tenantName.indexOf(model.get('tenantName')) < 0) {
```
dancgray: Getting the error `Uncaught ReferenceError: model is not defined  userCollection.js:78` when filtering or sorting by tenant.

dancgray:
This will be really useful with multi tenancy, but the tenant toggle might be better as a drop down or search. The screen will really fill up with lots of tenants:
screenshot 2018-12-10 at 16 07 59

lc-thomasberger:
@dancgray have update the PR. Tenants can now be filtered using selectize dropdowns (used also for tags).

frontend/src/modules/userManagement/less/userManagement.less
```diff
@@ -1,13 +1,106 @@
-@import 'sharedStyles';
-@import 'colours';
+// @import 'sharedStyles';
+// @import 'colours';
```
canstudios-nicolaw: Could these be removed rather than commented out?

taylortom:
Thanks for this 😄 

As a general point, I think we could do with making the UI a bit more consistent with the other searches/filters. The dashboard, asset management and content all have this UI in the sidebar, so think we need to either bring this in line with the others, or go with a new UI/UX direction for all. Not saying the current solution is the best...

For the dates and failed login count, I'm not sure how much these will be used -- sorts may be enough? In the case of logins, locked/not locked could be more useful?

With the sorting, could we make the whole table headers clickable, rather than just the arrow icons? It might also be useful to have some visual indicator showing which field (if any) is being used to sort. Hiding the icons if not in use may make most sense here?

Happy to go with the general consensus on the above points, so interested to hear any other opinions 😄

routes/lang/en-application.json
```diff
@@ -337,6 +337,7 @@
   "app.reset": "Reset",
   "app.titlelock": "Link with title",
   "app.titleunlock": "Unlink from title",
+  "app.searchByMail": "Search by E-Mail",
```
tomgreenfield: `E-mail`  → `email`.

frontend/src/modules/userManagement/views/filterView.js
```diff
@@ -0,0 +1,81 @@
...
+    onSearchInput: function(event) {
+      var searchTerm = $(event.currentTarget).val();
+      this.collection.mailSearchTerm = searchTerm.toLowerCase();
+      this.collection.sortCollection();
+    },
+
+    onFormChange: function() {
+      var attributeMap = {};
+
+      var selected = this.$('input:checked').each(function(index, input) {
```
tomgreenfield: Probably don't need to assign this to a variable.

frontend/src/modules/userManagement/collections/userCollection.js
```diff
@@ -1,15 +1,100 @@
...
+    filter: function() {
+      this.models.forEach(function(model) {
+        this.filterTenants(model);
+        this.filterRoleNames(model);
+      }, this);
+    },
+
+    filterTenants: function(model) {
+      if (this.filterGroups.tenantName
+        && this.filterGroups.tenantName.indexOf(model.get('tenantName')) < 0) {
```
tomgreenfield: Suggest caching:
```js
var tenantName = this.filterGroups.tenantName;

if (tenantName && tenantName.indexOf(model.get('tenantName')) < 0) {
```

frontend/src/modules/userManagement/collections/userCollection.js
```diff
@@ -1,15 +1,100 @@
...
+        model.set('_isHidden', true);
+      }
+    },
+
+    filterRoleNames: function(model) {
+      if (this.filterGroups.roleNames
+        && this.filterGroups.roleNames.indexOf(model.get('roleNames')[0]) < 0) {
```
tomgreenfield: Suggest caching as above.

lc-thomasberger:
Thanks for testing @tomgreenfield .
What's your Screenresolution? The odd behavior is due to the current implementation of the Sidebar. We would need to fix this.
I will reduce the spacing between the tenant toggle switches to reduce the height of the filter screen.

tomgreenfield:
One of my laptops is 1280 x 800 and I find that's a fairly good base resolution to test with.

frontend/src/modules/userManagement/collections/userCollection.js
```diff
@@ -1,15 +1,100 @@
...
+        b = new Date(b || '01.01.1900');
+      }
+
+      if (typeof a === 'string' && typeof b === 'string') {
+        a = a.toLowerCase();
+        b = b.toLowerCase();
+      }
+
+      if (a > b) return this.direction;
+      else if (a < b) return this.direction*-1;
```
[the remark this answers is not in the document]
lc-thomasberger: We return 1, -1 or 0

lc-thomasberger:
Regarding cases, there are no cahnges made to user table.
The filter displays the displayName and uses the name to filter the user collection: https://github.com/adaptlearning/adapt_authoring/blob/issue/1509/frontend/src/modules/userManagement/templates/userManagementFilter.hbs#L14

frontend/src/modules/userManagement/collections/userCollection.js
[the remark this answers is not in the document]
oliverfoster:
```js      
      if (a > b) return this.direction;
      if (a < b) return this.direction*-1;
      return 0;
```

taylortom:
I think the difference between case is due to showing name/displayName, which was my error 😅 
Changing [userModel.js#L52-L56](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/modules/userManagement/models/userModel.js#L52-L56) to use displayName is probably the best way to go.

frontend/src/modules/sidebar/less/sidebar.less
```diff
@@ -93,7 +93,8 @@
 }
 
 .sidebar-inner {
-  padding: 61px 0 25px 0;
+  padding: 0;
+  padding-bottom: 25px;
```
tomgreenfield: Can we please remove `padding-bottom`? Then there's absolutely no scrollbar with a screen resolution of 1280 x 800 👍.

frontend/src/modules/userManagement/less/userManagement.less
```diff
@@ -1,13 +1,133 @@
...
+      }
+
+      .form-container-style;
+      max-width: inherit;
+      .tb-heading {
+        > div button {
+          display: flex;
+          align-items: center;
+          text-align: left;
+          width: 100%;
```
tomgreenfield: Was thinking the sort arrow looks better grouped without the `width: 100%`. Happy to hear other opinions though:

#### Before
before
#### After
after

lc-thomasberger: I agree

frontend/src/modules/sidebar/less/sidebar.less
lc-thomasberger: I've tested with 800 as well and it worked for me. Did you have zoom applied?
Happy to remove the padding though 😄

tomgreenfield: Don't be silly, I wouldn't use something as horrible as browser zoom 😉.

This is my Chrome window maximised on Windows 10 with a 1280 x 800 display:

fullscreen

PR #1584, opened 2017-03-10
Allow 'forced' rebuild of courses

taylortom:
Holding ctrl/cmd changes the preview button to force a full build of the current course.

Fixes #1497.

Issue #1497, opened 2017-02-03
Should be able to force a full rebuild of a course


simondate:
When updating a plugin for an existing course it would be nice to have a force-rebuild button that rebuilds all the CSS + JS. At the moment I have to add a random extension and remove it everytime to make the Authoring Tool rebuild fully.

taylortom:
👍  A _slightly_ better to do this is to just go into the theme picker and click save. I've done something like this in the Kineo fork, will dig it out and submit a PR.

However, I think this is only a temporary solution; the real problem here is that the build process is pretty poor. I think we need to get to a place where preview uses an 'instant build', and the full build is only run on download/export. This being said, I don't think this is possible using the current framework set-up (particularly the way we use require).

simondate:
It would be nice if there were some kind of 'developer-mode' that can be enabled for certain accounts (maybe super admins) that have access to advanced features that are only required by developers/system admins.

taylortom:
@simondate I've set up a 'super toolbar' in the Kineo code which adds a toolbar to the top of the main window for super admins only, and can house any kind of functionality. See here for code: https://github.com/cgkineo/adapt_authoring/tree/staging/frontend/src/plugins/superToolbar

I'm not sure how I feel about this kind of thing in the open-source code though, feels a bit hacky.

simondate:
Is there a way to automate this? 

If any version change has occurred rebuild the course?

PR #2128, opened 2018-11-09
Course force rebuild 

lc-thomasberger:
fixes #1584 #1497 

Based on https://github.com/adaptlearning/adapt_authoring/pull/1584 without the keyboard modifier.

image


dancgray:
Works for me, good to surface this in the UI, but I think users will want it for publish and export. Would it be easier to have a toggle below the buttons with a rebuild option that would apply to all three and be stored as a preference?

lc-thomasberger:
Does export and publish not always use a force rebuild? That's what I expected. If not a keyboard shortcut would probably the easiest solution to add force rebuild to all export types.

dancgray:
@lc-thomasberger don't think it does. [Publish](https://github.com/adaptlearning/adapt_authoring/blob/release/0.5.1/plugins/output/adapt/publish.js#L131) is still checking for rebuild flag. Export always runs publish.

taylortom:
Sorry to dive in late on this, but does anyone else think we should update publish & export to force a rebuild each time? (I also thought this happened already) - I don't think it makes sense to skip it with those functions.

lc-thomasberger:
@taylortom that is probably what users expect.

Should we keep the toggle button or should we roll back to the "dropdown" style?

lc-thomasberger:
@canstudios-nicolaw this would increase preview time as far as I am aware.

canstudios-nicolaw:
@lc-thomasberger Ok, so just to be clear we are thinking of having publish and export force a rebuild every time, and then preview will have it's own force rebuild option in the UI?

lc-thomasberger:
Exactly.
If you only change the json, preview will be faster without a force build. This is already handled by the Tool. If you would trigger a force rebuild you would need to change the theme or toggel the setting to create sourcemaps. This was a hack. The toggle button exposes this (useful) feature to the UI.

If you download or export a course, you would actually expect that the course is exported with the "latest" settings and plugin versions.

canstudios-nicolaw:
:+1: Makes sense

taylortom:
I hate to throw a spanner in the works, but did we mean to go back to the dropdown UI button? (@dancgray suggested the toggle back when it also affected publish/export). I really liked the DD button, and think it makes more sense from a UX perspective:
- Probably not something we should encourage too much, so good that it's partly hidden
- Feels a bit clunky to: turn on > publish > turn off (reminds me of those people who use caps lock instead of shift to capitalise a word...), I don't think this a feature that users will want to turn on/off globally

lc-thomasberger:
I can bring the dropdown back.

lc-thomasberger:
It does. It closes when preview is completed or when you click somewhere else. Except inside the drop-down element.

PR #2129, opened 2018-11-09
build app as part of CI

### LINKED ISSUE #1949: CI process should build the app
We don't currently run a `grunt build` as a part of the CI. I think the following should suffice:
```
script:
  - grunt build:prod
```

lc-thomasberger:
fixes #1949


.travis.yml
```diff
@@ -29,13 +29,12 @@ services:
...
   - git --version
   - mongod --version
-  - npm install -g adapt-cli
-  - adapt --version
   - npm install -g grunt-cli
 
 install:
   - npm config set spin false
-  - npm install
+  - npm install --production
```
tomgreenfield: These needs to remain without the `--production` flag so that the tests can run.

PR #2139, opened 2018-11-29
Sidebar refactor

### LINKED ISSUE #1496: Make sidebar buttons dynamic to allow for 'hot-swappable' pages/areas
Main example being theme picker/potential theme settings page -- but would also open up the possibility of adding extra custom plugins.

taylortom:
# ☠️ work in progress ☠️

Fixes #1496 

## Headlines
- SidebarView is no longer an empty container for each page's specific sidebar, but rather a customisable view in itself
- Sidebar elements are now grouped into the following categories: back button, **action buttons** (i.e. save/cancel, preview), **link buttons** (i.e. 'My courses', 'Project settings'), **fieldsets** (the filters used in forms -- think these should eventually go into the editors module as a custom widget) and **custom widgets** (anything else...) -- this allows us to remove a lot of shared styles and functionality
- Sidebar can be customised via the global `Origin.sidebar` object (add different types of buttons, show/hide etc.)
- All child sidebars removed from page views (`.hbs` and `.js`). There was a huge amount of duplication here

## Extra amends
- Location view is now automatic, and tied to router events, rather than having to manually set the page title when adding the view (if there's no string, the title is hidden)
- I've started to move language strings across to the frontend folder, with each module-specific set of strings in `modules/MODULE/lang/en.json` (and misc general strings in `core/lang`) - the `routes/lang` folder should only really contain the final concatenated file (note I've also started to namespace them in nested objects)


frontend/src/modules/editor/extensions/index.js
```diff
@@ -2,23 +2,21 @@
...
-    Origin.sidebar.addView(new EditorExtensionsEditSidebarView().$el, {
-      "backButtonText": backButtonText,
-      "backButtonRoute": backButtonRoute
+    var isMenu = data.type === "menu";
+
+    console.log(data);
+
+    Origin.sidebar.update({
+      breadcrumb: {
+        label: isMenu ? "Back to menu" : "Back to page",
```
lc-thomasberger: should be localisable

PR #1936, opened 2018-04-05
Snackbar Notify plugin

taylortom:
Adds [snackbar](https://material.io/guidelines/components/snackbars-toasts.html) functionality to the authoring tool.

PR #2153, opened 2018-12-11
Snackbar Notify extension

### LINKED ISSUE #934: Fix toast notifications
See https://github.com/adaptlearning/adapt_authoring/tree/issue/934 for demo code.

---
### Notify.snackbar()

Displays a small notification at the top of the screen. Can contain an action (by means of a button with a callback). You can customise your snackbars with the following attributes:

| Name | Description |
| :-- | :-- |
| type | The style for your snackbar: alert, info (default) or error |
| text | The text to be displayed |
| buttonText | The text shown by the action button (when used in conjunction with persist) |
| persist | Keeps the notification on-screen until the user dismisses it (overrides any timeout values) |
| animTime | The duration of the snackbar's animation (default: 250ms) |
| timeout | The amount of time the snackbar should stay on-screen (default: 3000ms) |
| callback | A function to call when the notification closes (either automatically, or from the user clicking the close button) |
#### Usage

```
Origin.Notify.snackbar({
  type: 'error',
  text: "Could not connect to server!",
  buttonText: 'Click to log out',
  persist: true,
  animTime: 500,
  callback: function() {
    Origin.Notify.alert('Snackbar closed!');
  }
});
```

It is also possible to just pass a string value to the Snackbar function. In this case, all defaults will be used:

```
Origin.Notify.snackbar("Just a string!");
```

lc-thomasberger:
fixes #934 

based on https://github.com/adaptlearning/adapt_authoring/pull/1936


frontend/src/modules/notify/plugins/snackbar/index.js
```diff
@@ -0,0 +1,71 @@
...
+  var processQueue = function() {
+    el = document.createElement('div');
+    el.id = 'snackbar';
+    var data = queue[0];
+    var html = ['<div class="body">',data.text,'</div>'];
+    if (data.persist && data.buttonText) {
+      html.push('<button class="close">');
+      html.push(data.buttonText);
+      html.push('</button>');
+    }
```
tomgreenfield: Just to add that these elements can be much easier to create using jQuery:
```js
$el = $('<div>', { id: 'snackbar', 'class': data.type })
  .append($('<div>', { 'class': 'body', text: data.text });
```
or:
```js
$el = $('<div>', {
  id: 'snackbar',
  'class': data.type,
  html: $('<div>', { 'class': 'body', text: data.text })
});
```
Appending the button later would be:
```js
$el.append($('<button>', { 'class': 'close', text: data.buttonText }));
```

dancgray:
Nice functionality. Not sure about the styling though. Would we want to be able to customise the styling depending on the module that created the snackbar?

canstudios-nicolaw:
I think that the different types (info, warning etc.) offer enough in terms of different styling to start with. I'd be hesitant to say we should add styles for different modules without a specific use case in mind, unless you have one already @dancgray ?

dancgray:
Nothing specific in mind @canstudios-nicolaw. If the existing types are enough to differentiate I'm happy.

PR #2156, opened 2018-12-13
drag and drop hotgraphic pin editor

lc-thomasberger:
## screenshot
image

## Features
- "Add Item" Button below image opens modal (same as clicking the add button to add items)
- Each item adds a draggable pin that can be used to change position of the pin
- clicking o the pin opens the modal to edit the item (same as clicking the list item below)
- editor attributes are used to customize the behavior:

| input config | description |
|---|---|
|items|name of the items attribute |
|topAttribute|name of the top / y psoition attribute |
|leftAttribute|name of the left / x psoition attribute|
|precision|floating precision defaults to 2 |
|pluginType| `"menu"` or `"extension"` |
|targetAttribute| name of the extension attribute (required to make it work with custom extensions) |
|addLabel | Add button label, defaults to 'app.add' |

### example schema config for hotgraphic
```
"src": {
          "type": "string",
          "required": false,
          "default": "",
          "inputType": {
                "type": "AssetItem",
          	"items": "_items",
          	"topAttribute": "_top",
          	"leftAttribute": "_left",
          	"precision": 2
          },
          "validators": [],
          "help": "This is the image that appears behind the pins"
}
```


lc-thomasberger:
@adaptlearning/authoring-tool-push 

This will require a way to specify compatible Autoringtool versions inside plugins. I think we should add a `server` or `authoringtool` attribute as we did for the `framework`.

We might also compare the used editors specified in the properties and compare them with the available editors in the AT. This would probably be much more work than introducing a compatible Authoringtool version number.

If a user tries to upload a plugin that uses an editor that is not available in the current Authoringtool version it will break.

lc-thomasberger:
Example setup for Hotgraphic is available here: https://github.com/adaptlearning/adapt-contrib-hotgraphic/tree/feature/assetItemEditor

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,211 @@
...
+    },
+
+    initialize: function(options) {
+      ScaffoldAssetView.prototype.initialize.apply(this, arguments);
+      this.init(options);
+    },
+
+    init: function(options) {
+      this.dragItems = [];
+      this.precision = options.schema.editorAttrs['data-precision'] || 3;
```
tomgreenfield: We don't need to shoehorn attributes into Backbone Form's `editorAttrs` – that's designed for attributes to be used in the DOM.

Instead, use the object form of the `inputType` attribute e.g.
```json
"inputType" : {
  "type": "AssetItem",
  "precision": 2
}
```
In the view:
```js
var attributes = options.schema.inputType;

this.precision = attributes['data-precision'] || 3;
```

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,211 @@
...
+        }, this);
+      }.bind(this));
+    },
+
+    findItemView: function() {
+      var form = Origin.scaffold.getCurrentForm();
+
+      switch (this.schema.editorAttrs['data-type']) {
+        case 'extension':
+          return form.fields._extensions.editor.nestedForm.fields[this.schema.editorAttrs['data-targetAttribute']].editor.nestedForm.fields[this.schema.editorAttrs['data-items']].editor;
```
tomgreenfield: Can we somehow make these a little easier to digest?

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,211 @@
...
+      this.leftAttr = options.schema.editorAttrs['data-leftAttribute'] || '_left';
+      this.itemsPath = options.schema.editorAttrs['data-items'] || 'properties._items';
+      this.itemsPath = this.itemsPath.split('.');
+    },
+
+    postRender: function() {
+      if (!this.itemListView) {
+        this.itemListView = this.findItemView();
+        this.listenTo(this.itemListView, {
+          'add': this.onItemListAdd,
```
tomgreenfield: No need for quotes.

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,211 @@
...
+    },
+
+    removeDragItems: function() {
+      this.dragItems.forEach(this.removeDragItem, this);
+      this.dragItems = [];
+    },
+
+    addDragItem: function(itemView) {
+      var value = itemView.value;
+      var index = itemView.$el.closest('.list-item').index();
```
tomgreenfield: Is it possible to switch to utilise a `data-index` attribute or similar instead of the DOM order?

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,211 @@
...
+          'add': this.onItemListAdd,
+          'remove': this.onItemListRemove,
+          'change': this.onItemListChange
+        });
+      }
+      // double defer:
+      // if external asset, postRender is triggered before el is injected into Dom
+      _.defer(function() {
+        this.removeDragItems();
+        this.itemListView.items && this.itemListView.items.forEach(function(item) {
```
tomgreenfield: Please cache `this.itemListView.items`.

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,211 @@
...
+        css: {
+          top: parseInt(value[this.topAttr]).toFixed(this.precision)+'%',
+          left: parseInt(value[this.leftAttr]).toFixed(this.precision)+'%',
+          position: 'absolute'
+        },
+        text: index + 1
+      });
+
+      this.dragItems.push({
+        elm: elm,
```
tomgreenfield: Suggest renaming with a jQuery prefix e.g. `$el`.

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,211 @@
...
+      this.addDragItem(listItemView);
+    },
+
+    onItemListChange: function(listView) {
+      this.dragItems.forEach(this.updateDragItem, this);
+    },
+
+    onDragStop: function(event, ui) {
+      var $parent = this.$('.scaffold-asset-item-img-holder');
+      var left = (parseInt($(event.target).css("left")) / ($parent.width() / 100)).toFixed(this.precision);
```
tomgreenfield: Please cache `event.target` & `$(event.target)`.

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
lc-thomasberger: We would need to patch scaffoldListView for this.

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,221 @@
...
+      this.init(options);
+    },
+
+    init: function(options) {
+      var attrs = options.schema.inputType;
+      this.dragItems = [];
+      this.precision = attrs.precision || 3;
+      this.topAttr = attrs.topAttribute || '_top';
+      this.leftAttr = attrs.leftAttribute || '_left';
+      this.itemsPath = attrs.items || 'properties._items';
```
tomgreenfield: Apologies if I'm missing something obvious but what is `itemsPath` used for?

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,221 @@
...
+      switch (this.schema.inputType.type) {
+        case 'extension':
+          baseEditor = form.fields._extensions.editor.nestedForm.fields[targetAttr].editor;
+        case 'menu':
+          baseEditor = form.fields.menuSettings.editor.nestedForm.fields[targetAttr].editor;
+        default:
+          baseEditor = form.fields.properties.editor;
+      }
+
+      return baseEditor.nestedForm.fields[itemAttr].editor;
```
tomgreenfield: This function is looking a lot better. A few things:
* We ought to be able to handle a schema with just `"inputType": "AssetItem"` since we've already gone some of the way by applying fallbacks in the `init` function
* `this.schema.inputType.type` needs a different name since `type` would be `AssetItem` now. Is there a way to test the extension/menu situations?
* I'd cache `this.schema.inputType` in any case
* Switch statement is missing its `break`s

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
lc-thomasberger: Thanks, this must be a left over. Initially I thought it could be used to specify the path to the items attribute that is used for the pins.

lc-thomasberger:
@tomgreenfield thanks for testing and reviewing.

To test menus / extensions, I copied the example json into a plugin and compared the generated json using devtools Network panel in course preview.
If you are on a devsetup you can also update the schema in the databse directly.

lc-thomasberger:
@tomgreenfield I created a dummy extension. There is no functionality. Once enabled, you can use it for blocks and test the generated JSON from the editor. 
[adapt-assetItemExample.zip](https://github.com/adaptlearning/adapt_authoring/files/2860533/adapt-assetItemExample.zip)

frontend/src/modules/scaffold/views/scaffoldAssetItemView.js
```diff
@@ -0,0 +1,225 @@
...
+      child.view.setValue(_.extend({}, child.view.value, value));
+      child.view.renderSummary();
+    },
+
+    onAssetButtonClicked: function(event) {
+      event.preventDefault();
+
+      Origin.trigger('modal:open', AssetManagementModalView, {
+        collection: new AssetCollection,
+        assetType: 'Asset:image',
```
dancgray: This limits it to image type assets. It would be good to be able to specify in the schema which type of asset can be selected. Alternatively the naming should imply this is image only.

tomgreenfield: On a related note, as part of the Scaffold refactor, I did lay the groundwork to avoid these colon-separated types:
#2156 
#### Old
```json
"inputType": "Asset:image"
```
#### New
```json
"inputType": {
  "type": "Asset",
  "media": "image"
}
```
Asset management does still need tweaking to work with it though, e.g.
https://github.com/adaptlearning/adapt_authoring/blob/32be3c1bb96d7ecef7d4bf46e1bf6b18f36226ef/frontend/src/modules/assetManagement/views/assetManagementModalView.js#L28

lc-thomasberger: Yes it is currently only limited to images.
For what other media types would you like to have x/y coordinates @dancgray ?

dancgray: Video would be a possibility @lc-thomasberger

lc-thomasberger: Then you would have to handle time. So it would be better to have a new scaffold editor for this.

dancgray: Only if you wanted the items to record a time position as well as x/y. But happy to leave this until we have a more fleshed out requirement for other asset types.

lc-thomasberger: In my opinion it does only make sense to have the position dependent on time for videos.

Link2Twenty:
@lc-thomasberger is it possible to use this if the pins are at a deeper level? I have a component that can have several images and each image can have pins.

```json
"_items": {
  "type": "array",
  "required": true,
  "title": "Angles",
  "help": "Each angle presented",
  "items": {
    "type": "object",
    "required": true,
    "properties": {
      "image": {
        "type": "string",
        "required": false,
        "default": "",
        "title": "Image",
        "inputType": "Asset:image",
        "validators": [],
        "help": "Image to be displayed at this specific angle",
        "translatable": false
      },
      "_pins": {
        "type": "array",
        "required": true,
        "title": "Pins",
        "help": "Each pin",
        "items": {
          "type": "object",
          "required": true,
          "properties": {
            "x": {
              "type": "number",
              "required": false,
              "default": "",
              "title": "X coord",
              "inputType": "Text",
              "validators": [],
              "help": "X coordinate for pin",
              "translatable": false
            },
            "y": {
              "type": "number",
              "required": false,
              "default": "",
              "title": "Y coord",
              "inputType": "Text",
              "validators": [],
              "help": "Y coordinate for pin",
              "translatable": false
            },
          }
        }
      }
    }
  }
}
```

lc-thomasberger:
@Link2Twenty I have not tested it with the editor in an array. 
Listitem Editors are rendered in a drawer, with a limited width. I think this will break the dragAndDrop Editor.

Link2Twenty:
The drag and drop seems to work in there, but if I do

```js
"inputType": {
  "type": "AssetItem",
  "items": "_items",
  "topAttribute": "x",
  "leftAttribute": "x",
  "precision": 2
}
```

it adds a point for each item rather than per pin, as you would expect. I tried `"items": "_pins"` and `"items": "_items._pins"` but they both gave an error.

lc-thomasberger:
This would probably require a more complex "attribute find algorithm". Not sure how useful and bulletproof it will be.

PR #2160, opened 2018-12-17
handle unshared courses on user delete

### LINKED ISSUE #2124: A user's courses are 'orphaned' on user delete
## Expected Behaviour
Courses should still be accessible after the owner user is deleted. If a course hasn't been set to shared, this won't be the case.

## Possible solutions:
- Owner is changed (possibly to the user which deleted the original owner)
- Course is set to shared automatically on user delete
- All the user's courses are deleted along with the user
- Admin users are given access to all system courses (related to https://github.com/adaptlearning/adapt_authoring/issues/1922)

## Actual Behaviour
If a course isn't shared, and its original owner deleted, it won't be accessible via the UI.

## Versions
  - Authoring Tool Version: >= 0.3.0

lc-thomasberger:
fixes #2124

When a user is deleted a admin user can choose what to do with the users unshared courses.
Supported options are:
- transfer courses to admin user
- delete unshared courses
- share all courses 

image


lib/helpers.js
```diff
@@ -146,22 +147,28 @@ function hasCoursePermission (action, userId, tenantId, contentItem, next) {
...
         logger.log('error', err);
-        return next(err);
-      }
-      if (!results || results.length !== 1) {
-        return next(new Error('Course ' + courseId + ' not found'));
+        return err;
       }
 
-      var course = results[0];
+      db.retrieve('course', { _id: courseId }, { jsonOnly: true }, function (err, results) {
```
dancgray: This line is missing }); after argument list.

lib/usermanager.js
```diff
@@ -325,32 +325,72 @@ exports = module.exports = {
     });
   },
 
+  /**
+    * handle a users unshared courses on delete
+  */
+  handleUserCoursesOnDelete: function(userId, courseOption, done) {
+    // todo handle user courses
```
canstudios-nicolaw: Please can you remove this now

dancgray:
Looks good. For the transfer ownership and share, how much do we care that there will be orphaned createdBy ID's in a lot of collections (contentobjects, article, block, components, assets, courseassets, tags)?

Also the share option will leave courses that cannot be deleted.

canstudios-nicolaw:
Does #1926 need to go in before this?

lc-thomasberger:
@canstudios-nicolaw should be fine.
You can test the "share all" option with this PR, then checkout #1926 and you should be able to delete the shared courses.

@dancgray thanks for pointing out the orphaned A/B/C createdBy id's. Will have a look. Thought `CourseContent` does all the work for me 😄

lc-thomasberger:
Might be useful to add a `updateChildren` method similar to `destroyChildren` as implemented here https://github.com/adaptlearning/adapt_authoring/blob/master/lib/contentmanager.js#L376

For the "share" action, we can set the `createdBy` id to the user id that deleted the user?

canstudios-nicolaw:
I have just tested this, a few small points.

- I think the wording on this could be a little clearer. Perhaps "You will be the owner of all of their unshared courses." instead?
transferownershiptome

- I think the option in the dropdown should read "Delete all unshared courses" rather than "Delete all courses". This is more consistent with the following warning modal and the functionality.

- Similar to above, I suggest a text change here to "All of their courses will be shared."
shareall

routes/lang/en-application.json
```diff
@@ -14,7 +14,13 @@
   "app.addusertitle": "Add new user",
   "app.disableuser": "Disable user",
   "app.deleteuser": "Delete user",
-  "app.confirmdeleteuser": "Deleting<br/><b>%{email}</b>.<br/><br/>This is a one-way trip. Continue?",
+  "app.deleteusertransfer": "Transfer ownership to me",
+  "app.deleteuserdelete": "Delete all unshared courses",
+  "app.deleteusershare": "Share all courses",
+  "app.confirmdeleteuser": "Deleting<br/><b>%{email}</b>.<br/><br/>%{courseOption}<br/><br/>This is a one-way trip. Continue?",
+  "app.confirmdeleteusertransfer": "<b>You will be the owner of all * their unshared courses.</b>",
```
canstudios-nicolaw: "of" instead of "*" here?

routes/lang/en-application.json
```diff
@@ -14,7 +14,13 @@
...
+  "app.confirmdeleteuserdelete": "<b>All unshared courses will be deleted.</b>",
```
canstudios-nicolaw: Perhaps "All of their unshared courses will be deleted." to be consistent with the other messages

PR #2165, opened 2018-12-19
Add defensive coding for import process

### LINKED ISSUE #2164: Error during import: Cannot read property 'pluginLocations' of undefined
```console
TypeError: Cannot read property 'pluginLocations' of undefined
    at /mnt/data/adapt/testing.kineoadapt.com/plugins/output/adapt/importsource.js:335:30
```
https://github.com/adaptlearning/adapt_authoring/blob/41a8185ece9c6f90d35cbf97878aeab6a0477e24/plugins/output/adapt/importsource.js#L333

On the `release/0.5.1` branch, I encounter this error due to `plugin.properties` being non-existent on a couple of plugins' metadata already in the database. Should just need an amend to something like:
```js
if(plugins.properties && plugin.properties.pluginLocations) {
```

tomgreenfield:
Resolves #2164.

### DIFF
--- a/plugins/output/adapt/importsource.js
+++ b/plugins/output/adapt/importsource.js
@@ -321,6 +321,8 @@ function ImportSource(req, done) {
         return cb(error);
       }
       async.each(results, function(plugin, cb2) {
+        const properties = plugin.properties;
+        const locations = properties && properties.pluginLocations;
         if(!metadata[`${type}Map`]) {
           metadata[`${type}Map`] = {};
         }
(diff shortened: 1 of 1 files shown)


tomgreenfield:
That looks like an intentional error has been chosen to be thrown as opposed to the uncaught one in the original issue.

I don't think specifying a valid framework version is an unfair requirement for a plugin.

That said, I'm not an import guru!

lc-thomasberger:
Not sure either. I would expect to get a warning if the bower is missing a value though.

tomgreenfield:
Do you mean changing the wording of the existing error on the frontend? Might be good to record this in a separate issue.

image

lc-thomasberger:
Separate issue sounds good.

PR #2166, opened 2018-12-19
expose installed versions 

### LINKED ISSUE #1944: Expose authoring tool version in the HTML
Would be nice to easily check the version of an authoring tool from the HTML. This could also be exposed via the JS and the config module, but it seems more useful for debugging to be able to check this without needing JS.

This could also be used to apply custom version-specific features/styling (not suggesting we do this in core, but can imagine it would be useful for custom work).

lc-thomasberger:
fixes #1944


routes/index/index.js
```diff
@@ -1,26 +1,45 @@
...
+
+  getVersions(function(versions) {
+    res.render('index', Object.assign({
+      isProduction: isProduction,
+      dateStampAsString: dateStampAsString,
+      loading: app.polyglot.t('app.loading'),
+      productName: app.polyglot.t('app.productname'),
+      useAnalytics: useAnalytics,
+      trackingId: trackingId
+    }, versions));
```
tomgreenfield: Is it possible to use async/await here for readability?

routes/index/index.hbs
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" data-adapt-framework="{{adapt_framework}}" data-adapt-authoring="{{adapt_authoring}}">
```
tomgreenfield: Please can we stick with `data-adapt-framework-version` etc., for alignment with the [framework](https://github.com/adaptlearning/adapt_framework/blob/1cf22d9cb4dae444723f69bff52f3f3883d3d3d9/src/core/js/app.js#L209)?

routes/index/index.js
```diff
@@ -1,26 +1,44 @@
...
     productName: app.polyglot.t('app.productname'),
     useAnalytics: useAnalytics,
     trackingId: trackingId
-  });
+  }, versions));
+});
 
-});
\ No newline at end of file
+function getVersions(cb) {
```
tomgreenfield: Putting the `async` keyword in front of the function would mean you won't have to explicitly return a promise – you'd only need to return `_versions` which would be wrapped automatically.

routes/index/index.js
```diff
@@ -1,26 +1,42 @@
...
     productName: app.polyglot.t('app.productname'),
     useAnalytics: useAnalytics,
     trackingId: trackingId
-  });
+  }, versions));
+});
 
-});
\ No newline at end of file
+async function getVersions(cb) {
```
tomgreenfield: Please remove unused `cb`.

PR #2171, opened 2018-12-21
use npm to install ffmpeg and ffprob

### LINKED ISSUE #1860: Remove requirement for users to install ffmpeg manually
Moving comments on #850 to a separate issue

@lc-thomasberger:

> I would recommend to always show an image for assets as it is very tedious to work with assets that have no thumbnail. If there is no thumbnail generated, I would fallback to the original image and use this.
>
> also recommend that for simple thumbnail creation from uploaded images we should not only depend on ffmpeg as it requires the user to manually install ffmpeg.
> There are quite some nice npm packages available that could be used under the hood and installed with npm during setup.
>
> https://www.npmjs.com/package/jimp
> https://www.npmjs.com/package/sharp

lc-thomasberger:
fixes #1860 

## Advantages:
- make installation easier
- ffmpeg / ffprobe will be available on all systems
- we can safely create thumbnails
- all installs will have the same ffmpeg / ffprobe version

## Disadvantages:
- the static npm modules install the binaries for all plattforms (slightly more diskspace)

## Todos
- [x] remove useffmpeg from config and install
- [ ] update docs ( **after release** )

## Server-rewrite
Thumbnail creation should be it's own module so it can be replaced with other implementations.
We may also use different libraries for different media types (ffmpeg -> video, sharp|jimp -> images).


taylortom:
Looks really good, would be great to get this in asap 🎉

plugins/filestorage/localfs/index.js
```diff
@@ -351,9 +355,8 @@ LocalFileStorage.prototype.copyAsset = function(asset, sourceTenantName, destina
 
 LocalFileStorage.prototype.createThumbnail = function (filePath, fileType, options, next) {
   // early return if we can't create thumbnails
-  if (!configuration.getConfig('useffmpeg')) {
-    return next(null, false);
-  }
+  if (!ffmpegStatic || !ffmpegStatic.path) return next(null, false);
```
tomgreenfield: Hi Thomas, I'm not entirely sure when this sort of logic would be satisfied? If I don't have the NPM modules installed, an error is thrown much earlier:
```console
application.js:630 Error: Cannot find module 'ffmpeg-static'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (D:\adapt_authoring\plugins\filestorage\localfs\index.js:6:22)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at D:\adapt_authoring\lib\filestorage.js:204:23
    at PluginManager.getPlugin (D:\adapt_authoring\lib\pluginmanager.js:175:9)
    at Object.getStorage (D:\adapt_authoring\lib\filestorage.js:197:17)
    at D:\adapt_authoring\lib\assetmanager.js:750:19
    at D:\adapt_authoring\node_modules\mongoose\lib\model.js:4669:16
    at D:\adapt_authoring\node_modules\mongoose\lib\utils.js:258:16
    at _hooks.execPost (D:\adapt_authoring\node_modules\mongoose\lib\query.js:4065:11)
    at D:\adapt_authoring\node_modules\mongoose\node_modules\kareem\index.js:135:16
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

lc-thomasberger: Tanks for testing.
Valid point. I thought about when the module failed to install the binaries for the platform.
Not sure how realistic this is though.

I basically used the existing check and mended it. I am happy to remove it though.

PR #2172, opened 2018-12-21
change add button position

### LINKED ISSUE #1587: Components with "Add" items,  button location
A small UI thing, on components with the add item button, including when it is in drill down menus ( ie matching question component)

Can the Add button be stuck to the top where it says Items:

image

i can add about 3-4 items before the button drops off the screen
knowing where the button will always be will help improve speed when adding multiple items, instead of:
add item 
scroll to to get button into view
add item
scroll to to get button into view
add item
etc

lc-thomasberger:
fixes #1587


canstudios-nicolaw:
Just checking - this should definitely be going into develop and not a release/0.7.0 branch?

tomgreenfield:
I believe the plan we discussed was to ditch the develop branch for good, leaving just release branches as appropriate.

PR #2185, opened 2019-01-17
Get schema from framework clone

### LINKED ISSUE #2177: Load in schema from framework clone
Currently the schema is [hard-coded](https://github.com/adaptlearning/adapt_authoring/tree/master/plugins/content) to each authoring tool release. We discussed instead loading in the schema which is brought down with the [framework clone](https://github.com/adaptlearning/adapt_framework/tree/accessibilityv4/src/core/schema) used for building courses.

This could aid with supporting the upcoming framework 4 release for clients that would like to use it immediately. We could also see this as a stepping stone towards supporting multiple framework versions.

tomgreenfield:
Resolves #2177.

Tested with various versions of framework 2, 3 & 4.

**This update requires that the authoring tool is running framework 2.0.14 or above.**

### DIFF
--- a/lib/contentmanager.js
+++ b/lib/contentmanager.js
@@ -105,7 +105,7 @@ ContentPlugin.prototype.onDatabaseCreated = function (db, next) {
 
   fs.readFile(schemaPath, function (err, data) {
     if (err) {
-      logger.log('info', `skipping ${schemaPath} as framework clone not yet present`);
+      logger.log('info', 'could not read schema file at ' + schemaPath);
       return next(null);
     }
 
--- a/lib/contentmanager.js
+++ b/lib/contentmanager.js
@@ -105,8 +105,8 @@ ContentPlugin.prototype.onDatabaseCreated = function (db, next) {
 
   fs.readFile(schemaPath, function (err, data) {
     if (err) {
-      logger.log('error', 'could not read schema file at ' + schemaPath);
-      return next(err);
+      logger.log('info', `skipping ${schemaPath} as framework clone not yet present`);
+      return next(null);
     }
 
     try {
--- a/lib/contentmanager.js
+++ b/lib/contentmanager.js
@@ -24,7 +24,15 @@ var path = require('path'),
  * CONSTANTS
  */
 var MODNAME = 'contentmanager',
-    WAITFOR = 'pluginmanager';
+    WAITFOR = 'pluginmanager',
+    COURSE_COLLECTIONS = [
+      'article',
+      'block',
+      'component',
+      'config',
+      'contentobject',
+      'course'
+    ];
 
 // errors
 function ContentTypeError (message) {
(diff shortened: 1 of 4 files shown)


canstudios-nicolaw:
Thanks for this @tomgreenfield very exciting :) 

Please could you advise on the best way to test this? Is a fresh install required per test?

tomgreenfield:
Only need fresh installs if you'd like to check the `node install` side of things. Travis takes care of the changes to the setup of the test environment.

Otherwise I've been using our own upgrade script to switch between framework versions:
```console
$ node upgrade
This script will update the Adapt authoring tool and/or Adapt Framework. Would you like to continue?
> Continue? Y/n (Y)
> Update automatically? Y/n (Y) n
> Specific git revision to be used for the authoring tool. Accepts any valid revision type (e.g. branch/tag/commit)
> Specific git revision to be used for the framework. Accepts any valid revision type (e.g. branch/tag/commit) v4.0.0
```
You can see if the schema has been successfully changed by checking for new form fields:

aria

tomgreenfield:
You may have noticed I tinkered with the idea of the `release/schema` branch if you glance at the PR's history. Changed my mind to avoid the subsequent duplicate PR into master if we're releasing imminently.

Happy to reinstate the branch if you like?

lib/contentmanager.js
```diff
@@ -93,8 +105,8 @@ ContentPlugin.prototype.onDatabaseCreated = function (db, next) {
 
   fs.readFile(schemaPath, function (err, data) {
     if (err) {
-      logger.log('error', 'could not read schema file at ' + schemaPath);
-      return next(err);
+      logger.log('info', `skipping ${schemaPath} as framework clone not yet present`);
```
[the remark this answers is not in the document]
tomgreenfield: A very valid point; was being a bit light-touch with the amends.

Open to suggestions to make this more robust?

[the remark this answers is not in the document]
canstudios-nicolaw: Yes perhaps just revert the message at this stage

canstudios-nicolaw:
I appreciate that for the sake of a single PR release, it is hassle to merge via an identical release branch. At the same time I think it's cleaner/safer to try and keep master to release branches only. I think I'd be in favour of a release branch even though it seems a bit unnecessary in this case.

lib/contentmanager.js
```diff
@@ -93,8 +105,8 @@ ContentPlugin.prototype.onDatabaseCreated = function (db, next) {
 
   fs.readFile(schemaPath, function (err, data) {
     if (err) {
-      logger.log('error', 'could not read schema file at ' + schemaPath);
-      return next(err);
+      logger.log('info', 'could not read schema file at ' + schemaPath);
+      return next(null);
```
canstudios-nicolaw: Should this log still be type error and should this return the error rather than null?

tomgreenfield: If we intentionally return the error then it will revert to the previous functionality where none of the schema is loaded if one is missing.

canstudios-nicolaw: Ok :+1: I did wonder but wanted to check

PR #2186, opened 2019-01-18
Fix config.dbAuthSource

### LINKED ISSUE #1977: Gruntfile: `authSource` should be `dbAuthSource `
In Gruntfile, the task [`migration-conf`](https://github.com/adaptlearning/adapt_authoring/blob/develop/Gruntfile.js#L256) refers to `config.authSource` but it should be `config.dbAuthSource` as the [database](https://github.com/adaptlearning/adapt_authoring/blob/develop/lib/dml/mongoose/index.js#L90) use that key.

Fixed in #1970

canstudios-nicolaw:
Fixes #1977


canstudios-simonm:
Internally verified as working

PR #2194, opened 2019-01-23
Admins should be able to give new users a name

### LINKED ISSUE #2188: Admins should be able to give new users a name
## Affected Area
User management

## Requested Feature
When an admin (super or tenant) creates a new user they can only give them an email address and a password. That person must then log in and set their own name via their profile. It would be useful if the admin could set this on creation.

## Use Case
Admins can create new users with all their details.

## Current Workaround
A newly created user must log in and set themselves a name via their profile.

canstudios-nicolaw:
Fixes #2188

### DIFF
--- a/frontend/src/modules/userManagement/templates/addUser.hbs
+++ b/frontend/src/modules/userManagement/templates/addUser.hbs
@@ -4,27 +4,27 @@
   </div>
   <form class="addUser forms" action="/api/user" method="post" autocomplete="off">
       <div class="form-group">
-        <label for="firstName">{{t 'app.firstname'}} <span class="req">*</span></label>
+        <label>{{t 'app.firstname'}} <span class="req">*</span></label>
         <div class="field-error display-none" data-error="">{{t 'app.validationrequired'}}</div>
         <input type="text" name="firstName" autocomplete="off">
       </div>
       <div class="form-group">
-        <label for="lastName">{{t 'app.lastname'}}<span class="req">*</span></label>
+        <label>{{t 'app.lastname'}}<span class="req">*</span></label>
         <div class="field-error display-none" data-error="">{{t 'app.validationrequired'}}</div>
         <input type="text" name="lastName" autocomplete="off">
       </div>
       <div class="form-group">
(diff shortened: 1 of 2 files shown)


frontend/src/modules/userManagement/views/addUserView.js
```diff
@@ -19,17 +19,36 @@ define(function(require){
...
+
+      if (!validEmail) {
+        this.$('#emailError').removeClass('display-none');
+        valid = false;
+      }
+
+      if (!validPassword) {
+        this.$('#passwordError').removeClass('display-none');
+        valid = false;
+      }
```
tomgreenfield: Would you mind tweaking this to be extensible? Something iterative like the below:
```js
  var $error = $(element);

  var isValid = $input.attr('name') === 'email' ?
    Helpers.isValidEmail($input.val().trim()) :
    $input.val().trim().length > 0;

  $error.toggleClass('display-none', isValid);

  if (!isValid) {
    valid = false;
  }
});
```

We can then lose the IDs in the DOM.

frontend/src/modules/userManagement/templates/addUser.hbs
```diff
@@ -3,15 +3,25 @@
     {{t 'app.adduserformtitle'}}
   </div>
   <form class="addUser forms" action="/api/user" method="post" autocomplete="off">
+      <div class="form-group">
+        <label for="firstName">{{t 'app.firstname'}} <span class="req">*</span></label>
```
lc-thomasberger: label's for attribute must match an input element id. Since we don't have any id's I would suggest to remove the for attribute at all. The label may wrap the input element so that it sets focus if a user clicks on the label text.

frontend/src/modules/userManagement/views/addUserView.js
```diff
@@ -19,17 +19,25 @@ define(function(require){
...
+
+        $error.toggleClass('display-none', isValid);
+
+        if (!isValid) {
+          valid = false;
+        }
+      });
+
+      if (valid) {
         this.$('.field-error').addClass('display-none');
```
tomgreenfield: This line's now superfluous now the loop's been added.

PR #2200, opened 2019-01-25
Remove early return

### LINKED ISSUE #1955: Options buttons broken if not in a group
If you've added an options button without specifying a 'group' attribute, the button doesn't fire callback event when clicked due to an early return. See [here](https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/modules/options/views/optionsView.js#L72-L74).

## Steps to Reproduce
1. Add a new options button without specifying group attribute
1. Click button
1. Note no event

taylortom:
Fixes #1955

### DIFF
--- a/frontend/src/modules/options/views/optionsView.js
+++ b/frontend/src/modules/options/views/optionsView.js
@@ -73,15 +73,16 @@ define(function(require) {
     },
 
     setSelectedOption: function(selectedOption) {
+      // if this item is in a group toggle all group items to not selected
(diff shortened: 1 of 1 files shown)


lc-thomasberger:
The selected class is not applied to the element when you click it.

taylortom:
@lc-thomasberger have altered the logic to toggle the class for group buttons

tomgreenfield:
What are the steps to test this one, @taylortom?

PR #2201, opened 2019-01-25
Only remove one matching course asset record

### LINKED ISSUE #2141: Removing an asset from a component removes all other instances of that asset
## Expected Behaviour
Removing an asset should only affect that asset

## Actual Behaviour
If the same asset is used multiple times in a component all all records of that asset are removed from the courseassets collection for that component.

See scaffoldAssetView.js [#L116](https://github.com/adaptlearning/adapt_authoring/blob/develop/frontend/src/modules/scaffold/views/scaffoldAssetView.js#L116). Matches on componentID (contentTypeId), content type and asset name.
 
## Steps to Reproduce
1. Add the same image in a graphic components large and small image fields and save.
2. Replace the first image and save
3. Second image courseasset collection value is removed as well

## Versions
  - Framework Version: N/A
  - Node.js Version: v8.12.0
  - MongoDB Version: v3.6
  - Operating System: N/A
  - Browser: N/A

dancgray:
Fixes #2141

### DIFF
--- a/frontend/src/modules/scaffold/views/scaffoldAssetView.js
+++ b/frontend/src/modules/scaffold/views/scaffoldAssetView.js
@@ -126,16 +126,17 @@ define([
           return;
         }
 
-        // delete all matching courseassets and then saveModel
-        Helpers.forParallelAsync(courseassets, function(model, index, cb) {
-          model.destroy({
-            success: cb,
+        var listModels = courseassets.models ? courseassets.models.slice() : courseassets.slice();
+
+        if (listModels.length) {
+          listModels[0].destroy({
+            success: this.saveModel(),
             error: function() {
-              console.error('Failed to destroy courseasset record', model.get('_id'));
-              cb();
+              console.error('Failed to destroy courseasset record', listModels[0].get('_id'));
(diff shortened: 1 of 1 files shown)


frontend/src/modules/scaffold/views/scaffoldAssetView.js
```diff
@@ -126,16 +126,17 @@ define([
...
+            success: this.saveModel(),
             error: function() {
-              console.error('Failed to destroy courseasset record', model.get('_id'));
-              cb();
+              console.error('Failed to destroy courseasset record', listModels[0].get('_id'));
+              return;
             }
           });
-        }, this.saveModel.bind(this));
+        }
```
tomgreenfield: Suggest the logic be reversed to return early:
```js
var listModels = courseassets.models ? courseassets.models.slice() : courseassets.slice();
var listModel = listModels[0];

if (!listModel) return;

listModel.destroy({
  success: this.saveModel(),
  error: function() {
    console.error('Failed to destroy courseasset record', listModel.get('_id'));
    return;
  }
});
```

PR #2204, opened 2019-01-28
Share course with specific users

### LINKED ISSUE #2123: Share courses with specific users
## Requested Feature
Ability to share a course with specific users, rather than blanket share with all.

## Use Case
I just want to share a course with a limited number of people.

taylortom:
Fixes #2123 (framework issue [#2345](https://github.com/adaptlearning/adapt_framework/issues/2345)).

Also removed the unused shared project view and associated files.


lib/role/CourseCreator.json
```diff
@@ -6,6 +6,7 @@
     { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/shared/*", "action":["create","read"], "effect":"allow" },
     { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/subscribed/*", "action":["create","read"], "effect":"allow" },
     { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/asset/*", "action":["create", "read", "update"], "effect":"allow" },
+    { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/user/*", "action":["read"], "effect":"allow" },
```
taylortom: Adds read access to all users in the same tenant

plugins/content/course/model.schema
```diff
@@ -584,9 +584,22 @@
       "editorOnly": true,
       "inputType": "Checkbox",
       "validators": [],
-      "title": "Share with other users",
+      "title": "Share with all users",
       "help": "Controls whether or not your colleagues will be able to see this course from the 'Shared courses' option"
     },
+    "_shareWithUsers": {
```
taylortom: Adds extra sharing field

canstudios-nicolaw: TomG's PR is going to mean that these schemas get pulled in from the framework (https://github.com/adaptlearning/adapt_authoring/pull/2185) so I think that putting this here will mean that it is overwritten

lib/helpers.js
```diff
@@ -137,37 +137,40 @@ function isMasterPreviewAccessible(courseId, userId, next) {
...
+      const course = results[0];
 
-      var course = results[0];
-
-      if ((course._isShared || course.createdBy.toString() == userId) && course._tenantId.toString() == tenantId) {
-        // Shared courses on the same tenant are open to users on the same tenant
+      if(course._tenantId.toString() !== tenantId) { // no sharing across tenants
+        return next(null, false);
+      }
+      if(course.createdBy.toString() === userId) { // user's own course
```
canstudios-nicolaw: extra line above this

plugins/content/course/model.schema
taylortom: Have opened a framework PR here: https://github.com/adaptlearning/adapt_framework/pull/2346

lib/helpers.js
taylortom: Sorry, is this asking to remove the line break?

frontend/src/modules/scaffold/views/scaffoldUsersView.js
```diff
@@ -0,0 +1,98 @@
...
+
+      var $div = $('<div class="option">')
+        .append('<span class="name">' + (name || email) + '</span>');
+
+      if(name) {
+        $div.append('<span class="email">&lt;' + email + '&gt;</span>');
+      }
+      if(item.disabled) {
+        $div.attr('data-disabled', true);
+      }
```
tomgreenfield: Could switch to object-style syntax if you fancied:
```js
var $div = $('<div>', { 'class': 'option', 'data-disabled': item.disabled })
  .append($('<span>', { 'class': 'name', text: name || email }));

if (name) {
  $div.append($('<span>', { 'class': 'email', text: '<' + email + '>' }));
}

return $div;
```

frontend/src/modules/scaffold/views/scaffoldUsersView.js
```diff
@@ -0,0 +1,98 @@
...
+        labelField: 'email',
+        valueField: '_id',
+        options: users,
+        searchField: [ 'email', 'firstName', 'lastName' ],
+        render: {
+          item: this.renderItem,
+          option: this.renderItem
+        },
+        onItemRemove: function(value, $item) {
+          if(value === Origin.sessionModel.get('id')) {
```
tomgreenfield: Return early?

frontend/src/modules/scaffold/views/scaffoldUsersView.js
taylortom: Have replaced the above with a hbs template

lc-thomasberger:
When I share a course with all users, other users can't see the course.

lc-thomasberger:
A course that is shared with all users can't be deleted by the user that created it.

taylortom:
This is working with #1926, so will get that one merged first before looking again.

taylortom:
Have just noticed that there's some discrepancy between the styling for tags and users (I styled the users to match the AT theme colours). Should I apply this to the tags too?

screenshot 2019-02-20 at 12 10 15

canstudios-nicolaw:
@taylortom Would you expect anything to be required to test this other than npm install, grunt build and a restart?

I am getting a lot of errors in the console when I try and view dashboards or courses

screenshot from 2019-02-20 13 57 26

I can see `_shareWithUsers` in my course model.schema, I'm testing on an install that existed before TomG's schema release so I think it is using the schemas in the authoring tool.

lib/helpers.js
canstudios-nicolaw: I don't know :upside_down_face: happy to resolve

canstudios-nicolaw:
- If I log in and share the course with a user, then that user does appear to be able to edit, delete, download and export the course but they cannot preview it - this gives a 401 permission denied

- If I share the course with User A, then User A can share the course with everyone. This is a new scenario that we won't have come across before. Should the course owner be the only person who can share a course with everyone?

- I think it's a little strange that if I share a course with somebody I don't really have any way to see that from the dashboard. I think I would have expected to see it on the shared courses dashboard because it has been shared with somebody. The sharing settings are right next to each other in the project settings so I would hope that upon opening the course from there it would be easy to determine the type of sharing. I'm open to other opinions on this

taylortom:
Good questions,
1. Will have a look into this one.
2. This is as designed. If we agree on a different strategy, happy to change it.
3. I think this is more a failing of the dashboard in general - I've never liked the separate pages for mine/shared. Think a better approach would be to merge the two, and use filters/views to alter what's displayed. I also think we need a set of icons to show more info about each course (i.e. mine, shared, ... possibly other stuff)

canstudios-nicolaw:
Ok thanks @taylortom 

I'm fine with 2 & 3 for now then :+1:

lc-thomasberger:
I can replicate @canstudios-nicolaw issue above.

canstudios-nicolaw:
@taylortom I'm still having one of the issues above, if I share a course via the 'share with all' checkbox then I don't see it on my shared courses dashboard. This issue occurs whether I have also shared with some individuals or not. The rest is working well now though :+1:

lc-thomasberger:
I can not replicate. @canstudios-nicolaw can you please check your tenants?

taylortom:
@lc-thomasberger @canstudios-nicolaw I think this is related to courses which you own? I think I changed the API query slightly to omit any shared courses _not_ owned by you (as they're already shown in 'My courses'). Will revert this in light of the upcoming https://github.com/adaptlearning/adapt_authoring/issues/2238

PR #2207, opened 2019-01-30
broken sources in preview 

### LINKED ISSUE #2202: Course preview page favicon path is incorrect
https://github.com/adaptlearning/adapt_authoring/blob/master/routes/preview/error.hbs

favicon points to the wrong url

## Versions
  - Authoring Tool Version: 0.6.2
  - Framework Version: --
  - Node.js Version: --
  - MongoDB Version: --
  - Operating System: --
  - Browser: --

lc-thomasberger:
fixes #2202

### DIFF
--- a/routes/preview/error.hbs
+++ b/routes/preview/error.hbs
@@ -3,9 +3,9 @@
   <head>
     <!-- TODO needs to be localisable -->
     <title>Adapt Authoring Tool</title>
-    <link rel="icon" type="image/png" href="/adaptbuilder/css/assets/favicon.png" />
-    <link rel="stylesheet" href="/adaptbuilder/css/adapt.css" />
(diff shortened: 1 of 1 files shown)


canstudios-nicolaw:
I was expecting the favicon to be the adapt logo, is this definitely correct?

favicon

lc-thomasberger:
This PR updates the template used for course preview errors. 
image

The course preview does not include a link rel icon tag in the header as far as I can see.

canstudios-nicolaw:
Sorry I'm still confused. When a course preview errors the preview tab closes itself and then I get an error modal on the authoring tool tab, this tab already has the adapt logo. I'm not sure what this PR fixes. I can see that the paths in this template file are incorrect, but I can't see where this template file is used

lc-thomasberger:
If you try to open the preview in a different browser tab. Might by edge-case, but non the less the src is wrong so I thought it would be a good idea to fix this as it is just a minor change.

PR #2208, opened 2019-02-01
Change your own email via the profile view

### LINKED ISSUE #2021: Change your own email via the profile view

canstudios-nicolaw:
Fixes #2021

### DIFF
--- a/frontend/src/modules/user/views/userProfileView.js
+++ b/frontend/src/modules/user/views/userProfileView.js
@@ -108,6 +108,7 @@ define(function(require){
       var prevEmail = self.model.get('email');
 
       this.$('.error-text').addClass('display-none');
+      this.$('.error').text('');
 
       var toChange = {
         firstName: self.$('#firstName').val().trim(),
--- a/frontend/src/modules/user/views/userProfileView.js
+++ b/frontend/src/modules/user/views/userProfileView.js
@@ -105,12 +105,14 @@ define(function(require){
 
     saveUser: function() {
       var self = this;
+      var prevEmail = self.model.get('email');
 
       this.$('.error-text').addClass('display-none');
 
       var toChange = {
         'firstName': self.$('#firstName').val().trim(),
-        'lastName': self.$('#lastName').val().trim()
+        'lastName': self.$('#lastName').val().trim(),
+        'email': self.$('#email').val().trim()
       };
 
       if (self.model.get('_isNewPassword')) {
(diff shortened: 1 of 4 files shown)


canstudios-nicolaw:
I am closing https://github.com/adaptlearning/adapt_authoring/pull/2022 and replacing it with this. The solution was still the same, it was just easier to make a fresh branch

lc-thomasberger:
Looks good. Can you please add email validation as we have for creating a user.

canstudios-nicolaw:
@lc-thomasberger email validation appears to already be happening as part of the model's validation. Strangely, I can see the validate function in `frontend/src/modules/user/models/userProfileModel.js` but this seems to only check for the presence of an email.

invalidemail

frontend/src/modules/user/views/userProfileView.js
```diff
@@ -105,12 +105,14 @@ define(function(require){
...
     saveUser: function() {
       var self = this;
+      var prevEmail = self.model.get('email');
 
       this.$('.error-text').addClass('display-none');
 
       var toChange = {
         'firstName': self.$('#firstName').val().trim(),
-        'lastName': self.$('#lastName').val().trim()
+        'lastName': self.$('#lastName').val().trim(),
```
tomgreenfield: Suggest ditching the quotes on these keys.

lib/usermanager.js
```diff
@@ -293,14 +293,28 @@ exports = module.exports = {
...
+        if (update.email) {
+          // verify the email does not already exist
+          db.retrieve('user', { email: update.email }, function (error, results) {
+            if (error) {
+              return callback(error);
+            }
+            if (results && results.length) {
+              return callback(new DuplicateUserError());
+            } else {
+              db.update('user', search, update, callback);
```
tomgreenfield: No need for `else` when using `return`.

dancgray:
@canstudios-nicolaw, the duplicate username check does not allow the user to change first name, last name or password without also changing the email.

canstudios-nicolaw:
Good spot thanks @dancgray I'll get this updated today

lib/usermanager.js
```diff
@@ -293,14 +293,27 @@ exports = module.exports = {
...
-        }, configuration.getConfig('dbName'));
-      } else {
-        callback(new Error('No matching user record found'));
+        return callback(error);
       }
+      if (!result) {
+        return callback(new Error('No matching user record found'));
+      }
+      database.getDatabase(function(err, db) {
+        if (update.email_prev && update.email !== update.email_prev) {
```
tomgreenfield: `update.email_prev` never seems to be defined for me and so the duplicate user error is not caught:

error

taylortom:
**Found a very minor issue**: the validation messages don't seem to be removed once they're shown. 

In below screenshot, I did the following: 
1. Entered an invalid email, no first name/surname. 
2. On clicking save, I was shown validation errors as expected.
3. After correcting the email, and trying to save again, I was still shown the invalid email label (this applies to any combination of the fields).

Regardless of which messages are shown, you are able to save as expected, so it's not affecting the actual functionality.

screenshot 2019-02-19 at 13 10 54

canstudios-nicolaw:
@taylortom Validation issue should be fixed now :+1:

Issue #2187, opened 2019-01-22
Add UI for theme editing


canstudios-nicolaw:
It should be possible to edit themes using the authoring tool.

This work will be based on the prototype code started here (also see link for WIP screenshots): https://github.com/adaptlearning/adapt_authoring/pull/1499

Full proposal for the work that will be carried out can be found here: https://docs.google.com/document/d/1XmOMhdSJIZcStPlC0E-IrbEiUD_J6kwxN5sv0woH0Bc/edit#heading=h.q3ww183fpn4h

canstudios-nicolaw:
We have been thinking about the UI of this feature, I know it has been mentioned before that it wanted some improvement.

We want to support grouping of theme variables, which could then be displayed in distinct boxes like so:

![screenshot 2019-01-23 at 11 53 52](https://user-images.githubusercontent.com/25502448/51614763-421a2380-1f1e-11e9-8f5d-341c5dd23e74.png)

We are also planning to move the "Save preset" button out of the sidebar so that it is closer to the manage presets button. This could go on the page with the theme variables rather than the theme selection as this feels different to picking a theme.

![screenshot 2019-01-23 at 11 59 13](https://user-images.githubusercontent.com/25502448/51614849-6a098700-1f1e-11e9-8368-7e32dda5e5af.png)

We also propose rewording the "Restore default" button to "Restore to preset" and moving it closer to the save preset button.

Hopefully this will make the feature more intuitive to use.

If anyone has any opinions on this please let me know.

PR #2209, opened 2019-02-01
Theme editing feature

canstudios-nicolaw:
Please see feature information on https://github.com/adaptlearning/adapt_authoring/issues/2187

With these changes editable themes can expose less variables to the authoring tool. The values of these variables can be set from the theme editor page for a course. Edited less variables are saved to a file called "yyyyy.less" in the same fashion as custom less is currently saved to "zzzzz.less".

The "yyyyy.less" is included in export and these values are also applied to imported courses. If a preset has been selected then the preset will not be imported but the values will be.

If a course has an edited theme, and custom less, then the base theme styles should be applied, followed by the edited theme variables and lastly the custom less.

-------------------------------------------------------------

I have made a theme "Editable Vanilla" to be used for testing this feature. We would like to update the existing vanilla theme to make it editable (this theme is only to be used for testing purposes until the PR is out in vanilla). I've included the zip file below.

[adapt-contrib-editablevanilla.zip](https://github.com/adaptlearning/adapt_authoring/files/2822394/adapt-contrib-editablevanilla.zip)

-------------------------------------------------------------

Our QA team have had a look at this, and there are two small issues. As I am away next week I'm going to make this PR now and describe the issues I am aware of, I will fix these as soon as I am back.

1) If you select a preset on a theme, and then change the theme and come back, the preset will be selected in the dropdown when it shouldn't be (the values will be the default no preset values).
2) Selecting asset, then toggle preset results in default asset thumbnail. Clicking save after this and previewing the course will result in the old asset appearing, despite saving the themes preset.

-------------------------------------------------------------

If anybody has time to look at this while I'm away it would be much appreciated, hopefully we can get this long-awaited feature out soon!


lib/outputmanager.js
```diff
@@ -728,9 +516,194 @@ OutputPlugin.prototype.applyTheme = function(tenantId, courseId, jsonObject, des
...
+          if (theme.properties[prop].inputType === 'Asset:image') {
+            // Split the path so we can process the filename
+            var assetPathArray = savedSettings[prop].split('/');
+            // Encode the filename (removing spaces, etc.)
+            assetPathArray[assetPathArray.length - 1] = encodeURIComponent(assetPathArray[assetPathArray.length - 1]);
+            // Make the required substitution for image assets
+            savedSettings[prop] = "\"" + assetPathArray.join('/').replace('course/', '') + "\"";
+          }
+
+          modifiedProperties += '@' + props[prop] + ' : ' + savedSettings[prop] + ';\n';
```
tomgreenfield:
```suggestion
          modifiedProperties += '@' + props[prop] + ': ' + savedSettings[prop] + ';\n';
```

lib/outputmanager.js
```diff
@@ -728,9 +516,194 @@ OutputPlugin.prototype.applyTheme = function(tenantId, courseId, jsonObject, des
...
+              // No assets to process
+              seriesCallback(null);
+            }
+        });
+      });
+    }
+  ], function(error) {
+    if (error) {
+      logger.log('error', error);
+      return next(error);
```
tomgreenfield: Suggest moving this to replace line 703 – only one `return` needed.

plugins/content/config/model.schema
```diff
@@ -291,6 +291,10 @@
     },
     "themeSettings": {
       "type": "object"
+    },
+    "_themePreset": {
```
tomgreenfield: Note: needs adding to framework.

plugins/content/course/model.schema
```diff
@@ -590,6 +590,9 @@
     "themeSettings": {
       "type": "object"
     },
+    "themeVariables": {
```
tomgreenfield: Note: needs adding to framework.

plugins/content/themepreset/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+ThemePresetContent.prototype.getModelName = function() {
+  return 'themepreset';
+};
+
+ThemePresetContent.prototype.updatePreset = function(presetId, courseId, res, next) {
+  // save to config
+  app.contentmanager.retrieve('config', { _courseId: courseId }, function(err, config) {
+    if (err) return next(err);
+      var delta = config[0];
```
tomgreenfield: Please sort the indentation in this file.

plugins/output/adapt/export.js
```diff
@@ -76,6 +78,49 @@ function copyFrameworkFiles(results, filesCopied) {
   });
 }
 
+function writeThemeVariables(results, variablesWritten) {
+  self.getCourseJSON(TENANT_ID, COURSE_ID, function(error, data) {
+    if (error) {
+      return variablesWritten(error);
+    }
+    var themeVariables = data['course'][0].hasOwnProperty('themeVariables')
```
tomgreenfield: These variables do not need to be checked for existence; they can be falsies:
```js
const themeVariables = data.course[0].themeVariables;
```

plugins/output/adapt/export.js
```diff
@@ -76,6 +78,49 @@ function copyFrameworkFiles(results, filesCopied) {
...
+      ? data['course'][0].themeVariables
+      : false;
+
+    var themeName = data['config'][0].hasOwnProperty('_theme')
+      ? data['config'][0]._theme
+      : false;
+
+    const destinationFolder = path.join(EXPORT_DIR, 'src', 'theme', themeName);
+
+    if (themeVariables) {
```
tomgreenfield: Suggest reversing the logic to return early.

plugins/output/adapt/export.js
```diff
@@ -76,6 +78,49 @@ function copyFrameworkFiles(results, filesCopied) {
...
+
+        db.retrieve('themetype', {name: themeName}, {}, function(err, results) {
+          if (err || (results && results.length != 1)) {
+            return variablesWritten(err);
+          }
+
+          var theme = results[0];
+
+          self.writeThemeVariables(COURSE_ID, theme, themeVariables, destinationFolder, function(error) {
+            if (error) {
```
tomgreenfield: Don't need the anonymous function here – can simply pass in the callback `variablesWritten`.

routes/lang/en-application.json
```diff
@@ -1,4 +1,38 @@
 {
+  "app.themesettings": "Theme settings",
```
tomgreenfield: I don't believe this string is used anywhere.

routes/lang/en-application.json
```diff
@@ -1,4 +1,38 @@
...
+  "app.customisethemedesc": "You can adjust a number of variables in your chosen theme below.<br/><br/><em><b>Tip:</b> You can also save your selections here as a 'preset' for quick access later. Do this using the button in the top corner of this panel.</em>",
+  "app.themeoverwrite": "By continuing, you will overwrite the saved settings for the %{preset} preset",
+  "app.presetinputtext": "Enter a name for the new preset",
+  "app.duplicatepreseterror": "A preset with that name already exists.",
+  "app.presetedit": "Manage presets",
+  "app.presetedittitle": "Manage saved presets",
+  "app.preseteditdesc": "You can view, change name and delete the saved presets for the chosen theme here.",
+  "app.presetdeletetext": "This will delete the saved '<b>%{preset}</b>' preset. <br/><br/><b>IMPORTANT:</b> <em>This will affect any existing courses using this preset.<em>",
+  "app.presetnameplaceholder": "Enter new name",
+  "app.themecustomisation": "Theme customisation",
```
tomgreenfield: Is this string being used anywhere?

routes/lang/en-application.json
```diff
@@ -1,4 +1,38 @@
 {
+  "app.themesettings": "Theme settings",
+  "app.themepicker": "Theme Picker",
+  "app.themeeditor": "Theming",
+  "app.savepreset": "Save as preset",
+  "app.themingtitle": "Set the style for your course",
+  "app.themingdesc": "You can modify the appearance of your course on this page.",
+  "app.choosethemetitle": "Choose theme",
+  "app.choosethemedesc": "Select the base theme for your course and apply any saved 'preset' styles using the two dropdowns below.<br/><br/><em><b>IMPORTANT:</b> Changing the options here will replace any unsaved selections you've made in the form below; make sure there's nothing you want to keep!</em>",
```
tomgreenfield: Please remove all `<em>`s from these strings – we don't embed the italic version of Open Sans so we get undesirable obliques.

lib/outputmanager.js
```diff
@@ -442,29 +448,23 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
 
 OutputPlugin.prototype.applyTheme = function(tenantId, courseId, jsonObject, destinationFolder, next) {
   var self = this;
-
-  var themeAssetsFolder = path.join(destinationFolder, Constants.Folders.Assets);
-  var themeSettings = jsonObject['course'][0].hasOwnProperty('themeSettings')
-    ? jsonObject['course'][0].themeSettings
+  var themeVariables = jsonObject['course'][0].hasOwnProperty('themeVariables')
```
tomgreenfield: These variables do not need to be checked for existence; they can be falsies:
```js
var themeVariables = jsonObject.course[0].themeVariables;
```

taylortom:
Many thanks for tidying this up Nicola, it's looking really good.

A few points:
- With regards to the customisations of your editable theme, I'm not sure what should have changed on preview, so could you please give a very quick summary of what each var does? In particular, I don't seem to be seeing:
  - Tertiary
  - Quaternary 
  - Header font colour 
  - Item font color (typo in colour here)

- Can we get some styling for the preset dropdown when disabled? Currently it looks clickable
- It'd be nice if we could style the page to better match the other editor pages by separating out the nested form items into separate container 'blocks' (see screenshot below). Not sure how much work would be involved in this (as a start, I think we can remove the container style on the `.theme-customiser`, as the fieldsets have this styling already).

I can ask around to see if we have any themes with a different look & feel to vanilla that we can repurpose and release alongside this feature to give the community some variety.

screenshot 2019-02-08 at 13 53 03

canstudios-nicolaw:
Thanks for all your feedback @tomgreenfield and @taylortom :)

canstudios-nicolaw:
I'm making changes to this feature branch now, I will comment when it is ready for review.

canstudios-nicolaw:
@taylortom Regarding the editable theme, I think it would be best if it had it's own repo (which will start as a copy of vanilla) and then I can make a PR to that so that these comments don't get mixed in with the feature comments. If I do this, would we be happy to go with "adapt-contrib-editableVanilla" as the theme name?

Good point about the styling of different blocks. I will look into that.

plugins/content/config/model.schema
canstudios-nicolaw: I've added a separate issue to this milestone, otherwise this is going to get lost https://github.com/adaptlearning/adapt_authoring/issues/2220

plugins/content/course/model.schema
canstudios-nicolaw: I've added a separate issue to this milestone, otherwise this is going to get lost https://github.com/adaptlearning/adapt_authoring/issues/2220

canstudios-nicolaw:
@taylortom the more I think about styling different blocks the less certain I am. I understand your point about making the UI as consistent as possible, and I agree that is important. At the moment all of the settings are contained within the "Customise theme" block, and I think that makes it clear that the "Save as preset" button applies to them all. If I put them in their own blocks, I'm not sure if the "Customise theme" block would make sense on it's own. I could remove the blue outlines from the groups of settings? That could make it feel more like other parts of the UI?

taylortom:
@canstudios-nicolaw cool, I get your point.

One of the main reasons I mention the styling is I'd like to see us add support for the sidebar filters like we have for the other content. Think this could be particularly useful for theming (could group colour settings/paddings etc. etc.). Were we to do this, then we may have to move the save preset button as you mention. If there's no appetite for this, then your suggestion to remove the borders would be ideal.

canstudios-nicolaw:
@taylortom It's a good point. We will have a think about this internally as well. I'll be opening this up for re-review soon and we can see what the general consensus is :)

plugins/content/theme/index.js
```diff
@@ -4,23 +4,23 @@
...
+const database = require('../../../lib/database');
+const defaultOptions = require('./defaults.json');
+const fs = require('fs');
+const logger = require('../../../lib/logger');
+const origin = require('../../../');
+const path = require('path');
+const rest = require('../../../lib/rest');
+const usermanager = require('../../../lib/usermanager');
+const util = require('util');
+const _ = require('underscore');
```
tomgreenfield: Not sure that we need Underscore declared here?

plugins/content/theme/index.js
```diff
@@ -4,23 +4,23 @@
...
+const async = require('async');
+const bower = require('bower');
+const BowerPlugin = require('../bower');
+const configuration = require('../../../lib/configuration');
+const contentmanager = require('../../../lib/contentmanager');
+const ContentPlugin = contentmanager.ContentPlugin;
+const ContentTypeError = contentmanager.errors.ContentTypeError;
+const database = require('../../../lib/database');
+const defaultOptions = require('./defaults.json');
+const fs = require('fs');
```
tomgreenfield:
```suggestion
const fs = require('fs-extra');
```

lib/outputmanager.js
```diff
@@ -442,295 +448,244 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
 
 OutputPlugin.prototype.applyTheme = function(tenantId, courseId, jsonObject, destinationFolder, next) {
   var self = this;
+  var themeVariables = jsonObject.course[0].themeVariables;
+  var themeName = jsonObject['config'][0]._theme;
```
tomgreenfield: Suggest mirroring the line above:
```suggestion
  var themeName = jsonObject.config[0]._theme;
```

lib/outputmanager.js
```diff
@@ -442,295 +448,244 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
+      // Process assets
+      database.getDatabase(function(err, db) {
+        if (err) {
+          return next(err);
+        }
+
+        db.retrieve('courseasset', {_courseId: courseId, _contentType: 'theme'}, function(err, results) {
+          if (err) {
+              seriesCallback(err);
+          } else if (results) {
```
tomgreenfield: Suggest returning early if no assets to process.

lib/outputmanager.js
```diff
@@ -442,295 +448,244 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
+  var savedSettings = {};
+  var SEPARATOR = '-';
+  var themeAssetsFolder = path.join(destinationFolder, Constants.Folders.Assets);
+  async.series([
+    function(seriesCallback) {
+      // Flatten the property names to allow two levels
+      // This is in the case where an object has been used to group
+      // theme properties
+      async.eachSeries(_.keys(theme.properties), function(key, innerCallback) {
+        if (theme.properties[key].hasOwnProperty('properties')) {
```
tomgreenfield: Good candidate for caching and reversing the logic to return early?
```js
var nestedProps = theme.properties[key].properties;

if (!nestedProps) {
```

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,546 @@
...
+      this.updateSelects();
+      this.setViewToReady();
+
+      this.$el.show();
+    },
+
+    loadCollections: function() {
+      this.themes = new ThemeCollection();
+      this.listenTo(this.themes, {
+        'sync': this.onCollectionReady,
```
tomgreenfield: No need for quotes on keys here.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,546 @@
...
+        'error': this.onError
+      });
+      this.themes.fetch();
+
+      this.presets = new PresetCollection();
+      this.listenTo(this.presets, {
+        'sync': this.onCollectionReady,
```
tomgreenfield: No need for quotes on keys here.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,546 @@
...
+    },
+
+    restorePresetSettings: function(event) {
+      event && event.preventDefault();
+      var self = this;
+      Origin.Notify.confirm({
+        type: 'warning',
+        text: Origin.l10n.t('app.restorepresettext'),
+        callback: function(confirmed) {
+          if (confirmed) {
```
tomgreenfield: Suggest returning early.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,546 @@
...
+
+    updateThemeSelect: function() {
+      var select = this.$('.theme select');
+      var oldVal = select.val();
+      // remove options first
+      $('option', select).remove();
+      // add 'no presets'
+      select.append($('<option>', { value: "", disabled: 'disabled', selected: 'selected' }).text(Origin.l10n.t('app.selectinstr')));
+      // add options
+      _.each(this.themes.models, function(item, index) {
```
tomgreenfield: Not sure whether we can be bothered to update but any `_.each()` loops on arrays can use the native `.forEach()`. 

Similarily, `_.keys()` → `Object.keys()`.

We only ever used the Underscore versions in Adapt for IE8(!).

frontend/src/modules/editor/themeEditor/views/editorPresetEditView.js
```diff
@@ -0,0 +1,82 @@
...
+    },
+
+    onDeleteClicked: function(event) {
+      event && event.preventDefault();
+      var self = this;
+      var presetName = $(event.currentTarget).closest('.preset').attr('data-name');
+      Origin.Notify.confirm({
+        text: Origin.l10n.t('app.presetdeletetext', { preset: presetName }),
+        callback: function(confirmed) {
+          if (confirmed === true) {
```
tomgreenfield: Reverse logic?
```js
if (!confirmed) return;
```

frontend/src/modules/editor/themeEditor/models/editorModel.js
```diff
@@ -0,0 +1,125 @@
...
+      },
+
+      serialize: function() {
+        return JSON.stringify(this);
+      },
+
+      // Remove any attributes which are not on the whitelist
+      // Useful to call before a save()
+      pruneAttributes: function() {
+        var self = this;
```
tomgreenfield: Suggest:
```js
if (!this.whitelistAttributes) return;

// Ensure that only valid attributes are pushed back on the save
Object.keys(this.attributes).forEach(function(key) {
  if (!_.contains(this.whitelistAttributes, key)) {
    this.unset(key);
  }
}, this);
```

plugins/content/theme/index.js
canstudios-nicolaw: You're right (there were a few others as well, I've removed them)

plugins/content/theme/index.js
canstudios-nicolaw: This wasn't used so was removed

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
canstudios-nicolaw: I've changed the ones in the themeEditor directory

plugins/content/themepreset/index.js
canstudios-nicolaw: Sorted now

frontend/src/modules/editor/themeEditor/templates/editorPresetEdit.hbs
```diff
@@ -0,0 +1,30 @@
...
+        </div>
+        <div class="nameEdit">
+          <input type="text" value="{{attributes.displayName}}">
+          <a href="" class="button save btn secondary">
+            {{t 'app.themepresetsave'}}
+          </a>
+          <a href="" class="button cancel btn primary-hollow">
+            {{t 'app.themepresetcancel'}}
+          </a>
+        </div>
```
lc-thomasberger: We should not use `<a>` for buttons. Can you please update to use `<button>`.

lc-thomasberger:
Thanks for this PR and providing the test theme! I like the feature.

image

I think the "manage presets" modal is a bit to big. This also makes the alignment of the preset-items look a bit odd.
I would suggest adding a max-width in px. The items should use the whole width of the modal. This can easily done using flexbox.

The close button of the modal is not well aligned. Suggest you use position absolute instead of float right.

When you click "Edit name" it is hard to see that the name is actually editable. I think it would be good and easy to set the focus to the input once you click the "Edit name" button.

In general, I think we should try to avoid `<a href="#">` for buttons. Also we should try to use flexbox instead of floats. Flexbox is far more powerful and well supported.

taylortom:
The bad styling is probably on me, so apologies in advance for that 😅

lc-thomasberger:
I think there is an issue with updating presets.
If you update a colour of an existing preset and hit save and revisit the theme editor, everything is fine.
If you select no-preset and then the updated preset, it displays the initial settings. So the updated values are lost.

canstudios-nicolaw:
Thank you for the review @lc-thomasberger all good points about the modal styling :) 

I'm not sure that the issue with updating presets sounds like an issue to me. I would expect that if you change presets in the dropdown you should see the default values for that preset. In my mind when some settings are saved they are just saved as individual settings, they are not saved as 'this preset with these settings on top', so if I changed presets and came back I wouldn't expect them to be remembered as part of the preset. I would also imagine that if you want to change some settings on a preset and then keep them, then that should be a new preset. Alternatively, if you were to click 'Cancel' in the scenario you describe then I think that the updated values would not have been lost

@taylortom no worries :joy: I should have spotted it before PR!

lc-thomasberger:
@canstudios-nicolaw valid point. Altough I think there is a use case for updating presets. Maybe we can add a update preset button once a preset is changed? Can be a separate issue though if it requires to much refactoring.

canstudios-nicolaw:
@lc-thomasberger I don't think updating presets is MVP for this feature as it's already possible to create and delete presets. I will raise an issue and add it to this milestone for now so that it isn't forgotten - I do think it would be nice to have.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,547 @@
...
+
+    updateThemeSelect: function() {
+      var select = this.$('.theme select');
+      var oldVal = select.val();
+      // remove options first
+      $('option', select).remove();
+      // add 'no presets'
+      select.append($('<option>', { value: "", disabled: 'disabled', selected: 'selected' }).text(Origin.l10n.t('app.selectinstr')));
+      // add options
+        this.themes.models.forEach(function(item) {
```
tomgreenfield: Please fix indentation.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,547 @@
...
+      var selectedPresetId = null;
+      if (selectedPreset) selectedPresetId = selectedPreset.get('_id');
+
+      $.post('/api/themepreset/' + selectedPresetId + '/makeitso/' + this.model.get('_courseId'))
+      .error(_.bind(this.onSaveError, this))
+      .done(_.bind(callback, this));
+    },
+
+    postSettingsData: function(callback) {
+      if (this.form) {
```
tomgreenfield: Suggest reversing logic to return early.

lib/outputmanager.js
```diff
@@ -442,295 +448,248 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
+              savedSettings[key + SEPARATOR + childKey] = themeVariables[key][childKey];
+
+              secondInnerCallback();
+            }, function(err) {
+              if (err) {
+                logger.log('error', 'Theme customisations 2 of 4 -- error flattening themeVariables');
+              }
+            });
+          } else {
+              savedSettings[key] = themeVariables[key];
```
tomgreenfield: Please fix indentation.

lib/outputmanager.js
```diff
@@ -442,295 +448,248 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
-              fs.readFile(path.join(destinationFolder, Constants.Folders.Less, Constants.Filenames.Variables), function(err, file) {
-                if (err) {
-                  logger.log('error', err);
-                  return next(null, themeName);
-                }
+              var asset = assets[0],
+                  outputFilename = path.join(themeAssetsFolder, asset.filename);
+
+              // Ensure that an asset is only copied once
+              if (!processedAssets[asset.filename]) {
```
tomgreenfield: Suggest switching logic to return early.

lib/outputmanager.js
```diff
@@ -442,295 +448,248 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
   var themeAssetsFolder = path.join(destinationFolder, Constants.Folders.Assets);
-  var themeSettings = jsonObject['course'][0].hasOwnProperty('themeSettings')
-    ? jsonObject['course'][0].themeSettings
-    : false;
+  async.series([
+    function(seriesCallback) {
+      // Flatten the property names to allow two levels
+      // This is in the case where an object has been used to group
+      // theme properties
+      async.eachSeries(_.keys(theme.properties), function(key, innerCallback) {
```
tomgreenfield: Please switch to native `Object.keys()`.

lib/outputmanager.js
```diff
@@ -442,295 +448,248 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
+        if (!nestedProps) {
+          props[key] = key;
+          innerCallback();
+        }
 
-  var themeName = jsonObject['config'][0].hasOwnProperty('_theme')
-    ? jsonObject['config'][0]._theme
-    : false;
+        // There are nested properties to process
+        async.eachSeries(_.keys(nestedProps), function(childKey, secondInnerCallback) {
```
tomgreenfield: Please switch to native `Object.keys()`.

lib/outputmanager.js
```diff
@@ -442,295 +448,248 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
+        });
+      },
+      function(err) {
+        if (err) logger.log('error', 'Theme customisations 1 of 4');
+        seriesCallback(err);
+      });
+    },
+    function(seriesCallback) {
+      // Now flatten the themeVariables
+      async.eachSeries(_.keys(themeVariables), function(key, innerCallback) {
```
tomgreenfield: Please switch to native `Object.keys()`.

lib/outputmanager.js
```diff
@@ -442,295 +448,248 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
+        if (key !== '_type') {
+          if (typeof themeVariables[key] === 'object') {
+            // Iterate the properties and add them to the array
+            async.each(_.keys(themeVariables[key]), function(childKey, secondInnerCallback) {
```
tomgreenfield: Please switch to native `Object.keys()`.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,547 @@
...
+          this.postSettingsData(this.onSaveSuccess);
+        });
+      });
+    },
+
+    postThemeData: function(callback) {
+      var selectedTheme = this.getSelectedTheme();
+      var selectedThemeId = selectedTheme.get('_id');
+      $.post('/api/theme/' + selectedThemeId + '/makeitso/' + this.model.get('_courseId'))
+        .error(_.bind(this.onSaveError, this))
```
tomgreenfield: Please could you switch to native `.bind()` in this file?

frontend/src/modules/editor/themeEditor/less/editorPresetEdit.less
```diff
@@ -0,0 +1,76 @@
...
+          display: inline-block;
+          width: 50%;
+          border: 1px solid #f1f1f1;
+          border-radius: 5px;
+          font-weight: bold;
+          word-break: break-word;
+        }
+
+        .button {
+          padding-right: 10px;
```
tomgreenfield: Not sure what this styling does apart from make the text slightly off-centre?

frontend/src/modules/editor/themeEditor/less/editorPresetEdit.less
```diff
@@ -0,0 +1,76 @@
...
+    border-radius: 5px;
+    margin: 0 auto;
+    padding: 35px;
+    max-height: 90%;
+    min-width: 552px;
+    position: relative;
+
+    .button {
+        display: inline-block;
+        width: 115px;
```
tomgreenfield: Needs a `vertical-align: middle;` to stop the buttons going out of alignment.

frontend/src/modules/editor/themeEditor/less/editorPresetEdit.less
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        .label,
+        .nameEdit input {
+          padding: 11px;
+          margin-right: 15px;
+          display: inline-block;
+          width: 50%;
+          border: 1px solid #f1f1f1;
+          border-radius: 5px;
```
tomgreenfield: Needs a `vertical-align: middle;` to align with the buttons.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,546 @@
...
+      if (selectedPreset) selectedPresetId = selectedPreset.get('_id');
+
+      $.post('/api/themepreset/' + selectedPresetId + '/makeitso/' + this.model.get('_courseId'))
+      .error(this.onSaveError.bind(this))
+      .done(callback.bind(this));
+    },
+
+    postSettingsData: function(callback) {
+      if (!this.form) {
+        callback.apply(this);
```
tomgreenfield: Missing a `return`.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
tomgreenfield: Sorry, this is still off; lines 141 & 142 need to be less indented.

canstudios-nicolaw: For some reason, this looks absolutely fine in my IDE :( I'll have to have a play around with it

canstudios-nicolaw:
I have now made PRs to update the vanilla theme, please could these themes be used for testing and any comments made on the relevant PR https://github.com/adaptlearning/adapt_framework/issues/2359

canstudios-nicolaw:
@tomgreenfield Happy to move the buttons if that is the preference. I had a look around the UI of the authoring tool and I couldn't find anywhere that had save buttons in the main body of the page, I also thought that a benefit of putting them in the sidebar would mean that users wouldn't have to scroll back to the top to save the preset.

@taylortom @lc-thomasberger Please could you let me know your opinions on button placement?

I will look into tweaking the code as you suggest TomG

lc-thomasberger:
I do like the rows suggested by Tom G' Screenshot.
Manage Presets makes sense to be next to the preset dropdown.

"Save as Preset" and "Restore to Preset" is related to the variables and therefore I think it should be part of the variable group below.

image

canstudios-nicolaw:
@lc-thomasberger I don't think the restore and save buttons can be with the variables anymore. Variables could be split across several separate boxes now (eg Colours, Fonts etc) so it would be strange if these buttons only existed for the first box but applied to variables across all the boxes.

lc-thomasberger:
Then maybe we leave them in the sidebar for now ... 😄

lib/outputmanager.js
```diff
@@ -442,295 +448,248 @@ OutputPlugin.prototype.clearBuildFlag = function(pathToBuildFlag, next) {
...
+    function(seriesCallback) {
+      // Flatten the property names to allow two levels
+      // This is in the case where an object has been used to group
+      // theme properties
+      async.eachSeries(Object.keys(theme.properties), function(key, innerCallback) {
+        var nestedProps = theme.properties[key].properties;
+
+        if (!nestedProps) {
+          props[key] = key;
+          innerCallback();
```
tomgreenfield: Missing a `return`.

routes/lang/en-application.json
```diff
@@ -1,4 +1,36 @@
 {
+  "app.themepicker": "Theme Picker",
+  "app.themeeditor": "Theming",
+  "app.savepreset": "Save as preset",
+  "app.themingtitle": "Set the style for your course",
+  "app.themingdesc": "You can modify the appearance of your course on this page.",
```
tomgreenfield: Is this string is used anywhere?

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,549 @@
...
+    },
+
+    updateThemeSelect: function() {
+      var select = this.$('.theme select');
+      var oldVal = select.val();
+      // remove options first
+      $('option', select).remove();
+      // add 'no presets'
+      select.append($('<option>', { value: "", disabled: 'disabled', selected: 'selected' }).text(Origin.l10n.t('app.selectinstr')));
+        // add options
```
tomgreenfield:
```suggestion
      // add options
```

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
tomgreenfield:
```suggestion
      this.themes.models.forEach(function(item) {
```

plugins/content/config/model.schema
tomgreenfield: With this attribute I'm getting the classic error on the configuration settings page:
```console
Uncaught TypeError: constructorFn is not a constructor
    at child.createEditor (backbone-forms.js:842)
    at child.initialize (backbone-forms.js:803)
    at child.Backbone.View (backbone.js:991)
    at new child (backbone.js:1541)
    at child.createField (backbone-forms.js:136)
    at child.<anonymous> (backbone-forms.js:83)
    at Array.forEach (<anonymous>)
    at Function._.each._.forEach (underscore.js:79)
    at child.initialize (backbone-forms.js:81)
    at child.Backbone.View (backbone.js:991)
```

canstudios-nicolaw: Were you using the correct framework branch? https://github.com/adaptlearning/adapt_framework/issues/2360

tomgreenfield: Unrelated – this branch currently utilises the core schema which is where it's picking up this attribute.

I would've thought it just needs adding to the [blacklist](https://github.com/CanStudios/adapt_authoring/blob/48d005e64a57627e9beff44b2911487321b839de/lib/contentmanager.js#L561-L586).

canstudios-nicolaw:
@adaptlearning/authoring-tool-push Could I have a hand?

I have rebased the feature/theme-editing branch in the canstudios fork onto the latest master. I have also rebased release/theme-editing onto the latest master.

If I compare these branches as if I was going to make a new PR from the canstudios fork I can see the correct number of commits (just the theme editing changes): https://github.com/adaptlearning/adapt_authoring/compare/release/theme-editing...CanStudios:feature/theme-editing

However this PR is showing all the commits for 0.7.0. 

Is there something I need to do to 'refresh' this PR on github?

taylortom:
@canstudios-nicolaw just a wild stab in the dark 😅

lib/outputmanager.js
```diff
@@ -77,18 +77,17 @@ var Constants = {
...
       Download: 'download.zip',
       Main: 'index.html',
       Rebuild: '.rebuild',
-      CustomStyle: 'zzzzz.less',
+      CustomStyle: '2-customLess.less',
       Bower: 'bower.json',
       Package: 'package.json',
       Metadata: 'metadata.json',
-      ThemeVariables: 'yyyyy.less',
+      ThemeVariables: '1-themeVariables.less',
```
tomgreenfield: I don't think `Constants.Filenames.Variables` is used any longer so could switch to that instead of having `ThemeVariables`.

lib/outputmanager.js
```diff
@@ -77,18 +77,17 @@ var Constants = {
...
         Plugins: 'plugins',
-        ThemeVariablesDirectory: 'yyyyy',
         CustomStyleDirectory: 'zzzzz'
     },
     Filenames: {
       Download: 'download.zip',
       Main: 'index.html',
       Rebuild: '.rebuild',
-      CustomStyle: 'zzzzz.less',
+      CustomStyle: '2-customLess.less',
```
tomgreenfield: Suggest `2-customStyles.less`.

lib/outputmanager.js
```diff
@@ -606,12 +599,18 @@ OutputPlugin.prototype.writeThemeVariables = function(courseId, theme, themeVari
...
+        fs.removeSync(path.join(destinationFolder, Constants.Folders.Less, Constants.Folders.CustomStyleDirectory, Constants.Filenames.ThemeVariables));
+        return seriesCallback(null);
+      }
+
       fs.outputFile(destinationFile, modifiedProperties, 'utf8', function(err) {
         if (err) {
           logger.log('error', 'Theme customisations 4 of 4');
         }
-        seriesCallback(err);
+        seriesCallback(null);
```
tomgreenfield: Did you mean to change from `err` to `null`? Looks like any error would now be ignored.

tomgreenfield:
Hi @canstudios-nicolaw, this still isn't working for me I'm afraid – upon loading of the editor it looks as though the code is comparing against the default (no preset) settings rather than those of the current preset.

canstudios-nicolaw:
@tomgreenfield I think perhaps I fixed a slightly different issue then (oops). The one I fixed was that for themes with nested properties, the 'Restore to preset' button wasn't shown if I made changes from the preset, saved and then returned to the page.

So, just to confirm, the issue that you're seeing happens when:
- Select a theme with unnested properties
- Select a preset
- Make some changes on top of that preset
- Save
- Go back to the theme editor
- 'Save to Preset' and 'Restore to preset' buttons not visible

Is that correct?

(Sorry I've just switched the testing server over to bugpatch 1 for QA and I'm in the middle of something right now so don't have an easy way to look at this)

canstudios-nicolaw:
Ok brilliant thanks, I will have a look at this tomorrow.

canstudios-nicolaw:
@tomgreenfield I've fixed this for the restore to preset button. At the moment the 'Save to preset' button is always visible by design. Personally I think I'm in favour of keeping it that way. I think it makes sense that you can only restore to preset when you have made changes, but I don't think there is a need to restrict when you can create one.

Worst case scenario you don't realise you already have a preset and end up with two identical ones, but it is possible to delete duplicates.

If you feel strongly that it shouldn't always be visible I can quite easily make it appear in the same fashion as the restore button I'm not sure if it could get confusing.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,560 @@
...
+    onPresetChanged: function(event) {
+      var preset = this.presets.findWhere({ _id: $(event.currentTarget).val() });
+      var settings = preset && preset.get('properties') || this.getDefaultThemeSettings();
+      this.setPresetSelection($(event.currentTarget).val());
+      this.restoreFormSettings(settings);
+      this.updateRestorePresetButton(false);
+    },
+
+    onFieldChanged: function() {
+        this.updateRestorePresetButton(true);
```
tomgreenfield: I'm seeing that when you change a colour and then put it back to the preset's value, the restore to preset button is still showing when it shouldn't.

It should be as simple as removing the parameter from this function call but the code doesn't seem to be properly calculating the current selections?

Btw, completely agree with you re: the save button.

canstudios-nicolaw: Yes you're right, the code seems to be comparing what is currently saved against the preset. It should be comparing what is currently selected in the form against the preset.

Looking in to how to get what is currently selected in the form. If you know how to to this, or you know of somewhere else in the authoring tool that might be doing something similar, then please let me know because it doesn't seem straightforward to me at the moment.

tomgreenfield: Reverse-engineering from the way values are set in the current code, this sort of thing will work for unnested properties:
```js
getCurrentSettings: function() {
  if (!this.form) {
    return Origin.editor.data.course.get('themeVariables');
  }

  return _.mapObject(this.form.fields, function(field) {
    return field.getValue();
  });
},
```

The get functions (above and `getDefaultThemeSettings()`) should probably be modified to flatten any nested properties so that it's simpler to compare.

canstudios-nicolaw:
I have updated vanilla: https://github.com/adaptlearning/adapt-contrib-vanilla/pulls

lib/outputmanager.js
```diff
@@ -234,6 +234,11 @@ OutputPlugin.prototype.sanitizeCourseJSON = function(mode, json, next) {
       }
     }
 
+    // _themePreset should not be included in export
+    if (mode === Constants.Modes.Export) {
```
tomgreenfield: Question: why solely not in export? It's included in publishes at present.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,576 @@
...
+    className: 'theming',
+
+    settings: {
+      presetSelection: null
+    },
+
+    events: {
+      'change .theme select': 'onThemeChanged',
+      'change .preset select': 'onPresetChanged',
+      'change .form-container form': 'updateRestorePresetButton',
```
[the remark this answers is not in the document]
canstudios-nicolaw: I'm fine with putting this back into the onFieldChanged handler, please could you explain how it's DRYer though? I just figured it made more sense to have the event call the function directly rather than call a function whose only purpose is to call another function?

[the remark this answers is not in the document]
tomgreenfield: Sure. I envisioned the UI to be tweaked somehow to make it obvious we've customised our theme/preset, e.g.:
modified
I nearly requested this in a review but I didn't want to burden the workload any more at this point! With the current code, any further UI changes would require repeating the event handler:
```js
'change .form-container form': 'updateRestorePresetButton',
'change .form-container form': 'updateSelectText'
```
The other undesired side effect of not having a dedicated event handler, as you've already noticed, is that we don't have control over the arguments. The logic had to be changed [here](https://github.com/adaptlearning/adapt_authoring/pull/2209/commits/97179a923c5cda712cc5e9d3dcdd1e94db4e7cd4#diff-18eb4043de18e1156fb7293f6fa554e1R453) because the `shouldShow` parameter is now sometimes the change event.

In an ideal world, there would be no comparisons being done in the UI functions – instead we'd derive `shouldShow`/`isThemeModified` in the event handler or even in its own function so it's reusable, and then pass a toggle to one/more UI setters.

Hope I'm making sense!

lib/outputmanager.js
canstudios-nicolaw: I wish I could remember the answer, my fault for not documenting a reason at the time.. 

I will test removing it in publishes and hopefully remove from there also.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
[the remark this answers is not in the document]
canstudios-nicolaw: Ok fair point, I will put it back to how it was. Thanks for explaining.

I would have created an event handler function instead of having the same event calling multiple functions for sure but I see your point about the logic.

I like the idea of tweaking the UI in this way, happy to raise an issue for it and work on it (perhaps as part of a bug patch) but I agree we shouldn't be adding more to this PR/release at this point.

frontend/src/modules/editor/themeEditor/views/editorThemingView.js
```diff
@@ -0,0 +1,576 @@
...
+        var selectedTheme = this.getSelectedTheme();
+        if (!selectedTheme) return;
+        var parent = selectedTheme.get('theme');
+        return this.presets.findWhere({ '_id': this.model.get('_themePreset'), parentTheme: parent });
+      }
+    },
+
+    getDefaultThemeSettings: function() {
+      var defaults = {};
+      var props = this.getSelectedTheme().attributes.properties;
```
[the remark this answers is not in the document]
canstudios-nicolaw: I will look into this.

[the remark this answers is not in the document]
canstudios-nicolaw: I think within the variables object it is still going to be important to specify a title for each group of variables (colours, fonts etc.), so I think it will have to be more like this:

```
"variables": {
      "_colors": {
        "type": "object",
        "required": false,
        "title": "Colours",
        "properties": {
          "primary-color": {
            "title": "Primary",
            "help": "This is the primary colour used to style the course, and is applied to component items and pop-up notifications.",
            "type": "string",
            "inputType": "ColourPicker",
            "default": "#078292"
          }
        }
      }
    }
```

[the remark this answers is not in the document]
tomgreenfield: That's what I was going for; just cut a bit too many attributes out of my snippet! Basically the same structure at present (nested & unnested) with a parent `variables` property.

canstudios-nicolaw:
@tomgreenfield I've added support for the change to the schema as suggested. I've also updated the vanilla PRs here https://github.com/adaptlearning/adapt-contrib-vanilla/pulls

tomgreenfield:
Thanks @canstudios-nicolaw! I have passed this onto our QA team.

PR #2213, opened 2019-02-06
Scaffold fallback editors

lc-thomasberger:
This PR is not meant to be merged ... 

We can't really do anything to prevent old / existing AT users from installing plugins that introduce new scaffold editors.

To make things more robust, we could patch Backbone Forms to support editor fallbacks.
Plugins could specify a fallback editor:
```
"src": {
          "type": "string",
          "required": false,
          "default": "",
          "inputType": {
                "type": "AssetItem",  // <-- drag-and-drop UI
                "fallbackType": "Asset:image", // <-- old UI
          	"items": "_items",
          	"topAttribute": "_top",
          	"leftAttribute": "_left",
          	"precision": 2
          },
          "validators": [],
          "help": "This is the image that appears behind the pins"
}
```
If the fallback editor is also not available, we could simply use the TextEditor. This will make sure that the CourseEditor still works.

Ideally this is combined with a UI to change Plugin Versions to roll back to a previous version.

### DIFF
--- a/frontend/src/modules/scaffold/backboneFormsOverrides.js
+++ b/frontend/src/modules/scaffold/backboneFormsOverrides.js
@@ -33,6 +33,25 @@ define([
     });
   };
 
+  Backbone.Form.Field.prototype.createEditor = function() {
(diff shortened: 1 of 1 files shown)


lc-thomasberger:
image

Issue #2219, opened 2019-02-11
Add ability to unmask Password inputs


lc-thomasberger:
## Affected Area
- Create new User
- User Profile (already implemented)

taylortom:
Closing due to inactivity & impending deprecation of the legacy codebase.

PR #2227, opened 2019-02-12
update Password policy

### LINKED ISSUE #2217: Refactor password policy
Expose in configuration.
Unify validation and Feedback.

## Affected Area
- create new User
- update password by User
- forgot password
- admin password reset

## Requested Feature
As an admin I would like to define a pattern that is used to validate password strength.
As an admin I would like to define what minimum password strength is required.

I think we can build upon the approach / code from here but make it more universal and allow configuration. https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/modules/user/views/userProfileView.js#L79-L103

lc-thomasberger:
fixes #2217 and #2219 

Will update target branch to 0.7.1 once we have the release branch.


frontend/src/modules/userManagement/views/addUserView.js
```diff
@@ -5,36 +5,60 @@ define(function(require){
...
     isValid: function() {
       var valid = true;
+      var map = Object.create(null);
 
       this.$('.field-error').each(function(index, element) {
         var $error = $(element);
         var $input = $error.siblings('input');
+        var values = {
+          $error: $error,
+          value: $input.val().trim()
```
tomgreenfield: Causes the form to break since `$input` is no longer the right selector; there's now a `password-holder` `<div>` in there.

frontend/src/core/helpers.js
```diff
@@ -263,6 +263,53 @@ define(function(require){
       return value.length > 0 && regEx.test(value);
     },
 
+    validatePassword: function(password, email, firstName, lastName) {
+      debugger;
```
tomgreenfield: ⚠️

frontend/src/core/helpers.js
```diff
@@ -263,6 +263,53 @@ define(function(require){
...
+      }
+
+      if (config.notIncludeNames && email && firstName && lastName) {
+        var userNameRegex = new RegExp([email.split('@')[0],firstName, lastName].join('|'),'gi');
+        if (userNameRegex.test(password)) {
+          errors.push(Origin.l10n.t('app.validationpasswordnames'));
+        }
+      }
+
+      _.keys(config).forEach(function(key) {
```
tomgreenfield: Suggest native `Object.keys()`.

frontend/src/modules/user/models/userPasswordResetModel.js
```diff
@@ -1,7 +1,9 @@
 // LICENCE https://github.com/adaptlearning/adapt_authoring/blob/master/LICENSE
-define(function(require) {
-  var Backbone = require('backbone');
-  var Origin = require('core/origin');
+define([
+  'backbone',
+  'core/origin',
+  'core/helpers'
+], function(Backbone, Origin, Helpers) {
```
tomgreenfield: Didn't think `Backbone` was needed?

Applies throughout.

install.js
```diff
@@ -574,7 +580,9 @@ function saveConfig(configItems, callback) {
 }
 
 function handleError(error, exitCode, exitMessage) {
-  if(error) {
+  if(error && error.message) {
```
tomgreenfield: Suggest simply:
```js
if (error) {
  console.error(`ERROR: ${error.message || error}`);
}
```

routes/lang/en-application.json
```diff
@@ -146,7 +146,12 @@
   "app.lastname": "Last name",
   "app.email": "Email",
   "app.validationrequired": "A value is required here",
-  "app.validationlength": "At least %{length} characters are required",
+  "app.validationpasswordlength": "At least %{length} characters are required",
+  "app.validationpassworddigits": "Password must include at least %{length} digit(s)",
+  "app.validationpasswordlowercase": "Password must include at least %{length} lower case character(s)",
+  "app.validationpassworduppercase": "Password must include at least %{length} upper case character(s)",
+  "app.validationpasswordspecial": "Password must include at least %{length} non alpha nummeric character(s)",
```
tomgreenfield: Typo `non alpha nummeric` → `non-alphanumeric`.

Also, should we be defining the characters from the regex here?

frontend/src/modules/user/views/userProfileView.js
```diff
@@ -26,17 +23,18 @@ define(function(require){
...
     handleValidationError: function(model, error) {
       Origin.trigger('sidebar:resetButtons');
 
-      if (error && _.keys(error).length !== 0) {
-        _.each(error, function(value, key) {
-          this.$('#' + key + 'Error').text(value);
-        }, this);
-        this.$('.error-text').removeClass('display-none');
-      }
+      if (!error || _.keys(error).length === 0) return;
```
tomgreenfield:
```suggestion
      if (!error || !Object.keys(error).length) return;
```
Also suggest changing following `_.each()` to native `.forEach()`.

lib/configuration.js
```diff
@@ -119,6 +121,14 @@ Configuration.prototype.load = function (filePath, callback) {
           return callback(error)
         }
 
+        config.conf.passwordPolicy = Object.assign({
+          length: 8,
+          digits: 1,
+          lowercaseLetters: 1,
+          uppercaseLetters: 1,
+          nonAlphaNumeric: 0
+        }, config.conf.passwordPolicy);
```
tomgreenfield: If these are defaults going forward, I think there needs to be an overall config you can set to override for dev/test installs:
```js
usePasswordPolicy: false
```

lc-thomasberger: Not sure about that. You could simply set every value to 0 to have the same effect.

lc-thomasberger:
Close this one and fix the password policy in the server rewrite.

PR #2232, opened 2019-02-15
Call next on error

### LINKED ISSUE #2113: Helpers.hasCoursePermission returns without calling callback on error
See the following: https://github.com/adaptlearning/adapt_authoring/blob/7f1a23f72884a710c38e17d9d21d7c3098e3b36b/lib/helpers.js#L150-L153

lc-thomasberger:
fixes #2113


taylortom:
Do we need this, if https://github.com/adaptlearning/adapt_authoring/issues/2124 is going into 0.7.0?

lc-thomasberger:
Yes I think so. As #2124 is already merged and there was one return without calling next.

Issue #2243, opened 2019-02-25
Sidebar filters need localising


tomgreenfield:
![filter](https://user-images.githubusercontent.com/922987/53335852-84a39700-38f4-11e9-9e79-4f5039acf820.png)

Namely, _Completioncriteria_, _Screensize_ & _Scrollingcontainer_ on the configuration settings page.

PR #2258, opened 2019-02-28
Fix language strings

### LINKED ISSUE #2218: Plugin management text is hardcoded
plug
https://github.com/adaptlearning/adapt_authoring/blob/3f060704f7a94bf3d19f1f69f26c60cdfc5e2fa8/frontend/src/modules/pluginManagement/index.js#L30
Needs to be switched to `app.pluginmanagement`.

tomgreenfield:
Resolves #2218 & #2243.

### DIFF
--- a/frontend/src/core/helpers.js
+++ b/frontend/src/core/helpers.js
@@ -44,7 +44,8 @@ define(function(require){
         return Origin.l10n.t(l10nKey);
       }
       // fall-back: remove all _ and capitalise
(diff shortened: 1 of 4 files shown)


taylortom:
This need to go into a release branch

canstudios-nicolaw:
Yes, would it be worth making a release/bug-patch-1 branch now as there are already a handful of PRs?

PR #2261, opened 2019-03-05
Change absolute url to relative; fix for subfolder reverse proxy.

### LINKED ISSUE #2137: App only runs correctly when served from document root
I have been trying serve adapt via nginx with url like `www.xxx.com/adapt`. Is there any configuration available to set base url path for adapt?

My configuration:
```
location /adapt/  {
  proxy_pass http://127.0.0.1:5000/;
}
```

oscarsiles:
I have made some changes to change all absolute paths to relative ones, which allows the authoring tool to be used on any subfolder rather than the server root. Fixes #2137.

I have tested it myself in both environments (root and subfolder) and all tasks appear to work correctly, but I would be happy for others to test it as well to make sure there are no issues.


oscarsiles:
I'm not quite sure why CI build is failing, but would love if someone has some time to find the issue!

oscarsiles:
It seems to be, but I cannot find anywhere to access that link? From my understanding it should navigate to the URL on course creation, but it appears to already work. I will remove the leading slash for consistency.

taylortom:
Thanks for this @oscarsiles, we're hoping to get this merged in the next patch release.

oscarsiles:
Just a quick update, it seems no Shared Courses are displaying on my install with this fix (Server Error 500). I will hopefully have another fix soon (if it hasn't been flagged already).

canstudios-nicolaw:
Hi @oscarsiles thanks for the update. Are you able to share the details of the 500 error?

We are hoping to release soon so need to make a decision as to whether this fix goes into the imminent release or the one after. 

Thanks

oscarsiles:
The error pops up when you click on "Shared Courses" and the course list is blank. From Chrome console:

`GET https://XXX/adapt/api/shared/course?search%5Btitle%5D=.*.*&operators%5Bskip%5D=0&operators%5Blimit%5D=1&operators%5Bsort%5D%5Btitle%5D=1 500 (Internal Server Error) - origin.js?20190508090112845:213`

Checking the URL, it returns "ContentPermissionError".

canstudios-nicolaw:
Thank you for your quick response.

@tomgreenfield do you have somewhere you are able to confirm this?

tomgreenfield:
Hi, I'm unable to replicate this on my local (root) instance with the latest _release/bugpatch_ branch.

oscarsiles:
I've just tested it on a root URL, and it works fine there. I have probably missed something that was added in 0.7.x...

PR #2269, opened 2019-03-06
Fix Custom LESS not working in authoring tools running framework v4

### LINKED ISSUE #2240: Custom LESS doesn't work in authoring tools which are running framework version 4
## Expected Behaviour
I should be able to add custom LESS to my course and have it applied.

## Actual Behaviour
I can add custom LESS but it is not applied to the course.

Framework 4 comes with Vanilla 4. In Vanilla 4 the LESS is organised into directories rather than a list of files. The authoring tool inserts zzzzz.less alongside the directories but it is applied before the styles contained in the directories.

screenshot from 2019-02-21 11 16 28

## Versions
  - Authoring Tool Version: 0.6.3+
  - Framework Version: 4

canstudios-nicolaw:


canstudios-nicolaw:
(I need to change the destination from master to release/theme-editing but waiting until 0.7.0 is released and everything can be rebased)

canstudios-nicolaw:
Merging so this can be tested with the theme-editing feature

PR #2277, opened 2019-03-07
Fix behaviour when deleting user with courses

### LINKED ISSUE #2270: Deleting all unshared courses when deleting a user should not delete partially shared courses
## Expected Behaviour
When an admin deletes a user, they have an option to "Delete all unshared courses". This should delete all of the courses that the user hasn't shared.

## Actual Behaviour
After deleting a user in this way, only courses with "Share with all users" enabled remain. Unshared, and partially shared courses will be deleted.

Partially shared courses should not be deleted

## Steps to Reproduce
1. As user A, Create a course and share it with a specific user
1. As a different user, delete user A
1. Note that the partially shared course has been deleted.

## Versions
  - Authoring Tool Version: 0.7.0

## Proposed Solution
When an admin deletes a user and they select "Delete all unshared courses" then only courses which have not been shared will be deleted. Shared with everyone, and partially shared courses will be preserved.

### LINKED ISSUE #2272: Course ownership should be transferred to me if I delete a user
## Expected Behaviour
It should not be possible to get orphaned courses in the system.

## Actual Behaviour
If a user is deleted with the "Share with all" option selected, then their courses get shared with everyone. An admin can then go to project setings and uncheck the "Share with all" option. At this point, the course still exists but nobody can see it.

## Versions
  - Authoring Tool Version: 0.7.0

## Proposed Solution
If "share all courses" or "delete all unshared courses" is checked when deleting a user then ownership is transferred to me. This will mean that all 3 options involve transferring ownership to me.

If a course was already partially shared with me before I deleted the user then it will no longer be partially shared with me. (Note, this could mean that it goes from being a partially shared course to an unshared course if it was only shared with me in the first place).

canstudios-nicolaw:
Fix #2270 & Fix #2272

This PR contains fixes to the issues above. The main points are that ownership of a user's courses is always transferred to the delete-r, and partially shared courses now behave the same way as shared courses.

I've attached some (fun) matrices below to show how this used to behave vs how it should behave now.

Old:
screenshot from 2019-03-07 16 24 05

New:
screenshot from 2019-03-07 16 25 54

General logic flow:
- Get all courses which were created by the user who is being deleted
- Transfer all of them to delete-r
- Handle Delete/Share options
- At the end of share and transfer options, remove delete-r from the partially shared list (if applicable)

I am still having one issue which I could use a little help with.

Lines 366 & 370 should remove the delete-r from the partially shared list but this doesn't happen. Just above those lines, if I logger.log out `course._shareWithUsers` it is undefined, however `course._shareWithUsers.length` returns a number and `typeof course._shareWithUsers` is object. I'm having a hard time getting the data in that property in order to modify it. `JSON.stringify` returns the data I'm expecting to see. I'm going to look at a couple of the other 0.7.0 issues and then return to this but if anyone has any ideas in the meantime please let me know. Most of the behaviour from the matrix above is working it's just the delete-r doesn't get removed from the _shareWithUsers list if they are on it.


canstudios-nicolaw:
This is now ready for review

lib/usermanager.js
```diff
@@ -340,36 +340,36 @@ exports = module.exports = {
...
-        default:
-          done(null);
-      }
+    if (courseOption === null) {
+      return done(null);
+    }
+    courseContent.retrieve({ createdBy: userId }, {}, function (err, results) {
+      var currentUser = this.getCurrentUser();
+      async.eachSeries(results, function(model, cb) {
+        // Transfer ownership of all courses to current user
```
tomgreenfield: Suggest adding a comment linking to this PR so we have a reference to your user deletion matrix.

lib/usermanager.js
```diff
@@ -340,36 +340,36 @@ exports = module.exports = {
...
+        courseContent.update({ _id: model._id }, { createdBy: currentUser._id }, function (err, course) {
+          if (!course._isShared && course._shareWithUsers.length === 0) {
```
tomgreenfield: Is it possible to reverse the logic here to cut out the nested `if`s?

lib/usermanager.js
```diff
@@ -340,36 +340,36 @@ exports = module.exports = {
...
+              courseContent.destroy({ _id: course._id }, false, cb);
+              return;
+            }
+            // Share private courses with everyone if this option was set
+            if (courseOption === 'share') {
+              courseContent.update({ _id: course._id }, { _isShared: true, _shareWithUsers: course._shareWithUsers.pull(currentUser._id) }, cb);
+              return;
+            }
+          }
+          courseContent.update({ _id: course._id }, { _shareWithUsers: course._shareWithUsers.pull(currentUser._id) }, cb);
```
tomgreenfield: Suggest adding a comment here, as you've done for the statements, to make it crystal clear what is happening.

lib/usermanager.js
canstudios-nicolaw: I tried changing this to the following. The share and delete options were fine, but this caused the transfer option to break. Transfer ownership was only working for private courses. 

I'm not sure how much you want to avoid nested ifs, if it was just supposed to be a trivial change then I suggest we leave it as it is, if it's a problem I'm happy to have another stab at it.

```
handleUserCoursesOnDelete: function(userId, courseOption, done) {
    var pluginManager = pluginmanager.getManager();
    var plugins = pluginManager.getPlugins();
    var CourseContent = require(plugins.content.course.fullPath);
    var courseContent = new CourseContent();
    if (courseOption === null) {
      return done(null);
    }
    courseContent.retrieve({ createdBy: userId }, {}, function (err, results) {
      var currentUser = this.getCurrentUser();
      async.eachSeries(results, function(model, cb) {
        // For a full description of behaviour see https://github.com/adaptlearning/adapt_authoring/pull/2277
        // Transfer ownership of all courses to current user
        courseContent.update({ _id: model._id }, { createdBy: currentUser._id }, function (err, course) {
          if (course._isShared || course._shareWithUsers.length > 0) {
            // Remove current user from the 'shared with' list if applicable
            return courseContent.update({ _id: course._id }, { _shareWithUsers: course._shareWithUsers.pull(currentUser._id) }, cb);
          }
          // Delete private courses if this option was set
          if (courseOption === 'delete') {
            return courseContent.destroy({ _id: course._id }, false, cb);
          }
          // Share private courses with everyone if this option was set, remove current user from the 'shared with' list if applicable
          if (courseOption === 'share') {
            return courseContent.update({ _id: course._id }, { _isShared: true, _shareWithUsers: course._shareWithUsers.pull(currentUser._id) }, cb);
          }
        }.bind(this));
      }.bind(this), done);
    }.bind(this));
  },
```

tomgreenfield: In this case it was less about the nested `if`s and more about confirming in exactly which scenarios the code falls through to line 372. Trying to refactor this sort of highlights the issue.

I'll try and have a stab at this myself.

lib/usermanager.js
```diff
@@ -340,36 +340,38 @@ exports = module.exports = {
...
-          break;
-        case 'share':
-          async.each(results, function(model, cb) {
-            courseContent.update({ _id: model._id }, { _isShared: true }, cb);
-          }, done);
-          break;
-        default:
-          done(null);
-      }
+    if (courseOption === null) {
```
tomgreenfield: Should this be at the top of the function?

lib/usermanager.js
tomgreenfield: I've got this so far, with the return of Thomas' switch statement:
```js
const currentUserId = this.getCurrentUser()._id;
courseContent.retrieve({ createdBy: userId }, {}, (err, results) => {
  async.eachSeries(results, (model, cb) => {
    // For a full description of behaviour see https://github.com/adaptlearning/adapt_authoring/pull/2277
    switch (courseOption) {
      case 'share':
        // Share private courses with everyone if this option was set, remove the current user from the 'shared with' list if applicable
        courseContent.update({ _id: model._id }, { createdBy: currentUserId, _isShared: true, _shareWithUsers: model._shareWithUsers.pull(currentUserId) }, cb);
        break;
      case 'delete':
        // Delete private courses if this option was set
        if (!model._isShared && !model._shareWithUsers.length) {
          courseContent.destroy({ _id: model._id }, false, cb);
          break;
        }
      case 'transfer':
        // Remove the current user from the 'shared with' list if applicable
        courseContent.update({ _id: model._id }, { createdBy: currentUserId, _shareWithUsers: model._shareWithUsers.pull(currentUserId) }, cb);
    }
  }, done);
});
```
💀 **NOTE**: whichever approach we go for, the code needs updating to to handle an undefined `_shareWithUsers`; this property isn't present in the existing versions of the framework and currently errors.

tomgreenfield: How's about this one then? Has an early return, handles frameworks without `_shareWithUsers` and I've tried to preserve your comments:
```js
const currentUserId = this.getCurrentUser()._id;
courseContent.retrieve({ createdBy: userId }, {}, (err, results) => {
  async.eachSeries(results, (model, cb) => {
    // For a full description of behaviour see https://github.com/adaptlearning/adapt_authoring/pull/2277
    const sharedWithList = model._shareWithUsers;
    const isPartiallyShared = sharedWithList && sharedWithList.length;
    // Delete private courses if this option was set
    if (courseOption === 'delete' && !model._isShared && !isPartiallyShared) {
      courseContent.destroy({ _id: model._id }, false, cb);
      return;
    }
    // Transfer ownership of all courses to current user
    let update = { createdBy: currentUserId };
    // Remove the current user from the 'shared with' list if applicable
    if (isPartiallyShared) {
      update._shareWithUsers = sharedWithList.pull(currentUserId);
    }
    // Share private courses with everyone if this option was set
    if (courseOption === 'share') {
      update._isShared = true;
    }
    courseContent.update({ _id: model._id }, update, cb);
  }, done);
});
```

canstudios-nicolaw: That is looking really good to me (can't test right now though). Seems much more elegant.

canstudios-nicolaw:
@tomgreenfield I've just tested this and it appears to be working well now, thanks for your help.

PR #2278, opened 2019-03-08
Give partially and fully shared courses the same deletion behaviour

### LINKED ISSUE #2266: Roles with lesser permissions cannot see any shared courses
## Expected Behaviour

All roles can see courses shared with everyone.

## Actual Behaviour

Roles other than `Super Admin` and `Tenant Admin` cannot see any shared courses.

Front end error received:
```console
jquery.js:9659 GET http://localhost:5000/api/shared/course?search%5Btitle%5D=.*.*&operators%5Bskip%5D=0&operators%5Blimit%5D=1&operators%5Bsort%5D%5Btitle%5D=1 500 (Internal Server Error)
```

## Steps to Reproduce
1. Share a course with all users
1. Log in as an authenticated user and navigate to the shared courses page

## Versions
  - Authoring Tool Version: 0.7.0
  - Node.js Version: 10
  - MongoDB Version: 3.2

### LINKED ISSUE #2271: Course delete permissions are different between shared with everyone and partially shared
## Expected Behaviour
Permissions should be consistent across both types of sharing.

## Actual Behaviour
Users that have been selected to view the course can also delete the course, whereas a course that has been shared with all can only be deleted by the course owner.

## Versions
  - Authoring Tool Version: 0.7.0

## Proposed Solution
If a course has been shared with everyone then everyone should be able to delete that course.

canstudios-nicolaw:
Fixes #2271
Fixes #2266

### DIFF
--- a/frontend/src/modules/projects/views/projectView.js
+++ b/frontend/src/modules/projects/views/projectView.js
@@ -75,8 +75,7 @@ define(function(require) {
 
     deleteProjectPrompt: function(event) {
       event && event.preventDefault();
-      var isAdmin = Origin.permissions.hasPermissions(['*/*:create','*/*:read','*/*:update','*/*:delete']);
-      var isShared = this.model.get('_isShared') === true;
+      var isShared = this.model.get('_isShared') || (this.model.get('_shareWithUsers') && this.model.get('_shareWithUsers').length > 0);
       var titleKey = isShared ? 'deletesharedproject' : 'deleteproject';
       var textKey = isShared ? 'confirmdeletesharedprojectwarning' : 'confirmdeleteprojectwarning';
 
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -152,11 +152,11 @@ function hasCoursePermission (action, userId, tenantId, contentItem, next) {
         if(course.createdBy.toString() === userId) { // user's own course
           return next(null, true);
         }
-        if(course._shareWithUsers && course._shareWithUsers.length) { // check if userId is in _shareWithUsers
-          return next(null, course._shareWithUsers.map(user => user.toString()).includes(userId));
-        }
         if(course._isShared) { // shared with _all_ users
           return next(null, true);
+        }
+        if(course._shareWithUsers && course._shareWithUsers.length) { // check if userId is in _shareWithUsers
+          return next(null, course._shareWithUsers.map(user => user.toString()).includes(userId));
         }
         next();
       });
--- a/lib/role/CourseCreator.json
+++ b/lib/role/CourseCreator.json
@@ -2,7 +2,7 @@
   "name": "Course Creator",
   "version": "0.0.1",
   "statement" : [
-    { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/content/*", "action":["create"], "effect":"allow" },
+    { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/content/*", "action":["create", "delete"], "effect":"allow" },
     { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/shared/*", "action":["create","read"], "effect":"allow" },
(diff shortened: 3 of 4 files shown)


canstudios-nicolaw:
Hm it's a good point, probably not... I suppose before the only reason that they couldn't delete courses was because they couldn't have created them.

taylortom:
Regarding authenticated users, they should only have read access to courses (see the [role definition](https://github.com/adaptlearning/adapt_authoring/blob/master/lib/role/AuthenticatedUser.json) for more info).

canstudios-nicolaw:
@tomgreenfield @taylortom 

I've spent some time revisting the logic that existed before this PR:

```
if(!canDeleteAll && docs[0]._isShared && docs[0].createdBy != user._id) {
```

To clarify, the value of "canDeleteAll" for different user types is:
Authenticated User - false
Course Creator - false
Product Manager - false
Tenant Admin - true
Super Admin - true

The original logic would have prevented AU/PM/CC users from deleting shared courses which did not belong to them. It is not possible for an AU to have created a course and so this would have always prevented them from deleting courses.

Now we want to change the logic so that anyone who can create a course can also delete it if it is shared.

I have been playing around with something along the lines of:

```
var resource = permissions.buildResourceString(tenantId, '/api/content/course/*');
      permissions.hasPermission(user._id, 'delete', resource, function(error, canDeleteAll) {
        permissions.hasPermission(user._id, 'create', resource, function(error, canCreateAll) {
            const sharedWithList = docs[0]._shareWithUsers;
            const isPartiallyShared = sharedWithList && sharedWithList.length;
            // Prevent Authenticated Users & Product Managers from deleting anything
            // Prevent Course Creators and Admins from deleting private courses which are not owned by them
            if(!canDeleteAll && !canCreateAll ||
              canCreateAll && (!docs[0]._isShared || !isPartiallyShared) && docs[0].createdBy !== user._id) {
                return next(new ContentPermissionError());
            }
            // Courses use cascading delete
            async.eachSeries(docs, function(doc, cb) {
                self.destroyChildren(doc._id, '_courseId', cb);
            }, function (err) {
                ContentPlugin.prototype.destroy.call(self, search, true, next);
            });
        });
      });
```

Where canCreateAll has the following values:
Authenticated User - false
Course Creator - true
Product Manager - false
Tenant Admin - true
Super Admin - true

I haven't tested this at all yet just posting my progress in case anyone else has any other thoughts. I will try and have another look at this on monday.

canstudios-nicolaw:
Actually that doesn't take into account that admins should be able to delete all courses..

So I guess it would need to be something more like:
```
var resource = permissions.buildResourceString(tenantId, '/api/content/course/*');
      permissions.hasPermission(user._id, 'delete', resource, function(error, canDeleteAll) {
        permissions.hasPermission(user._id, 'create', resource, function(error, canCreateAll) {
            const sharedWithList = docs[0]._shareWithUsers;
            const isPartiallyShared = sharedWithList && sharedWithList.length;
            // Prevent Authenticated Users & Product Managers from deleting anything
            // Prevent Course Creators from deleting private courses which are not owned by them
            if(!canDeleteAll && !canCreateAll ||
              !canDeleteAll && canCreateAll && (!docs[0]._isShared || !isPartiallyShared) && docs[0].createdBy !== user._id) {
                return next(new ContentPermissionError());
            }
            // Courses use cascading delete
            async.eachSeries(docs, function(doc, cb) {
                self.destroyChildren(doc._id, '_courseId', cb);
            }, function (err) {
                ContentPlugin.prototype.destroy.call(self, search, true, next);
            });
        });
      });
```

canstudios-nicolaw:
Discussed in stand up - could we possibly add delete permission to the course creator role so that canDeleteAll = true for course creators. Then revert this logic back to what it was before this PR?

canstudios-nicolaw:
With these changes SA/TA/CC users can delete shared courses, AU/PM users cannot.

PR #2291, opened 2019-03-26
Add a sort to ensure reverse-chron order

### LINKED ISSUE #2290: Automatic upgrade check broken
The code assumes the GitHub API returns releases in reverse-chronological order, which is not always the case (first noticed in v0.7.0). These need to be sorted first.

Users are therefore forced to enter a specific tag to upgrade until this is fixed.

***

**Update:** the GitHub API sorts the release data by an arbitrary ID value, so this issue is only replicable when you've created releases out of order. In our case: created `v0.7.0` draft, created and released `v0.6.4` & `v0.6.5` releases, released `v0.7.0`. As `v0.7.0` was created before the `v0.6.*`, it appears earlier.

taylortom:
Fixes #2290

### DIFF
--- a/lib/installHelpers.js
+++ b/lib/installHelpers.js
@@ -363,7 +363,12 @@ function checkLatestAdaptRepoVersion(repoName, versionLimit, callback) {
     }
     nextPage = parseLinkHeader(response.headers.link).next;
     try {
-      var releases = JSON.parse(body);
+      // parse and sort releases (newest first)
+      var releases = JSON.parse(body).sort((a, b) => {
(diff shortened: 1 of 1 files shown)


canstudios-nicolaw:
How do we go about testing this?

I was expecting the automatic upgrade to say I was up to date on a tool that was on 0.6.5 but it has picked up 0.7.0 fine.

taylortom:
Sorry yeah, to avoid any user-facing issues, I deleted and recreated the 0.7.0 release

PR #2293, opened 2019-03-26
Amend course preview to display the current page

### LINKED ISSUE #745: Preview shows the current page
If you are in a page when you click on Preview it could preview that page in the browser rather than always going to the menu, if possible. This would be another great time-saver for the author.

taylortom:
Fixes #745


tomgreenfield:
May I just re-gauge the thirst for this 'fix'?

Personally I'm quite used to the preview button being a refresh/pointer to the start of the course. Not sure of the knock-on effect with role-selector-type plugins that depend on upfront selections.

taylortom:
I couldn't see any plugins in the browser, so would stick my neck out and say it's probably not something the majority of community users will come across.

Happy to set it via `config.json` (indifferent to whether it should be enabled by default).

canstudios-nicolaw:
I did ask our content team how they would feel about this before I approved. They said that they felt it would be useful.

PR #2297, opened 2019-04-02
Fix dashboard infinite scrolling/pagination

### LINKED ISSUE #2253: Dashboard paging broken after deleting courses
## Expected Behaviour
On deleting a course, we need to do a fresh fetch to make sure the full amount of courses are shown. 

## Actual Behaviour
It's possible to break the paging by deleting enough courses to hide the scrollbar.

## Steps to Reproduce
1. Make sure you have enough courses to cause paging
1. Resize browser to fit a small number of courses (I had two rows of three)
1. Delete courses until you're down to a single row
1. Note no more courses are fetched, page must be refreshed to fix

taylortom:
Fixes #2253

### DIFF
--- a/frontend/src/modules/projects/views/projectView.js
+++ b/frontend/src/modules/projects/views/projectView.js
@@ -84,7 +84,7 @@ define(function(require) {
         title: Origin.l10n.t('app.' + titleKey),
         text: Origin.l10n.t('app.confirmdeleteproject') + '<br/><br/>' + Origin.l10n.t('app.' + textKey),
         destructive: isShared,
-        callback: _.bind(this.deleteProjectConfirm, this)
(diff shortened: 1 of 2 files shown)


frontend/src/modules/projects/views/projectView.js
```diff
@@ -97,7 +97,10 @@ define(function(require) {
 
     deleteProject: function(event) {
       this.model.destroy({
-        success: _.bind(this.remove, this),
+        success: _.bind(function() {
+          Origin.trigger('dashboard:refresh');
+          this.remove();
+        }, this),
```
canstudios-nicolaw: Please can you switch to native bind?

taylortom: 👍 have also updated all other uses of `_.bind` in the file.

PR #2346, opened 2019-05-23
Refactor import createCourseAssets fixes #2326

### LINKED ISSUE #2326: Import times out with missing asset
## Expected Behaviour
If the import is missing an asset referenced in the JSON it should either fail and report an error or finish importing but without mapping the missing asset.

## Actual Behaviour
Import times out with partially imported course.

## Steps to Reproduce
1. Create a course with one graphic component where one of the assets referenced in the component JSON is missing from the import (and is not already in the AT).
2. Import the zipped course

Although this is the result of bad data the AT should at least report an error or complete the import. In fact the createCourseAssets function never returns the callback for the missing asset. If there is no matching asset in the metaData.assetNameMap[ lines 631-655](https://github.com/adaptlearning/adapt_authoring/blob/master/plugins/output/adapt/importsource.js#L631-L655) never call the callback.

## Versions
  - Framework Version: n/a
  - Node.js Version: n/a
  - MongoDB Version: n/a
  - Operating System:
  - Browser:

dancgray:
Refactor part of the import createCourseAssets function to always call the callback. Steps to reproduce the issue can be found in #2326.

Note: have used release/bugpatch branch as base but this should be changed once the next RC branch is known.

### DIFF
--- a/plugins/output/adapt/importsource.js
+++ b/plugins/output/adapt/importsource.js
@@ -628,7 +628,7 @@ function ImportSource(req, done) {
       }
       var assetBaseName = path.basename(data);
       // get asset _id from lookup of the key of metadata.assetNameMap mapped to assetBaseName
-      var matchingAssetId = _.findKey(metadata.assetNameMap, (value, assetId) => { return (value == assetBaseName) ? assetId : false });
+      var matchingAssetId = _.findKey(metadata.assetNameMap, value => value === assetBaseName);
 
       if (!matchingAssetId) return callback();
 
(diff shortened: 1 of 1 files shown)


plugins/output/adapt/importsource.js
```diff
@@ -628,28 +628,27 @@ function ImportSource(req, done) {
       }
       var assetBaseName = path.basename(data);
       // get asset _id from lookup of the key of metadata.assetNameMap mapped to assetBaseName
-      _.findKey(metadata.assetNameMap, function(value, assetId) {
-        if (value !== assetBaseName) {
-          return;
+      var matchingAssetId = _.findKey(metadata.assetNameMap, (value, assetId) => { return (value == assetBaseName) ? assetId : false });
```
tomgreenfield:
```suggestion
      var matchingAssetId = _.findKey(metadata.assetNameMap, value => value === assetBaseName);
```

canstudios-nicolaw:
@tomgreenfield Did you want to see TomT's feedback specifically or is this good to merge now?

Issue #2348, opened 2019-05-24
Modal editor asset selection can cause orphaned courseassets


dancgray:
## Expected Behaviour
courseassets collection for a given course should accurately reflect the assets used in the course,contentobject etc JSON.

## Actual Behaviour
Selecting an asset in the item modal widow and then canceling leaves the courseasset record resulting in unwanted assets being included in publish and export zip.

## Steps to Reproduce
1. Add an accordion component to a course
2. add an item and select an asset
3. courseasset record is created in DB
4. In the item modal window click cancel
5. courseasset record is not removed from DB even though the accordion component has not selected asset. 

## Versions
  - Authoring Tool Version: v0.8.1
  - Framework Version: v2.5
  - Node.js Version: n/a
  - MongoDB Version: n/a
  - Operating System: n/a
  - Browser: n/a

Issue #2347, opened 2019-05-24
Item copy can result in broken courseassets in other items


dancgray:
## Expected Behaviour
Copying an item with pre-populated assets and replacing the assets in the copied item should not affect the original item's assets.

## Actual Behaviour
On removal of the asset in the copied item the original items asset is removed.

## Steps to Reproduce
1. add an accordion and create an item selecting a graphic 
2. copy that item and remove the asset in the copied item
3. original item's asset courseasset record is removed

## Versions
  - Authoring Tool Version: v0.8.1
  - Framework Version: v2.4
  - Node.js Version: v10.15.3
  - MongoDB Version: v3.6
  - Operating System: n/a
  - Browser: n/a

dancgray:
Also see #2348 . There are inconsistencies in the way course assets are created or deleted.  There is no courseasset creation on copy/paste item and no courseasset deletion on item deletion.

@tomgreenfield @taylortom @canstudios-nicolaw do you think we should just "patch" this in scaffoldListView.js or fix in a more fundamental way?

Issue #2345, opened 2019-05-22
Image thumbnails not showing in scaffold


dancgray:
## Expected Behaviour
Image thumbnail should show in component settings editor

## Actual Behaviour
Place holder image icon is always shown.

## Steps to Reproduce
1. Upload an image and check that the thumbnail has been made and is viewable in asset manager
2. Add a graphic component to a course using previous image
3. Placeholder icon will be shown e.g.

![Screenshot 2019-05-22 at 15 24 39](https://user-images.githubusercontent.com/6452430/58182692-03b32c00-7ca6-11e9-9784-d9661ec4192e.png)

[This](https://github.com/adaptlearning/adapt_authoring/commit/8c6540bc056d89f3140baf3c5a7afc805e903416) commit means that fetchCourseAsset uses course ID instead of block ID for _contentTypeParentId search criteria [here](https://github.com/adaptlearning/adapt_authoring/blob/release/bugpatch/frontend/src/modules/scaffold/views/scaffoldAssetView.js#L100).

## Versions
  - Authoring Tool Version: v0.8.0
  - Framework Version: v2.4.0
  - Node.js Version: 10.15.3
  - MongoDB Version: 3.6
  - Operating System: macOS 10.14.5
  - Browser: Chrome

dancgray:
Possible solution: change [searchCriteria._contentTypeParentId](https://github.com/adaptlearning/adapt_authoring/blob/release/bugpatch/frontend/src/modules/scaffold/views/scaffoldAssetView.js#L100) to searchCriteria._courseId ?

Issue #2336, opened 2019-05-17
Course preview image broken in settings


canstudios-simonasd:
## Expected Behaviour
After selecting preview image for new course, buttons 'Select an Asset' and 'Select an External Asset' should be hidden. Instead of them, image thumbnail should be shown.

## Actual Behaviour
After selecting image for new course, buttons are not hidden, and no image thumbnail is shown.

## Steps to Reproduce
1. From dashboard click 'Add new course'
1. Click 'Select an Asset'
1. Select any image and click 'Done'
1. *Notice* that form still shows 'Select an Asset' and 'Select an External Asset' buttons
1. Click 'Save'
1. Go back to dashboard
1. *Notice* that selected image is actually saved as preview image
1. Go to existing course -> Project settings
1. *Notice* it does not show correct thumbnail

## Versions
  - Authoring Tool Version: 0.8.1
  - Framework Version: 2.4.0
  - Browser: tested on Chrome and Firefox

PR #2349, opened 2019-05-24
Update scaffold fetchCourseAsset search criteria fallback to courseId…

dancgray:
Very minor fix for #2345. How scaffold fetches and saves courseassets may change given the other problems with courseassets collection, #2348 , #2336 , #2347 .


canstudios-nicolaw:
@dancgray I think this needs to go to the `release/bugpatch` branch rather than `master`

PR #2355, opened 2019-06-03
Correct alignment for "Failed logins" and "Last access" headers

### LINKED ISSUE #2276: Text misalignment under "Failed logins" and "Last access" headers
## Expected Behaviour
The text within the user list are aligned with their respective headers:
* Email address, Tenant, and Role are left-aligned.
* Failed logins, and Last access are center-aligned.

## Actual Behaviour
The text for the Failed logins, and Last access are not center-aligned with their headers. They are not left or right aligned either. 

## Steps to Reproduce
1. Navigate to User management. Notice that the text under "Failed logins" and "Last access" are further to the right of their respective headers.

## Versions
  - Authoring Tool Version: v0.7.0
  - Framework Version: v2.3.2

canstudios-simonasd:
Fixes #2276

### DIFF
--- a/frontend/src/modules/userManagement/less/userManagement.less
+++ b/frontend/src/modules/userManagement/less/userManagement.less
@@ -79,6 +79,10 @@
       .form-container-style;
       max-width: inherit;
       .tb-heading {
+        .col-centered {
+          display: flex;
+          justify-content: center;
+        }
         > div button {
           display: flex;
           align-items: center;


canstudios-simonasd:
There is sort arrow, that is hidden by default. If were are taking that arrow into account than it is centred correct. We can probably make that, when arrow appears (once you click to sort by "Failed logins" or "Last access", than text can re-centre itself.
  
image

PR #2356, opened 2019-06-06
Remove unused user roles

### LINKED ISSUE #1950: Remove unused user roles
It seems pointless (and not to mention confusing to users) that we include role types that don't offer any difference in features. As any kind of proper role support isn't anywhere on the horizon, I vote we remove these. We might need to write a migration to remove old roles from the DB, and any assigned to existing users.

Roles affected:
- Tenant admin
- Product manager
- Authenticated user

canstudios-simonasd:
Migrate all users from Product Manager or Tenant Admin to Course Creator. Then remove Product Manager and Tenant Admin roles from database.

Fixes #1950

### DIFF
--- a/lib/role/ProductManager.json
+++ b/lib/role/ProductManager.json
@@ -1,7 +1,0 @@
-{
-  "name": "Product Manager",
-  "version": "0.0.1",
- "statement" : [
-    { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/api/product/*", "action":["create","read","update","delete"], "effect":"allow" }
-  ]
-}
--- a/lib/role/TenantAdmin.json
+++ b/lib/role/TenantAdmin.json
@@ -1,7 +1,0 @@
-{
-  "name": "Tenant Admin",
-  "version": "0.0.1",
-  "statement" : [
-    { "condition":{}, "resource":"urn:x-adapt:{{tenantid}}/*", "action":["create","read","update","delete"], "effect":"allow" }
-  ]
-}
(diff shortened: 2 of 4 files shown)


canstudios-nicolaw:
@tomgreenfield When we have discussed this in previous calls we had agreed that it would be better to reduce permissions rather than elevate them, this is mentioned on the linked issue.

It's not clear to me that tenant admins should be elevated to super admins. I could see some scenarios where they may be being used to deliberately restrict certain permissions such as plugin upload and converting these users to super admins may cause some problems.

canstudios-nicolaw:
I've rebased so that `release/lessismore` and `issue/1950` are both on current master. Github is showing two extra commits in this PR and some extra files in the files changed diff but no additional changes.

tomgreenfield:
Do we know if there's a discernible difference between super and tenant admins? It's difficult to tell from the [role files](https://github.com/adaptlearning/adapt_authoring/tree/master/lib/role).

I can certainly upload plugins and do everything in user management as a tenant admin so am just slightly concerned that clients with these accounts will lose their ability to manage the tool.

canstudios-nicolaw:
@tomgreenfield Hm ok, I think I must be thinking of changes which must have gone to the tenant management branch which restricted plugin upload for tenant admins.

I will look at amending this to convert tenant admins -> super admins

canstudios-nicolaw:
I'm going to push some changes to this branch. They are not ready for testing yet, I will add the 'awaiting-review' label when that is the case.

canstudios-nicolaw:
@dancgray Please could you let me know if you think you would be able to test this this week?

dancgray:
@canstudios-nicolaw testing now 👍

canstudios-nicolaw:
@dancgray Fab thank you :grin:

PR #2359, opened 2019-06-09
Resolves #2358 - allow rgba values to be set via ColourPicker input

### LINKED ISSUE #2358: Theme editing issues
## Expected Behaviour

a) I should be able to select a transparent (rgba) colour for my theme
b) I should be able to specify a select option for a theme variable in the theme's properties schema

## Actual Behaviour

a) colour is saved as hex, so any alpha value selected using the colour picker is lost
b) a script error occurs as EditorThemingView assumes `fieldView.schema.inputType` is a string

## Steps to Reproduce
a) select a transparent colour in the theme editor, save and preview theme. 
b) create a theme with at least one property where InputType is `{ "type": "Select", "options": [...] }`

## Versions
  - Framework Version: 4.2.0

danielstorey:
- Fix error when a theme property has a select input type
- Remove duplicate property from scaffoldColourPickerView

This checks the alpha value when saving a colour and saves the rgba value if less than 1.

### DIFF
--- a/frontend/src/modules/editor/themeEditor/views/editorThemingView.js
+++ b/frontend/src/modules/editor/themeEditor/views/editorThemingView.js
@@ -212,7 +212,7 @@ define(function(require) {
       }
       if (fieldView.schema.inputType === 'ColourPicker') {
         fieldView.setValue(value);
-      } else if (fieldView.schema.inputType.indexOf('Asset:') > -1) {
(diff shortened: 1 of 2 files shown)


canstudios-nicolaw:
@danielstorey Please can you edit this PR so that it points to `release/bugpatch` instead of `master`

PR #2383, opened 2019-07-10
Password reveal

Link2Twenty:
I've added a password reveal to the main login page using the material design SVGs
https://material.io/tools/icons/?style=baseline

#2219

password_reveal

### DIFF
--- a/frontend/src/modules/user/templates/login.hbs
+++ b/frontend/src/modules/user/templates/login.hbs
@@ -12,13 +12,13 @@
                     <i class="fa fa-times"></i><span id="loginErrorMessage"/>
                 </div>
                 <input type="email" autocomplete="off" class="login-input-username" value="" id="login-input-username" placeholder="{{t 'app.usernameoremail'}}" tabindex="1" />
-                <div class="login-password-container">                
-                  <input type="password" autocomplete="off" class="login-input-password" value="" id="login-input-password" placeholder="{{t 'app.password'}}" tabindex="2" />
-                  <button aria-pressed="false" class="login-password-reveal">
-                    <i class="fa fa-eye login-password-reveal-show"></i>
-                    <i class="fa fa-eye-slash login-password-reveal-hide"></i>
+                <span class="reveal-password-container">
+                  <input type="password" class="width-15 reveal-input-password" id="password" value="" autocomplete="off">
+                  <button aria-pressed="false" class="reveal-password-button">
+                    <i class="fa fa-eye reveal-password-show"></i>
+                    <i class="fa fa-eye-slash reveal-password-hide"></i>
                   </button>
-                </div>
+                </span>
                 <button type="submit" class="btn secondary login-form-submit" tabindex="3">{{t 'app.login'}}</button>
             </div>
             <div class="login-remember-me">
--- a/frontend/src/modules/user/templates/login.hbs
+++ b/frontend/src/modules/user/templates/login.hbs
@@ -12,7 +12,13 @@
                     <i class="fa fa-times"></i><span id="loginErrorMessage"/>
                 </div>
                 <input type="email" autocomplete="off" class="login-input-username" value="" id="login-input-username" placeholder="{{t 'app.usernameoremail'}}" tabindex="1" />
-                <input type="password" autocomplete="off" class="login-input-password" value="" id="login-input-password" placeholder="{{t 'app.password'}}" tabindex="2" />
+                <div class="login-password-container">                
+                  <input type="password" autocomplete="off" class="login-input-password" value="" id="login-input-password" placeholder="{{t 'app.password'}}" tabindex="2" />
+                  <button class="login-password-reveal">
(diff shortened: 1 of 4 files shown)


frontend/src/modules/user/templates/login.hbs
```diff
@@ -12,7 +12,13 @@
...
+                <div class="login-password-container">                
+                  <input type="password" autocomplete="off" class="login-input-password" value="" id="login-input-password" placeholder="{{t 'app.password'}}" tabindex="2" />
+                  <button aria-pressed="false" class="login-password-reveal">
+                    <svg class="login-password-reveal-show" xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M0 0h24v24H0zm0 0h24v24H0zm0 0h24v24H0zm0 0h24v24H0z" fill="none"/><path d="M12 7c2.76 0 5 2.24 5 5 0 .65-.13 1.26-.36 1.83l2.92 2.92c1.51-1.26 2.7-2.89 3.43-4.75-1.73-4.39-6-7.5-11-7.5-1.4 0-2.74.25-3.98.7l2.16 2.16C10.74 7.13 11.35 7 12 7zM2 4.27l2.28 2.28.46.46C3.08 8.3 1.78 10.02 1 12c1.73 4.39 6 7.5 11 7.5 1.55 0 3.03-.3 4.38-.84l.42.42L19.73 22 21 20.73 3.27 3 2 4.27zM7.53 9.8l1.55 1.55c-.05.21-.08.43-.08.65 0 1.66 1.34 3 3 3 .22 0 .44-.03.65-.08l1.55 1.55c-.67.33-1.41.53-2.2.53-2.76 0-5-2.24-5-5 0-.79.2-1.53.53-2.2zm4.31-.78l3.15 3.15.02-.16c0-1.66-1.34-3-3-3l-.17.01z"/></svg>
```
canstudios-nicolaw: It looks like the user profile already has the ability to unmask password input. The user profile uses the font awesome eye icon. Please could you use the same thing here for consistency?

Link2Twenty: Is the position ok? Or would you rather have the eye outside the input (alternatively, I could user profile's version to be within the input)?

canstudios-nicolaw: I am happy with the position, I have seen other sites that do this and position the eye inside the password field. I think it would make sense to move the user profile eye icon inside the box as well for consistency if you don't mind.

I'm not sure if you've tested this on Edge, but just so that you're aware it has its own password reveal icon that it puts inside password fields - you may need to hide this on edge so that they don't interfere with each other.

Link2Twenty:
OK, I've moved it over to Font Awesome and made sure it works in IE and Edge.

password_reveal_fa

I'll do another PR for the user profile icon.

canstudios-nicolaw:
@Link2Twenty Fab thanks, I'd be happy for the user profile icon change to be part of this PR if you'd like as it makes more sense alongside these changes and is small :)

Link2Twenty:
@canstudios-nicolaw sorry, I've only just noticed your response I've made another PR #2384

canstudios-nicolaw:
@Link2Twenty It looks like both PRs contain very similar Less. It would be good if that could be put in one place that was used by both the login and the user profile. I'm not sure off the top of my head where that would be but can have a look tomorrow.

Link2Twenty:
I'd guess adapt_authoring/frontend/src/core/less/forms.less

We'd have to change the templates so the styles could be a little more generic but that's fine.

canstudios-nicolaw:
I was just having a look and thinking about frontend/src/modules/user/less/ and having a new file called "sharedUserStyles.less" or similar. Since both of the less files are in that directory anyway. I think I'd be equally happy for it to go in forms.less

Link2Twenty:
The third location in the original Issue is `adapt_authoring/frontend/src/modules/userManagement/` so outside of the user directory might be best, if we want to use the same styles again, of course.

Link2Twenty:
I think we're there 🙂

tomgreenfield:
Hi guys, sorry to jump onto this late.

The original issue (#2219) asks for a password reveal in places where the password is being saved to the database, i.e. on user creation or change. This PR adds it to the login screen – this isn't functionality I've noticed on any of the big websites. Have we general consensus for this to be added here?

Link2Twenty:
@tomgreenfield Google login forms (like Gmail) have a way to reveal passwords, though I can't find anything about it in the material spec. 

Also, as @canstudios-nicolaw said Edge and IE add the reveal functionality themselves.

canstudios-nicolaw:
@tomgreenfield @Link2Twenty 

I think Tom makes a good point, and I think it is best to discuss issues before working on them to avoid this sort of thing, and we should ask ourselves if we want changes before putting them in. Generally our better defined/higher priority issues are on the bugpatch milestone.

That being said, I know we have implemented this sort of thing before on some of our products, and IE/Edge do this as standard. I would be happy for it to go in. Tom do you feel strongly otherwise?

tomgreenfield:
@canstudios-nicolaw, you've caught my drift. Was mainly opening the discussion to avoid feature creep and/or security concerns.

I read somewhere a while back that companies tend to avoid toggles on their web-based login forms due to users being able to exploit password managers' auto-fill functionality to reveal plain text passwords. No strong feelings either way from me though.

canstudios-nicolaw:
I am happy for this to go in. I will give it a proper test/approve as soon as 0.9.0 is out

frontend/src/modules/user/views/loginView.js
```diff
@@ -30,6 +31,21 @@ define(function(require) {
       e && e.preventDefault();
       Origin.router.navigateToHome();
     },
+    
+    togglePassword: function(e) {
+      e && e.preventDefault();
+      
+      var passwordField = document.querySelector('.reveal-input-password');
```
canstudios-nicolaw: In adapt this type of selector isn't common, we tend to use jQuery selectors, please could you update this

```
```

fkcailiniyazi:
I think this is a great feature, shouldn't we have this feature in our latest release?

Link2Twenty:
Is it worth me keeping these pull requests open or shall I close them?

taylortom:
@Link2Twenty leave this for now, and I'll bring up in the weekly call :)

taylortom:

PR #2384, opened 2019-07-10
User profile password reveal

Link2Twenty:
The password reveal icon is now inside the input rather than next to it

#2219
#2383

password_reveal_UD


Link2Twenty:
This has been generalised as well now.

frontend/src/modules/user/views/userProfileView.js
```diff
@@ -69,10 +66,17 @@ define(function(require){
 
     togglePasswordView: function() {
       event && event.preventDefault();
+      
+      var passwordField = document.querySelector('.reveal-input-password');
```
canstudios-nicolaw: In adapt this type of selector isn't common, we tend to use jQuery selectors, please could you update this

```
```

taylortom:

PR #2389, opened 2019-07-15
Cross-framework import support

### LINKED ISSUE #2288: Add support for importing a course with an earlier framework version
## Affected Area
If I have made a course with an earlier version of the framework I should be able to import it into my current authoring tool. At the moment if I try to import a v2 framework course into an authoring tool that is running v4 framework then the import will fail.

## Requested Feature
In this case I should be guided through the process of importing my v2 course. Exact form of guidance is tbc but could be modals or additional import form fields.

Guidance will be a case of informing me that:
- this course was originally built on an older framework
- subsequently there may be some breakage and the course should be tested after import
- (possibly) switch the theme to v4 vanilla so that there is at least a basic theme that works
- check all course plugins are compatible with v4 framework. There are a few possibilities:
    - The plugins are compatible but need to be installed - install as part of normal import
    - The plugin that is on the course is not compatible with v4 but a newer version of that plugin is already installed - give warning
    - The plugin that is on the course is not compatible with v4, a newer version exists but is not installed - this is possibly not mvp, need to look into effort estimate for this
    - The plugin is not compatible with v4 and no newer versions of it exist - prevent import 

## Use Case
There are many users who would like to use v4 framework in the authoring tool but also need to maintain older courses written in v2. This process will help them import v2 courses into a v4 authoring tool for easier maintenance. 

## Current Workaround
Currently it is not possible to import a v2 course into the v4 authoring tool

canstudios-nicolaw:
Fixes #2288 

This is a PR for the cross framework import support feature, it matches the original proposal document which was laid out previously.

The user now gets guidance through the import process so that they can see which plugins they need to be update and which can be updated for them. There is no longer a blanket ban on importing courses that have an older framework version.

I have created a more detailed document to aid testing here: https://docs.google.com/document/d/1geSzp0OCPdCzJ4htRG4GsOxOkjE1pa103m-gQjBgXvM/edit?usp=sharing. This explains the routes that the code takes as well as some expected scenarios.


plugins/output/adapt/helpers.js
```diff
@@ -206,33 +209,69 @@ function checkFrameworkVersion(versionMetaData, cb) {
...
+                return cb(null, {category: 'green', newVersion: pluginMetaData.version});
+              }
+              cb(null, {category: 'red', newVersion: app.polyglot.t('app.na')});
+          } else {
+              // A version is already installed so could be:
+              // White - same version installed as in course
+              // Amber - newer version installed and will be used in the course instead
+              // Green - older version installed so this newer version will be upgraded in the AT
+              var serverPlugin = records[0];
+              var serverPluginVersion = semver.clean(serverPlugin.version);
```
tomgreenfield: While I remember, we should short circuit if someone has really messed up their versioning. We could probably use the red category for this.

```js
var serverPluginVersion = semver.valid(serverPlugin.version);
var importPluginVersion = semver.valid(pluginMetadata.version);

if (!serverPluginVersion || !importPluginVersion) {
  // return early
}
```

canstudios-nicolaw:
Thanks for taking the time to review this! Much appreciated. 

I will talk this through internally first thing tomorrow and get back to you.

canstudios-nicolaw:
@tomgreenfield 

Having discussed this internally we have a few more suggestions to the UI you have proposed.

- The text in the action column has an appropriate font colour but doesn't have the coloured background as it does now. We think that in the current design these labels look a bit button-y and having text like "Install" and "Update" could be more confusing if the button-y design was kept.

- Instead of 'Import Version' and 'Installed Version' headings, we suggest calling them 'Course Version' and 'Authoring Tool Version' which we hope will make the end result clearer

- We suggest that in each row (with the exception of red plugin rows) the 'winning' plugin version is made bold to indicate that this will be the end result. In yellow columns this would mean that the 3rd column would be bold, and in green columns the 2nd column would be bold

- Depending on how it looks, we wonder if it would be worth putting tooltips on actions to explain. Particularly "Skip" seems a little ambiguous (we couldn't think of a better alternative). I would complete the other parts first and we could decide on this later, just wanted to suggest it.

What do you think?

tomgreenfield:
> The text in the action column has an appropriate font colour but doesn't have the coloured background as it does now. We think that in the current design these labels look a bit button-y and having text like "Install" and "Update" could be more confusing if the button-y design was kept.

Completely agree, I originally tried to click the coloured boxes myself!

> Instead of 'Import Version' and 'Installed Version' headings, we suggest calling them 'Course Version' and 'Authoring Tool Version' which we hope will make the end result clearer

We indeed discussed having 'authoring tool version' as a heading so that seems okay. I'm still keen to avoid calling it the 'course version' due to a bit of ambiguity over what the 'course' is, and the fact that the import ZIP is an entire source package rather than just the course content. Can't think of anything more clear than 'import version' or 'imported version' at this point in the day!

> We suggest that in each row (with the exception of red plugin rows) the 'winning' plugin version is made bold to indicate that this will be the end result. In yellow columns this would mean that the 3rd column would be bold, and in green columns the 2nd column would be bold

Interesting idea, I like it.

> Depending on how it looks, we wonder if it would be worth putting tooltips on actions to explain. Particularly "Skip" seems a little ambiguous (we couldn't think of a better alternative). I would complete the other parts first and we could decide on this later, just wanted to suggest it.

I pondered over tooltips while reviewing. Not to duplicate the text in the legend/key but rather to give more specific information, such as why the plugin is incompatible, e.g. if it isn't supported in the installed framework version or if there's a dodgy semver somewhere. 'Skip' could possibly become something like 'Use installed version'.

canstudios-nicolaw:
:+1: I'll get working on this and we can discuss further on Monday

dancgray:
Thanks for all of the work that has gone into this. Works well so far.

On the second import page the import source button should have a state change after it is clicked. At the moment it is unclear if anything is happening until the import completes.

canstudios-nicolaw:
Thanks for testing @dancgray :)

Thanks for the feedback - it will be good to implement that at the same time as the other UI changes :)

canstudios-nicolaw:
@dancgray @tomgreenfield I've attached a screenshot with these changes, I've got a few things left to test before I commit. Any thoughts so far?

Screenshot from 2019-07-22 13:31:12

canstudios-nicolaw:
I have also updated the testing document to reflect these latest changes to the table.

canstudios-nicolaw:
I have added the invalid semver check - I believe all actions have been completed and this is ready for re-review.

frontend/src/modules/frameworkImport/less/frameworkImport.less
```diff
@@ -43,4 +55,62 @@
...
+      &.amber {
+        color: @framework-import-amber;
+      }
+
+      &.green {
+        color: @framework-import-green;
+      }
+    }
+    .description {
+      font-weight: 600;
```
tomgreenfield:
```suggestion
      font-weight: @font-weight-bold;
```

frontend/src/modules/frameworkImport/less/frameworkImport.less
```diff
@@ -43,4 +55,62 @@
...
+      }
+    }
+
+    .amber {
+      .status, .key {
+        color: @framework-import-amber;
+      }
+
+      .at-version {
+        font-weight: 600;
```
tomgreenfield:
```suggestion
        font-weight: @font-weight-bold;
```

frontend/src/modules/frameworkImport/less/frameworkImport.less
```diff
@@ -43,4 +55,62 @@
...
+      }
+    }
+
+    .green-install, .green-update {
+      .status, .key {
+        color: @framework-import-green;
+      }
+
+      .import-version {
+        font-weight: 600;
```
tomgreenfield:
```suggestion
        font-weight: @font-weight-bold;
```

frontend/src/modules/frameworkImport/less/frameworkImport.less
```diff
@@ -43,4 +55,62 @@
...
+      }
+    }
+
+    .status, .key {
+      font-weight: 600;
```
tomgreenfield:
```suggestion
      font-weight: @font-weight-bold;
```

frontend/src/modules/frameworkImport/templates/frameworkImport.hbs
```diff
@@ -1,5 +1,5 @@
-<div class="inner">
-  <form class="frameworkImport forms" action="importsource" method="post" autocomplete="off">
+<div id="import_upload" class="inner">
+  <form class="frameworkImport forms" action="/importsourcecheck" method="post" autocomplete="off">
```
tomgreenfield: Should this be without the slash to give a relative path? @oscarsiles?

frontend/src/modules/frameworkImport/templates/frameworkImport.hbs
```diff
@@ -40,3 +40,19 @@
       </div>
   </form>
 </div>
+
+<div id="import_details" class="display-none">
+    <form class="frameworkImportDetails forms" action="/importsource" method="post" autocomplete="off">
```
tomgreenfield: Relative path?

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -12,7 +12,8 @@ define(function(require){
 
     preRender: function() {
       Origin.trigger('location:title:update', { title: Origin.l10n.t('app.frameworkimporttitle') });
-      this.listenTo(Origin, 'frameworkImport:uploadCourse', this.uploadCourse);
+      this.listenTo(Origin, 'frameworkImport:showDetails', this.showDetails);
```
tomgreenfield: Please combine using object-style syntax, i.e.
```js
this.listenTo(Origin, {
  'frameworkImport:showDetails': this.showDetails,
  'frameworkImport:completeImport': this.completeImport
});
```

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -65,12 +67,112 @@ define(function(require){
...
+        $('.framework-import-sidebar-save-button').remove();
+      }
+
+      // Framework versions panel
+      if (data.frameworkVersions.imported !== data.frameworkVersions.installed) {
+          $framework_versions.html(Origin.l10n.t('app.importframeworkversions', {
+              importVersion: data.frameworkVersions.imported,
+              installedVersion: data.frameworkVersions.installed
+          }));
+          $framework_versions.removeClass('display-none');
```
tomgreenfield: Can chain this to the previous line if you liked.

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -65,12 +67,112 @@ define(function(require){
...
+      if (!_.isEmpty(data.pluginVersions.amber) || greenExists) {
+        $summary_title.addClass('amber');
+        $summary_description.text(Origin.l10n.t('app.coursecanbeimporteddesc'));
+        if (greenExists && _.isEmpty(data.pluginVersions.amber)) {
+          $summary_title.removeClass('amber').addClass('green');
+        }
+        return;
+      }
+
+      $summary_description.text(Origin.l10n.t('app.coursecanbeimportedwhitedesc'));
```
tomgreenfield: Can the logic be reversed here so we can return with this one-liner?

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -65,12 +67,112 @@ define(function(require){
...
+        },
+        'red': {
+          label: 'app.pluginredlabel'
+        }
+      };
+      var categories = ['red', 'amber', 'green-update', 'green-install'];
+
+      $plugin_list.append(new FrameworkImportPluginHeadingView().$el);
+
+      for (i = 0; i < categories.length; i++) {
```
tomgreenfield:
```suggestion
      for (var i = 0, j = categories.length; i < j; i++) {
```

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -65,12 +67,112 @@ define(function(require){
         },
 
         error: _.bind(this.onAjaxError, this),
-        success: _.bind(this.onFormSubmitSuccess, this)
+        success: _.bind(this.displayDetails, this)
```
tomgreenfield: Should be going for the native `.bind()`s really. Up to you whether you can be bothered to switch them!

plugins/output/adapt/importsourcecheck.js
```diff
@@ -0,0 +1,244 @@
...
+                        details.frameworkVersions = data;
+                        cb();
+                    });
+                });
+            },
+            checkContentJson: ['checkFramework', function(results, cb) {
+                async.eachSeries(Object.keys(contentMap), function(type, cb2) {
+                    var jsonPath = path.join(COURSE_JSON_PATH, (type !== 'config') ? COURSE_LANG : '', `${contentMap[type] || type}.json`);
+                    if (!fs.existsSync(jsonPath)) {
+                        return cb2(error);
```
tomgreenfield: `error` is undefined.

plugins/output/adapt/importsourcecheck.js
```diff
@@ -0,0 +1,244 @@
...
+            }],
+            checkAssetFolders: ['checkContentJson', function(results, cb) {
+                if (!cleanFormAssetDirs.length) {
+                    importInfo['assetFolders'] = Constants.Folders.ImportAssets;
+                    return cb();
+                }
+                var assetFolderError = false;
+                var missingFolders = [];
+                importInfo['assetFolders'] = cleanFormAssetDirs;
+                for (index = 0; index < importInfo['assetFolders'].length; ++index) {
```
tomgreenfield:
```suggestion
                for (var i = 0, j = importInfo['assetFolders'].length; i < j; ++i) {
```
(or similar)

plugins/output/adapt/importsourcecheck.js
```diff
@@ -0,0 +1,244 @@
...
+                }, function(err) {
+                    if(err) {
+                        return done(err);
+                    }
+
+                    importInfo['pluginIncludes'] = plugindata.pluginIncludes;
+
+                    async.each(plugindata.pluginIncludes, function (pluginData, cb2) {
+                        fs.readJSON(path.join(pluginData.location, Constants.Filenames.Bower), function(error, data) {
+                            if (error) return callback(error);
```
tomgreenfield: `callback` is undefined.

routes/lang/en-application.json
```diff
@@ -351,6 +352,23 @@
   "app.importassetfolderstitle": "Asset Folders",
   "app.importassetfoldersplaceholder": "assets, images, video, audio",
   "app.importinvalidpackage": "Encountered an unexpected file structure. Please check you have uploaded a valid export package and try again.",
+  "app.importframeworkversions": "Import framework version: %{importVersion}<br/>Installed framework version: %{installedVersion}<br/><br/>If you proceed with this import then your course will use framework %{installedVersion} and should be tested.",
+  "app.coursecanbeimported": "Course can be imported",
+  "app.coursecanbeimportedwhitedesc": "All of the plugins and the framework version used in this course are identical to that in this authoring tool.",
```
tomgreenfield:
```suggestion
  "app.coursecanbeimportedwhitedesc": "All of the plugins and the framework version used in this course are identical to those in this authoring tool.",
```

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -65,12 +67,112 @@ define(function(require){
...
       });
 
       return false;
     },
 
+    displayDetails: function(data) {
+      this.$('#import_upload').addClass('display-none');
+      this.sidebarView.$('.framework-import-sidebar-save-button').removeClass('show-details');
+      var $details = this.$('#import_details');
+      var $framework_versions = $details.find('.framework-versions');
```
tomgreenfield: We'd usually go for camel case variables e.g. `$frameworkVersions`.

frontend/src/modules/frameworkImport/views/frameworkImportView.js
canstudios-nicolaw: I'm not sure what the benefit of this is? It's not something I've come across before!

tomgreenfield: They're not biggies – first amend was from the linter complaining that `i` hasn't been defined.

The second amend I just slipped in at the same time for good practice – caching the length so it's not looked up with every iteration. Example [here](http://allthingsjavascript.com/blog/index.php/2016/11/16/caching-the-length-property-in-a-for-loop).

canstudios-nicolaw: :+1: That is a good point I hadn't thought of before, thanks

canstudios-nicolaw:
Indentation fixes are in progress

plugins/output/adapt/importsourcecheck.js
tomgreenfield: Need to switch the references from `index` to `i` inside the loop. Could do with a variable e.g.
```js
var assetFolder = importInfo.assetFolders[i];
```

frontend/src/modules/frameworkImport/views/frameworkImportView.js
tomgreenfield: See also the new code on lines 230 & 231.

plugins/output/adapt/importsourcecheck.js
```diff
@@ -0,0 +1,244 @@
...
+          // clear any previous import files
+          fs.emptyDir(COURSE_ROOT_FOLDER, function(error) {
+            if(error) return cb2(error);
+            // upzip the uploaded file
+            logger.log('info', 'unzipping');
+            logger.log('info', COURSE_ROOT_FOLDER);
+            logger.log('info', files.file.path)
+            helpers.unzip(files.file.path, COURSE_ROOT_FOLDER, function(error) {
+              if(error) return cb2(error);
+                importInfo['cleanupDirs'] = [files.file.path, COURSE_ROOT_FOLDER];
```
tomgreenfield: Indentation slightly off on lines 106–108.

canstudios-nicolaw:
As discussed in the call today I will be updating this so that 'backwards' imports are blocked (it will not be possible to import a FWv4 course into an AT that is running FWv2 for example)

canstudios-nicolaw:
Screenshot from 2019-07-30 09:51:24

Added a simple import not allowed message for downgrade cases.

canstudios-nicolaw:
Our QA team have now finished testing this and are satisfied that it is ready for release.

The only odd behaviour encountered was what happens if one user performs actions that affect a second users import while it is running. I think this is an acceptable edge case for now as there is similar behaviour if users edit the same course.

"user 1 removes a plugin that appears in the user's 2 course during user 2's cross framework import.
In the above scenario:
- User 2 hangs as the course import buttons, and a 502 proxy error eventually occurs.
- User 2 is still in the plugins list step of course import and can still click the Import source button.
- The course is imported, though without the missing plugin"

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -64,13 +68,117 @@ define(function(require){
...
+
+      $plugin_list.append(new FrameworkImportPluginHeadingView().$el);
+
+      for (var i = 0, j = categories.length; i < j; i++) {
+        var categoryData = data.pluginVersions[categories[i]];
+        if (categories[i] === 'white') continue;
+        if (_.isEmpty(categoryData)) continue;
+
+        // Sort plugins alphabetically
+        var pluginArray = Object.values(categoryData);
```
tomgreenfield: Just noticed this line might cause an issue for Internet Explorer 11 users.

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -64,13 +68,117 @@ define(function(require){
...
+      // Plugin list
+      this.displayPluginList(data);
+
+      $details.removeClass('display-none');
+    },
+
+    displaySummary: function(data) {
+      var $details = this.$('#import_details');
+      var $summary_title = $details.find('.import-summary .title');
+      var $summary_description = $details.find('.import-summary .description');
```
tomgreenfield: Suggest `$summaryTitle` & `$summaryDescription`.

frontend/src/modules/frameworkImport/views/frameworkImportView.js
```diff
@@ -64,13 +68,117 @@ define(function(require){
...
+
+      $summary_title.addClass('amber');
+      $summary_description.text(Origin.l10n.t('app.coursecanbeimporteddesc'));
+      if (greenExists && _.isEmpty(data.pluginVersions.amber)) {
+        $summary_title.removeClass('amber').addClass('green');
+      }
+    },
+
+    displayPluginList: function(data) {
+      var $plugin_list = this.$('#import_details').find('.plugin-list');
```
tomgreenfield: Suggest `$pluginList`.

Rahulc24:
Hi,

I'm currently trying to make some textual updates to a Adapt course that was developed by someone else a few years ago. Unfortunately, I do not have access to the original Adapt authoring instance used to create it. All I have are a few ZIP files, which I believe are the source packages. I attempted to import these ZIP files into my current Adapt authoring tool instance but encountered the following error:
“This course can't be imported.”
(Screenshot attached)

After investigating, it seems that the original course may have been built using custom or local plugins that are no longer available in the public Adapt plugin library. I suspect this is causing compatibility issues, especially since the framework and plugins used at that time appear to be outdated compared to the current Adapt version.

I also searched the Adapt public plugin library for the missing plugins referenced in the course files but was unable to find them.

Could you please advise on:

1) Whether there’s a workaround to import such older courses?
2) Please suggest ways to solve this.

Please note that I’ve only recently started using the Adapt authoring tool- just a few days ago- so I’m still learning the platform. I would really appreciate your guidance on how best to proceed. I would really appreciate your guidance at the earliest, as I’m on a tight deadline to update and redeploy this course.

Package contents (Source file sample) (1)
Screenshot 3466 (1)

ambivalent-axiom:
It is possible that you have a build. Either way I would attempt to make
changes directly into the code instead of importing it to authoring tool.

Rahulc24:
@ambivalent-axiom Thanks for the suggestion. I am not an expert in the method you mentioned, so are there any alternative approaches?. Also, could you explain the reason behind the issue and suggest possible solutions?

ambivalent-axiom:
Just extract and open content directory in code editor, then look for
places you need to edit by using search.

No I can’t, to explain anything I need to see at least the file and its
contents + adapt logs after unsuccessful import attempt.

Issue #1963, opened 2018-05-15
When "Add component" is clicked change focus


simondate:
## Affected Area

Page editor screen.

## Requested Feature

When you click the Add component button the focus should change to the component search bar, so the user can instantly typing for their component.

## Use Case

Speed up AT use.

## Current Workaround

Clicking on the searchbar

PR #2390, opened 2019-07-16
When "Add component" is clicked change focus (#1963)

Link2Twenty:
## Proposed changes

Set focus when "Add component" is clicked #1963

## Checklist
- [x] I have read the [CONTRIBUTING](https://github.com/adaptlearning/adapt_authoring/blob/master/.github/CONTRIBUTING.md) doc
- [x] I have checked that the code compiles correctly
- [x] I have checked that unit test suite passes locally with my changes
- [x] I have added new tests that cover my changes (where appropriate)
- [x] I have added documentation (where appropriate)
- [x] Any dependent changes have already been merged

### DIFF
--- a/frontend/src/modules/editor/contentObject/views/editorPageBlockView.js
+++ b/frontend/src/modules/editor/contentObject/views/editorPageBlockView.js
@@ -261,6 +261,8 @@ define(function(require){
         $parentElement: this.$el,
         parentView: this
       }).$el);
+      
+      $('.editor-component-list-sidebar-search-field input').focus();
     },
 
     setupPasteZones: function() {
--- a/frontend/src/modules/editor/contentObject/views/editorPageBlockView.js
+++ b/frontend/src/modules/editor/contentObject/views/editorPageBlockView.js
@@ -261,8 +261,6 @@ define(function(require){
         $parentElement: this.$el,
         parentView: this
       }).$el);
-      
-      $('.editor-component-list-sidebar-search-field input').focus();
     },
 
     setupPasteZones: function() {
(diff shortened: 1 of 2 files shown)


canstudios-nicolaw:
@Link2Twenty Thank you for your continued work, it is appreciated

I can understand why you picked this up, but the "Suggested Next" milestone is a little misleading. Many of the things on this milestone haven't been discussed/thought about in a long time and might not still be desired.

There is going to be a big issue sort/tidy soon to make this clearer, but in the meantime the best issues to work on are the ones on the bugpatch milestone (https://github.com/adaptlearning/adapt_authoring/issues?q=is%3Aopen+is%3Aissue+milestone%3ABugpatch) as these have been most recently sorted as high priority.

I understand there aren't many straightforward beginner issues on the bugpatch milestone right now, so if you'd like to work on an issue that is not on the bugpatch milestone I'd strongly suggest either asking in gitter or tagging myself or @tomgreenfield on the issue to talk about it first. I only suggest this because I don't want you to loose time working on things that turn out to be undesired.

@tomgreenfield I have asked our content team about this, they feel fairly neutral. I think it's acceptable for it to go in because I can't see it making usability worse for people that don't want to use this feature, what do you think?

canstudios-nicolaw:
I've had some more feedback that this would be useful for some people so I'm going to approve and leave the merging to you if you are happy.

Link2Twenty:
> Could you move the focus to the `postRender()` of the `editorPageComponentListView` itself? This feels like a better fit than the parent view appending the child and doing a new lookup for the element to focus on.

I've moved it over to the new file, I've kept the `.editor-component-list-sidebar-search-field` class in the selector just in case any other inputs are ever added to the sidebar, not that that looks likely to happen.

PR #2391, opened 2019-07-16
Add userAdd password field toggle

Link2Twenty:
Add a password reveal icon to toggle password visibility.

#2219
Relies on #2383

password_reveal_AU

### DIFF
--- a/frontend/src/modules/userManagement/views/addUserView.js
+++ b/frontend/src/modules/userManagement/views/addUserView.js
@@ -28,6 +28,9 @@ define(function(require){
       this.$('.field-error').each(function(index, element) {
         var $error = $(element);
         var $input = $error.siblings('input');
+        
+        // Password is no longer a sibling so needs special treatment
+        $input = $input.length === 0 ? $error.siblings('.reveal-password-container').children('input') : $input;
 
         var isValid = $input.attr('name') === 'email' ?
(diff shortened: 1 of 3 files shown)


frontend/src/modules/userManagement/views/addUserView.js
```diff
@@ -96,6 +103,21 @@ define(function(require){
         title: "Couldn't add user",
         text: data.responseText || error
       });
+    },
+
+    togglePasswordView: function() {
+      event && event.preventDefault();
+      
+      var passwordField = document.querySelector('.reveal-input-password');
```
canstudios-nicolaw: In adapt this type of selector isn't common, we tend to use jQuery selectors, please could you update this

```
```

frontend/src/modules/userManagement/views/addUserView.js
```diff
@@ -24,6 +28,9 @@ define(function(require){
       this.$('.field-error').each(function(index, element) {
         var $error = $(element);
         var $input = $error.siblings('input');
+        
+        // Password is no longer a sibling so needs special treatment
```
canstudios-nicolaw: Please could you remove this comment, somebody looking at this file after this PR isn't going to know that password ever used to be a sibling

taylortom:

PR #2397, opened 2019-07-22
hide navbar if empty

### LINKED ISSUE #1541: Allow for pages without a sidebar
If no sidebar is added, you just get the empty column. Would be nice if we could have nice styling in this case -- maybe hiding completely and centring the page.

lc-thomasberger:
fixes #1541


lc-thomasberger:
to test `$('html').addClass('sidebar-hide');`

canstudios-nicolaw:
Note this fixes #1541

PR #2398, opened 2019-07-23
formidable, clean up temp files 

### LINKED ISSUE #2115: Temp formidable upload files using disk space
We have a problem whereby files uploaded to the tool are clogging up the tmp folder (by default, [Formidable](https://github.com/felixge/node-formidable) uses the OS' default temporary files directory when processing uploads --  `/tmp` on linux). I've not been able to ascertain whether these files are left from unhandled errors, or some other reason.

Has anyone seen anything like this?

Aside from actually figuring out what's going on here, it may be useful to move the temp upload dir to the AT's temp folder, so it's a bit more obvious what files have been created by the AT.

lc-thomasberger:
fixes #2115 

Course import cleans up tempfile already.
This PR handles cleanup for uploaded files from assetmanager and pluginmanager.

Todo Serverrewrite:
- Abstract into a Class so it can be shared across the AT
- Expose uploade directory as config attribute. Default to folder inside AT-root

### DIFF
--- a/lib/assetmanager.js
+++ b/lib/assetmanager.js
@@ -4,7 +4,7 @@
  */
 
 var path = require('path'),
-    fs = require('fs'),
+    fs = require('fs-extra'),
     _ = require('underscore'),
     async = require('async'),
     util = require('util'),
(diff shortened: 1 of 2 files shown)


lib/assetmanager.js
```diff
@@ -447,34 +447,36 @@ exports = module.exports = {
...
-            );
-
-            self.createAsset(asset, function (createError, assetRec) {
-              if (createError) { // if the record creation fails, remove the file that was uploaded
-                logger.log('error', createError);
-                return storage.deleteFile(storedFile.path, function (delErr) {
-                  if (delErr) logger.log('error', 'Failed to delete stored file in assetmanager', storedFile.path);
-                  _respondError("An error occured during upload, please contact an Administrator.");
-                });
+            fs.remove(uploadedFilePath, function(error) {
```
tomgreenfield: Should there be some error handling here?

canstudios-nicolaw:
To aid testing - please could you explain which files normally get left in temp during asset and plugin upload?

plugins/content/bower/index.js
```diff
@@ -1142,8 +1142,21 @@ function handleUploadedPlugin (req, res, next) {
                   if (error) {
                     return next(error);
                   }
-                  res.statusCode = 200;
-                  return res.json({ success: true, pluginType: pluginType, message: 'successfully added new plugin' });
+
+                  function sendResponse() {
```
tomgreenfield: Could log any errors from `fs.remove()` like above but it's not crucial.

PR #2399, opened 2019-07-24
Add front end validation to preset names

### LINKED ISSUE #2382: Theme editor: Changes to error messages and behaviour when saving preset names
## Expected Behaviour
When creating or managing edit presets, error messages should appear when attempting to use taken preset names or save changes to the preset when the name field is empty.

## Actual Behaviour
a. During preset creation, user can still save when the name field is empty. The preset isn't saved and a 500 error appears in the browser console tool instead. Can the user be prevented from doing this, and an error modal warning about the field being empty appear?
b. During preset management,  user can still save when the name field is empty. The error modal displays "Error [object Object]". Please can the previous preset name be kept instead of being overwritten by nothing, and the resulting error message be similar to part (a)?
c. During preset management, the user can edit a preset's name to be the same as another preset (for that base theme). This is prevented during preset creation. Please can presets be prevented from being edited to have the same name as other presets within that base theme?

## Steps to Reproduce
1a. In Theme editor, click Save as preset. Open the browser dev console.
2a. With the name field empty, click Ok. Notice that the modal closes, yet the preset isn't saved, and a 500 error appears in the browser console.

1b. In Theme editor, click Manage presets.
2b. Click Edit name next to a preset. Clear the field.
3b. Click Save changes. Notice that the error modal reads "Error [object Object]", and that the name has been saved as an empty field.

1c. In Theme editor, click Manage presets.
2c. Click Edit name next to a preset. Edit the name to another preset on the list.
3c. Click Save changes. Notice that the changes are saved.

## Versions
  - Authoring Tool Version: v0.9.0
  - Framework Version:

canstudios-nicolaw:
Fixes #2382

### DIFF
--- a/frontend/src/modules/editor/themeEditor/less/editorPresetEdit.less
+++ b/frontend/src/modules/editor/themeEditor/less/editorPresetEdit.less
@@ -44,6 +44,10 @@
 
         .nameEdit {
           display: none;
+
+          .preset-error {
+            color: red;
+          }
         }
 
         .label,
--- a/frontend/src/modules/editor/themeEditor/templates/editorPresetEdit.hbs
+++ b/frontend/src/modules/editor/themeEditor/templates/editorPresetEdit.hbs
@@ -15,6 +15,7 @@
           </button>
         </div>
         <div class="nameEdit">
+          <div class="preset-error display-none"></div>
           <input type="text" value="{{attributes.displayName}}">
           <button class="button save btn secondary">
             {{t 'app.themepresetsave'}}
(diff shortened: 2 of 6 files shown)


frontend/src/modules/editor/themeEditor/less/editorPresetEdit.less
```diff
@@ -44,6 +44,10 @@
 
         .nameEdit {
           display: none;
+
+          .preset-error {
+            color: red;
```
lc-thomasberger: Do we have a color variable for errors?
i think it might be `@alert-color`
https://github.com/adaptlearning/adapt_authoring/blob/master/frontend/src/core/less/colours.less#L33

canstudios-nicolaw: I went with red because that is what is used on scaffold errors which is the design I am matching here (see discussion on issue)

lc-thomasberger: Cool, thanks for clarifying.
We should have the same styles for errors and warning in the Tool. But I think this can wait until a front-end refactoring.

PR #2400, opened 2019-07-29
Fix preset always defaulting to 'none'

### LINKED ISSUE #2379: Theme picker: Selected preset dropdown keeps showing "No preset"
## Expected Behaviour
When entering Theme picker area, it should show the currently applied base theme and (if one exists) preset.

## Actual Behaviour
The preset dropdown will always display "No preset" instead of the applied preset when entering the Theme picker area.

## Steps to Reproduce
1. In a course, navigate to Theme picker.
1. Select a theme and preset. Click Save.
1. Preview the course. Check preset is applied.
1. Return back to Theme picker. Notice that "No preset" is being shown.

## Versions
  - Authoring Tool Version: 0.8.1; 0.9.0
  - Framework Version:

canstudios-nicolaw:
Fixes #2379 

There will be a corresponding framework PR - I will link it below when this is made.

- When a preset has been selected, re-select it when returning to the page
- If the user switches theme and switches back, maintain their saved settings for original theme

### DIFF
--- a/frontend/src/modules/editor/themeEditor/views/editorThemingView.js
+++ b/frontend/src/modules/editor/themeEditor/views/editorThemingView.js
@@ -108,6 +108,7 @@ define(function(require) {
 
     postRender: function() {
       this.updateSelects();
+      this.setPresetSelection(Origin.editor.data.course.get('lastSelectedThemePresetId'));
       this.setViewToReady();
 
       this.$el.show();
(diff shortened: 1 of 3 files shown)


canstudios-nicolaw:
Framework PR https://github.com/adaptlearning/adapt_framework/pull/2482

canstudios-nicolaw:
I'm going to close this and create a new PR with the fix and explanation.

PR #2402, opened 2019-07-31
Prevent selects pushing button onto new line

### LINKED ISSUE #2361: Manage preset button wraps onto second line
## Expected Behaviour
Manage presets button should remain on the same line as the Base theme and Preset dropdowns.

## Actual Behaviour
Base theme and Preset dropdown widths increase to accommodate longest string within the dropdown. This will push the Manage presets button onto the line underneath and make it look untidy.

Can this be prevented?

## Steps to Reproduce
1. Import theme plugins that have 20+ characters.
1. Navigate into Theme picker within a course.
1. Create a preset with a name that has 20+ characters. Notice that the Manage presets button wraps onto the second line.

## Versions
  - Framework Version: 2.4.0

canstudios-nicolaw:
Fixes #2361 

Sets the size of the selects so that they don't push the button onto a new line and so that the size doesn't change and cause the button to jump around when selecting different themes.

### DIFF
--- a/frontend/src/modules/editor/themeEditor/less/editorTheming.less
+++ b/frontend/src/modules/editor/themeEditor/less/editorTheming.less
@@ -45,7 +45,7 @@
       padding-right: 15px;
       display: inline-block;
       select {
-        min-width: 150px;
+        width: 180px;
       }
     }
     .edit.btn.secondary {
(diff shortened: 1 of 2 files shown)


canstudios-nicolaw:
Good point - I'll take a look

PR #2445, opened 2019-10-14
Find or create assets tags on import, fixes #2439

### LINKED ISSUE #2439: Asset tags are not preserved on import
## Expected Behaviour
If I export a course which contains assets that have tags then the imported assets should have the same tags.

## Actual Behaviour
The imported assets do not have any tags.

## Steps to Reproduce
1. Create a course which contains an asset which has a tag
1. Export the course. In the assets.json you will be able to see the tag
1. Import the course. The imported asset will not have any tags.

The problem is with the import side rather than the export side. In importsource.js there is an addAssets function which contains line `var tags = formTags.slice();` I believe this is the problem.

## Versions
  - Framework Version: 4.3.0

dancgray:
Apply tags from assets.json for each asset.

### DIFF
--- a/plugins/output/adapt/importsource.js
+++ b/plugins/output/adapt/importsource.js
@@ -183,7 +183,7 @@ function ImportSource(req, done) {
                   app.contentmanager.getContentPlugin('tag', function(error, plugin) {
                     if(!error) {
                       plugin.create({ title: tag.title}, function(error, record) {
-                        if(error) logger.log('warn', 'Failed to create asset tag: ' + (tag.title || '') + ' ' + error);
+                        if(error) return logger.log('warn', 'Failed to create asset tag: ' + (tag.title || '') + ' ' + error);
                         tags.push(record._id);
                       });
                     };
--- a/plugins/output/adapt/importsource.js
+++ b/plugins/output/adapt/importsource.js
@@ -169,26 +169,25 @@ function ImportSource(req, done) {
           var fileStat = fs.statSync(assetPath);
           var assetTitle = assetName;
           var assetDescription = assetName;
+          var assetJson = assetsJson[assetName];
           var tags = [];
 
-          if (assetsJson[assetName]) {
-            assetTitle = assetsJson[assetName].title;
-            assetDescription = assetsJson[assetName].description;
+          if (assetJson) {
+            assetTitle = assetJson.title;
+            assetDescription = assetJson.description;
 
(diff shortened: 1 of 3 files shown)


canstudios-nicolaw:
I am getting error

```
ReferenceError: _ is not defined
    at /app/plugins/filestorage/localfs/index.js:235:17
    at /app/plugins/filestorage/localfs/index.js:464:12
```

When I try to import a course which contains a single asset with a tag.

Adding `const _ = require('underscore');` to plugins/filestorage/localfs/index.js fixes this for me

tomgreenfield:
If it's just that file, might be worth changing the one Underscore reference to something in the current spec:
https://github.com/adaptlearning/adapt_authoring/blob/95f04bdfd2037540247b279499976971e886e429/plugins/filestorage/localfs/index.js#L235
i.e.
```js
data = { ...data, ...withMeta };
```
or
```
data = Object.assign(data, fromMeta);
```

Link2Twenty:
> ```js
> data = Object.assign(data, fromMeta);
> ```

[`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) updates the first property so it can just be written 
```js
Object.assign(data, fromMeta);
```

plugins/output/adapt/importsource.js
```diff
@@ -169,14 +169,27 @@ function ImportSource(req, done) {
...
             assetsJson[assetName].tags.forEach(function(tag) {
-              tags.push(tag._id);
+              dbInstance.retrieve('tag', { title: tag.title}, { fields: '_id' }, function(error, results) {
+                if (results && results.length > 0) {
+                  tags.push(results[0]._id);
+                } else {
+                  app.contentmanager.getContentPlugin('tag', function(error, plugin) {
+                    if(!error) {
+                      plugin.create({ title: tag.title}, function(error, record) {
+                        if(error) logger.log('warn', 'Failed to create asset tag: ' + (tag.title || '') + ' ' + error);
```
taylortom: Do we need to return early if an error is returned by the cb? Assuming the next line would fail if the record isn't there.

taylortom:
```suggestion
                        if(error) return logger.log('warn', 'Failed to create asset tag: ' + (tag.title || '') + ' ' + error);
```

PR #2446, opened 2019-10-14
Add courseasset check on clone item, fixes #2347

### LINKED ISSUE #2347: Item copy can result in broken courseassets in other items

dancgray:


frontend/src/modules/scaffold/views/scaffoldListView.js
```diff
@@ -92,9 +94,42 @@ define([
     },
 
     cloneItem: function(event) {
+      var flatItem = Helpers.flattenNestedProperties(this.editor.value);
+      var itemValues = _.values(flatItem);
+      var parentAttributes = Origin.scaffold.getCurrentModel().attributes;
+      _.each(itemValues, function(item) {
```
tomgreenfield: No requirement for Underscore here:
```suggestion
      itemValues.forEach(function(item) {
```

frontend/src/modules/scaffold/views/scaffoldListView.js
```diff
@@ -92,9 +94,42 @@ define([
...
+        if (typeof item === 'string' && item.indexOf('course/assets') !== -1) {
```
tomgreenfield: Suggest returning early:
```js
if (typeof item !== 'string' || item.indexOf('course/assets') === -1) return;
```

frontend/src/core/helpers.js
```diff
@@ -342,6 +342,28 @@ define(function(require){
         '<span class="max-fileupload-size">',
         Origin.l10n.t('app.maxfileuploadsize', {size: Origin.constants.humanMaxFileUploadSize}),
         '</span>'].join(''))
+    },
+
+    flattenNestedProperties: function(properties) {
+      var flatProperties = {};
```
tomgreenfield: Suggest returning early:
```js
if (!properties) return {};
```

PR #2447, opened 2019-10-15
Display error information on plugin upload modal

### LINKED ISSUE #2444: Plugin upload failed modal should be more descriptive
## Expected Behaviour
When I try to upload a plugin and it fails to upload, I should be told the reason why so that I can fix it.

## Actual Behaviour
If I try to upload a plugin and it fails, I see an error modal with 'Plugin upload failed' which is not very useful. 

## Suggested
I suggest we have a more descriptive error modal for the following scenarios:
- The version number is not unique
- The plugin is not compatible with this authoring tool (underlying framework version not compatible)
- The file size of this plugin is too large
- The bower.json is invalid

The extra description can just be a simple sentence shown underneath the Plugin upload failed heading. Any information on this modal would be a big help.

## Versions
  - Framework Version: 4.3.0

canstudios-nicolaw:
Fixes #2444 

- Issue on front end meant that messages sent from the back end were not displayed on the modal, removed unnecessary decodeHTML function to fix this.

- Moved existing error messages to language file so that they can be translated.

### DIFF
--- a/routes/lang/en-application.json
+++ b/routes/lang/en-application.json
@@ -410,5 +410,11 @@
   "app.searchByMail": "Search by email",
   "app.maxfileuploadsize": "Maximum upload file size: %{size}.",
   "app.uploadsizeerror": "File size limit exceeded. Expected no more than %{max}, received %{size}.",
+  "app.fileuploaderror": "File upload failed.",
+  "app.cannotfindbower": "Cannot find expected bower.json file in the plugin root, please check the structure of your zip file and try again.",
+  "app.unrecognisedplugin": "Unrecognized plugin - a plugin should have a bower.json file.",
+  "app.unrecognisedpluginforpackage": "Unrecognized plugin type for package %{package}.",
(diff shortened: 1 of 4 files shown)


routes/lang/en-application.json
```diff
@@ -410,5 +410,11 @@
   "app.searchByMail": "Search by email",
   "app.maxfileuploadsize": "Maximum upload file size: %{size}.",
   "app.uploadsizeerror": "File size limit exceeded. Expected no more than %{max}, received %{size}.",
+  "app.fileuploaderror": "File upload failed.",
+  "app.cannotfindbower": "Cannot find expected bower.json file in the plugin root, please check the structure of your zip file and try again.",
+  "app.unrecognisedplugin": "Unrecognized plugin - a plugin should have a bower.json file.",
+  "app.unrecognisedpluginforpackage": "Unrecognized plugin type for package %{package}.",
```
taylortom: Minor typo
```suggestion
  "app.unrecognisedpluginforpackage": "Unrecognised plugin type for package %{package}.",
```

PR #2452, opened 2019-10-21
Component copy prevented if no location available

### LINKED ISSUE #2320: Component paste function seems to break if there are no available paste zones
## Expected Behaviour
You should either get some sort of feedback if paste isn't available, or be able to paste a component into a new block (the former is an easier and more logical solution). 

## Steps to Reproduce
1. Make sure the page you're on has no available component slots
1. Copy any component
1. Notice nothing happens, but you lose all 'add' buttons, so must refresh the page to add new content

canstudios-scottd:
Fixes #2320

### DIFF
--- a/frontend/src/modules/editor/global/views/editorOriginView.js
+++ b/frontend/src/modules/editor/global/views/editorOriginView.js
@@ -68,12 +68,12 @@ define(function(require){
     },
 
     componentPasteLocation: function() {
-      if ($('.block-inner .add-control').length > 0) { return; }
-        Origin.Notify.alert({
-          type: 'error',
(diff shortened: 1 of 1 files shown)


frontend/src/modules/editor/global/views/editorOriginView.js
```diff
@@ -67,10 +67,20 @@ define(function(require){
       }
     },
 
+    componentPasteLocation: function() {
+      if ($('.block-inner .add-control').length > 0) return;
+      Origin.Notify.alert({
+        type: 'error',
+        text: Origin.l10n.t('app.componentcopyerror')
```
tomgreenfield, line 74: Missing string.

PR #2456, opened 2019-10-23
Fix plugin update

### LINKED ISSUE #2455: Cannot update plugin via the UI
## Expected Behaviour
Plugin updates via UI by clicking the corresponding button.

## Actual Behaviour
Plugin does not update. Tried to update the Language Picker plugin, but nothing was happening and no errors were showing in server log. Did some debugging and it should be throwing the error "There is a plugin already installed with a target attribute of '_languagePicker'." but isn't. From what I can tell, this was probably introduced in #2440  .

## Steps to Reproduce
1. Update a plugin via the UI.

## Versions
  - Authoring Tool Version: 0.10.1
  - Framework Version: 4.3.0
  - Node.js Version: 12.13 LTS
  - MongoDB Version: 4.0.12
  - Operating System: Ubuntu
  - Browser: Chrome

tomgreenfield:
Resolves #2455.

### DIFF
--- a/frontend/src/modules/pluginManagement/views/pluginTypeView.js
+++ b/frontend/src/modules/pluginManagement/views/pluginTypeView.js
@@ -81,18 +81,18 @@ define(function(require){
       });
       $icon.removeClass().addClass('fa fa-refresh fa-spin');
 
-      $.post(this.model.urlRoot + '/update', { 'targets': [this.model.get('_id')] }, _.bind(function(data) {
-        if(!_.contains(data.upgraded, this.model.get('_id'))) {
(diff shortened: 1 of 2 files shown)


dancgray:
@tomgreenfield what's the best way to test this? Tried editing the language picker version number in DB and bower cache. Updating gives console error:

GET: http://localhost:5000/api/extensiontype/5db817a1e3d8fc1809190544 404 (Not Found)

This is the wrong extension id, but back end seems to update the plugin OK.

tomgreenfield:
That's the way I'd recommend, @dancgray. I'd also try out a manual plugin upload too.

Noticed that 404 error while developing – it looks to have been like that a long time before this PR and doesn't affect the update.

PR #2464, opened 2019-11-04
Add array tags in callback 

dancgray:
Stops race condition when importing assets with tags, fixes in #2439

### DIFF
--- a/plugins/output/adapt/importsource.js
+++ b/plugins/output/adapt/importsource.js
@@ -185,7 +185,7 @@ function ImportSource(req, done) {
             createdBy: app.usermanager.getCurrentUser()._id
           };
 
-          if (!assetJson) helpers.importAsset(fileMeta, metadata, doneAsset);
+          if (!assetJson) return helpers.importAsset(fileMeta, metadata, doneAsset);
 
           addAssetTags(assetJson, function(error, assetTags) {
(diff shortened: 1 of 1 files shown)


plugins/output/adapt/importsource.js
```diff
@@ -203,15 +184,38 @@ function ImportSource(req, done) {
             repository: repository,
             createdBy: app.usermanager.getCurrentUser()._id
           };
-          if(!fileMeta) {
-            return doneAsset(new helpers.ImportError('No metadata found for asset: ' + assetName));
-          }
-          helpers.importAsset(fileMeta, metadata, doneAsset);
+
+          if (!assetJson) helpers.importAsset(fileMeta, metadata, doneAsset);
```
tomgreenfield: Might need a `return` here to prevent the rest of the code running?

canstudios-nicolaw:
```suggestion
          if (!assetJson) {
            helpers.importAsset(fileMeta, metadata, doneAsset);
            return;
          }
```

canstudios-nicolaw: I have just tested this with courses that do and don't have an assets.json file, this works for me.

@tomgreenfield was correct, the original lack of return here caused problems if an asset.json wasn't present

PR #2488, opened 2020-01-09
Clear build folder on export and publish

### LINKED ISSUE #2442: If I switch theme, the old theme assets still seem to be retained on publish
## Expected Behaviour
If I switch course theme then the theme assets for the current theme should be included in publish.

## Actual Behaviour
I have a theme that contains a lot of assets. If I create a course with Vanilla and publish it then the size of the publish is 645kB. If I then switch to my big theme and publish the download is 8.7MB. 

If I then switch back to vanilla and publish the download remains at 8.7MB. It's like it hasn't cleared the assets from my theme.

## Steps to Reproduce
1. Create basic course with vanilla
1. Publish and note download size
1. Switch to a theme which uses lots of assets
1. Publish and note increased download size
1. Switch back to vanilla
1. Publish and note download size hasn't decreased

## Versions
  - Framework Version: 4.3.0

canstudios-nicolaw:
Fixes #2442

### DIFF
--- a/plugins/output/adapt/publish.js
+++ b/plugins/output/adapt/publish.js
@@ -94,6 +94,7 @@ function publishCourse(courseId, mode, request, response, next) {
         }
 
         if (mode === Constants.Modes.Export || mode === Constants.Modes.Publish) {
+          fs.emptyDirSync(BUILD_FOLDER);
           isRebuildRequired = true;
           return callback(null);
         }
--- a/plugins/output/adapt/publish.js
+++ b/plugins/output/adapt/publish.js
@@ -29,11 +29,11 @@ function publishCourse(courseId, mode, request, response, next) {
   let resultObject = {};
 
   // shorthand directories
-  let FRAMEWORK_ROOT_FOLDER = path.join(configuration.tempDir, configuration.getConfig('masterTenantID'), Constants.Folders.Framework);
-  let SRC_FOLDER = path.join(FRAMEWORK_ROOT_FOLDER, Constants.Folders.Source);
-  let COURSES_FOLDER = path.join(FRAMEWORK_ROOT_FOLDER, Constants.Folders.AllCourses);
-  let COURSE_FOLDER = path.join(COURSES_FOLDER, tenantId, courseId);
-  let BUILD_FOLDER = path.join(COURSE_FOLDER, Constants.Folders.Build);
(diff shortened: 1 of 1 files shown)


tomgreenfield:
Yet to give this a test but does/can this happen on a 'force rebuild' preview as well?

canstudios-nicolaw:
You're right @tomgreenfield it does happen on preview (force or regular) as well. 

In fact I can't think of a situation in which we wouldn't want the build folder to be emptied. What do you think of moving that line out of the if statement? (Or just having this as it's own waterfall step?)

tomgreenfield:
Not really sure about the normal, 'non-rebuild' preview - might be easier to leave that process untouched.

plugins/output/adapt/publish.js
```diff
@@ -15,26 +15,27 @@ const origin = require('../../../');
...
-  var FRAMEWORK_ROOT_FOLDER = path.join(configuration.tempDir, configuration.getConfig('masterTenantID'), Constants.Folders.Framework);
-  var SRC_FOLDER = path.join(FRAMEWORK_ROOT_FOLDER, Constants.Folders.Source);
-  var COURSES_FOLDER = path.join(FRAMEWORK_ROOT_FOLDER, Constants.Folders.AllCourses);
-  var COURSE_FOLDER = path.join(COURSES_FOLDER, tenantId, courseId);
-  var BUILD_FOLDER = path.join(COURSE_FOLDER, Constants.Folders.Build);
+  let FRAMEWORK_ROOT_FOLDER = path.join(configuration.tempDir, configuration.getConfig('masterTenantID'), Constants.Folders.Framework);
+  let SRC_FOLDER = path.join(FRAMEWORK_ROOT_FOLDER, Constants.Folders.Source);
+  let COURSES_FOLDER = path.join(FRAMEWORK_ROOT_FOLDER, Constants.Folders.AllCourses);
+  let COURSE_FOLDER = path.join(COURSES_FOLDER, tenantId, courseId);
+  let BUILD_FOLDER = path.join(COURSE_FOLDER, Constants.Folders.Build);
```
tomgreenfield: These are `const`s really.

canstudios-nicolaw:
@tomgreenfield I think it would be worth you approving again since I've committed since the last approval

PR #2489, opened 2020-01-09
Move course validation to back-end for efficiency

### LINKED ISSUE #2487: Heavy course validation on front-end
On every preview/export/download, the front-end is running a function called `validateCourseContent`. This function is looping through all pages/articles/blocks/components in a course to ensure that they all have children, each item is fetched from the back-end individually. This is inefficient as it can generate hundreds of requests and cause long delays (8+ seconds) on larger courses.

Remove the `validateCourseContent` function and replace it with back-end validation that happens as part of preview/export/publish. The user should still get an error modal with the same information if the course is invalid, but this should be a more efficient approach as we will have access to the db.

canstudios-nicolaw:
Fixes #2487


plugins/output/adapt/outputHelpers.js
```diff
@@ -277,6 +277,43 @@ function getPluginFrameworkVersionCategory(serverFrameworkVersion, pluginMetaDat
...
+function validateCourse(data, cb) {
+  let errors = '';
+  let contentObjects = data.contentobject;
+  let articles = data.article;
+  let blocks = data.block;
+  let components = data.component;
+
+  if (typeof contentObjects === 'undefined') {
+    let courseString = app.polyglot.t('app.course').charAt(0).toUpperCase() + app.polyglot.t('app.course').slice(1);
+    errors += courseString + ' "' + data.course[0].title + '" ' + app.polyglot.t('app.doesnotcontain') + ' ' + app.polyglot.t('app.page') + 's\n';
```
tomgreenfield: Localising pluralised words is unfortunately not as easy as appending an 's'. For example, in French the singular form of courses is _cours_.

The way round it in polyglot is by providing both forms in the language file. For example:
```js
const courseString = app.polyglot.t('app.courses', 1);
errors += app.polyglot.t('app.doesnotcontain', {
  type: courseString[0].toUpperCase() + courseString.slice(1),
  title: data.course[0].title,
  childType: app.polyglot.t('app.pages', 0)
});
```
with language strings:
```json
"app.courses": "course |||| courses",
"app.pages": "page |||| pages",
"app.doesnotcontain": "%{type} '%{title}' does not contain any %{childType}"
```

canstudios-nicolaw: Re testing, I tested all the combos, eg block without component, article without block etc. just to check the error messages. Only slightly weird one was if you have an empty menu the message was 

Menu "Section title" does not contain any articles

But I didn't add a special case to change articles to pages because it seemed self explanatory (and you can't add an article without a page anyway).

Great point about the string localisation, I will add this in!

Thanks for prompt :eyes: btw

canstudios-nicolaw: Updated - I'll leave this conversation for you to resolve if you're happy

plugins/output/adapt/outputHelpers.js
```diff
@@ -277,6 +277,63 @@ function getPluginFrameworkVersionCategory(serverFrameworkVersion, pluginMetaDat
...
+  parents.forEach(parent => {
+    let parentType = app.polyglot.t('app.' + parent._type, 1);
+    let childType = app.polyglot.t('app.children');
+
+    if (typeof children === 'undefined') {
+      errors += app.polyglot.t('app.doesnotcontain', {
+        type: parentType[0].toUpperCase() + parentType.slice(1),
+        title: parent.title,
+        childType: childType
+      }) + '\n';
```
tomgreenfield: Since the new code is identical to lines 327–331, suggest storing in a variable that can be called:
```js
const appendError = () => {
  errors += app.polyglot.t('app.doesnotcontain', {
    type: parentType[0].toUpperCase() + parentType.slice(1),
    title: parent.title,
    childType: childType
  }) + '\n';
};
```

canstudios-nicolaw:
I'm not going to get a chance to work on this for a little while. I think it's a good suggestion but it would be worth releasing this as-is for now.

Issue #2498, opened 2020-02-04
It should be possible to run the upgrade script and skip installing latest node dependencies


canstudios-nicolaw:
Each time the upgrade script is run it installs the latest node dependencies for both the authoring tool and the framework. 

It would be good if the upgrade script had the option to run without updating the node dependencies to reduce the testing time for managed systems.

There should be a new flag which can be passed to the upgrade script which skips the dependencies install. If the flag is not specified then the upgrade script will continue to work as it does now.

---

The installHelpers.js file contains a function `installDependencies` which is run for both the AT and FW. It removes the node_modules folder and runs npm install. I would hope it would be a case of skipping this function call if the option is set on the upgrade script. This may be an 'advanced' option - it is up to the person running the script to determine whether an npm install is required (eg new dependencies have been added) or not.

tomgreenfield:
I believe this has been made redundant by #2505.

PR #2303, opened 2019-04-09
Spring clean NPM dependency list

taylortom:
Fixes #2301

PR #2500, opened 2020-02-06
Maintaining Dependencies in Adapt

### LINKED ISSUE #2301: Generate a dependency lock on release
As a part of each release, we should generate an updated `package-lock.json` to reduce any module compatibility issues.

canstudios-andrews:
Fixes #2301 makes #2498 redundant
The aim of this Pull request is to reduce the number of incidents where updates to dependencies in the Authoring Tool create issues for users and partners. I've taken the work done by Tom Taylor on #2303 and updated it for the current Authoring Tool source, if this Pull Request is adopted the community would need update the dependencies with each release.

Note: the install script would also obey a lock file if it was added to the Adapt Framework source.

To test this PR run the install and upgrade scripts and observe that the dependencies follow the committed lockfile and the tool works as designed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,6 @@
 /nbproject/
 /master/
 
-package-lock.json
-
 /plugins/content/component/versions/
 /plugins/content/component/componentcache/
 /plugins/content/extension/versions/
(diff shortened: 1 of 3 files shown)


tomgreenfield:
Superseded by #2505.

PR #2501, opened 2020-02-11
Release 0.10.3

tomgreenfield:
PR for 0.10.3.


plugins/output/adapt/outputHelpers.js
```diff
@@ -277,6 +278,63 @@ function getPluginFrameworkVersionCategory(serverFrameworkVersion, pluginMetaDat
...
+  errors += iterateThroughChildren(contentObjects, articles);
+  errors += iterateThroughChildren(articles, blocks);
+  errors += iterateThroughChildren(blocks, components);
+
+  if (errors.length !== 0) return cb(errors, false);
+
+  return cb(null, true);
+}
+
+function iterateThroughChildren(parents, children) {
```
dancgray: This seems to be getting false positives. Getting the following error on menus:
`Menu 'menu' does not contain any articles`

canstudios-nicolaw: (Just so I'm clear) Does the menu have any pages in it? 

I did mention the error message for empty menus being slightly strange in https://github.com/adaptlearning/adapt_authoring/pull/2489#discussion_r364819998, but I think we said it was ok because pages will be added in order to add articles.

dancgray: Yes, the menu had a child page. The child page had a valid article, block and component.

canstudios-nicolaw: This sounds like a show-stopper. I'm not able to test this right now. 

@tomgreenfield can you confirm if this is something you have noticed on your 0.10.3 server? 

Do we know if this happens every time or only some of the time?

dancgray: It's every time for me, tried new course and import. All have sub menus with valid pages in the submenu. 

Used a clean install of release/bugpatch branch, node v12.11.1, mongo v3.6.12, Firefox 72.

tomgreenfield: Can replicate this, @canstudios-nicolaw.

tomgreenfield:
Ready for re-review @dancgray & @canstudios-nicolaw; planning to leave open till Friday to give it a run on our production server.

canstudios-nicolaw:
Thanks for sorting this @tomgreenfield. It wouldn't be an AT release without updating the release date in the Changelog at least once!

Issue #2504, opened 2020-02-18
Error: Failed to determine user's tenant!


tomgreenfield:
## Problem

The domain session code has gone awry. The culprit this time seems to be the 5.9.x line of the Mongoose package.

## Temporary workround
```console
npm install mongoose@5.8.13
```

## Stack trace
```console
Error: Failed to determine user's tenant!
    at Object.getDatabase (D:\_delete_adapt_authoring\lib\database.js:416:19)
    at D:\_delete_adapt_authoring\plugins\content\tag\index.js:76:16
    at D:\_delete_adapt_authoring\lib\rest.js:115:7
    at Layer.handle [as handle_request] (D:\_delete_adapt_authoring\node_modules\express\lib\router\layer.js:95:5)
    at next (D:\_delete_adapt_authoring\node_modules\express\lib\router\route.js:137:13)
    at Route.dispatch (D:\_delete_adapt_authoring\node_modules\express\lib\router\route.js:112:3)
    at Layer.handle [as handle_request] (D:\_delete_adapt_authoring\node_modules\express\lib\router\layer.js:95:5)
    at D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:281:22
    at Function.process_params (D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:335:12)
    at next (D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:275:10)
    at Layer.handle [as handle_request] (D:\_delete_adapt_authoring\node_modules\express\lib\router\layer.js:91:12)
    at trim_prefix (D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:317:13)
    at D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:284:7
    at Function.process_params (D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:335:12)
    at next (D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:275:10)
    at D:\_delete_adapt_authoring\lib\permissions.js:495:16
    at Layer.handle [as handle_request] (D:\_delete_adapt_authoring\node_modules\express\lib\router\layer.js:95:5)
    at trim_prefix (D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:317:13)
    at D:\_delete_adapt_authoring\node_modules\express\lib\router\index.js:284:7
[...]

PR #2505, opened 2020-02-18
Lock dependencies to known working versions

tomgreenfield:
This PR adds `package-lock.json` to ensure new installations pull down known working versions of npm packages.

This targets Mongoose 5.8.13 until #2504 is investigated and resolved.

Please test on a fresh install with no existing `package-lock.json`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,6 @@
 /nbproject/
 /master/
 
-package-lock.json
-
 /plugins/content/component/versions/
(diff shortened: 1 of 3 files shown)


canstudios-nicolaw:
This is very similar to https://github.com/adaptlearning/adapt_authoring/pull/2500. How would you feel about adding the npm ci change so that we can kill two birds with one stone?

PR #2525, opened 2020-07-14
Remove build includes after build

### LINKED ISSUE #2510: Export fails as it requires LESS variables for plugins that aren't in the project
## Expected Behaviour

When you export a course it shouldn't fail because it can't find a LESS variable not used in the project.

## Actual Behaviour

An exported course looks through all plugins installed in the Authoring Tool. This means that the export fails if it can not find the LESS variables required for these plugins which may not have been created in the theme.

Preview and Publish aren't affected by this (they must only look at the installed plugins as intended)

## Steps to Reproduce
1.  Add a course using certain plugins with a theme that supports these plugins.
1. Add a second course that doesn't use the above plugins, with a separate theme that doesn't have the LESS variables required for these plugins.
1. Attempt to export the second course.

## Versions
  - Authoring Tool Version: 0.10.3
  - Framework Version: 5.2.0

tomgreenfield:
As a side effect of #1586 & #1664, exports attempt to build with _all_ plugins bundled.

This PR ensures that build includes are always generated so that the grunt process uses the plugins relevant to the course. The build object is then removed from the config JSON after the build has completed.

Resolves #2510.

### DIFF
--- a/lib/outputmanager.js
+++ b/lib/outputmanager.js
@@ -266,9 +266,6 @@ OutputPlugin.prototype.sanitizeCourseJSON = function(mode, json, next) {
 
     async.waterfall([
       function(callback) {
-        if (mode === Constants.Modes.Export) {
-          return callback();
-        }
         self.generateIncludesForConfig(configJson, function(error, includes) {
           if (error) {
             return callback(error);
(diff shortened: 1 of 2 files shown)


deltanetdan:
@tomgreenfield does this require framework v5 to reproduce the issue?

tomgreenfield:
Morning @deltanet, attempting to export with an older plugin installed alongside framework 5 is the most common real-world case.

Equally, you could also upload a plugin with a fake Less variable that isn't defined anywhere.

PR #2533, opened 2020-08-06
Populate default globals on import

tomgreenfield:
Resolves https://github.com/adaptlearning/adapt_framework/issues/2841.


deltanetdan:
@tomgreenfield a lot of attributes are lost from the config collection on import with this. Such as _enabledExtensions and  _extensions.

Issue #2540, opened 2020-10-21
Users can change the passwords of other users


simondate:
It seems that when you create additional accounts on the latest version of the Authoring Tool they get shown the email address, first name and last name of another account. They can also change the password of this user!

## Steps to Reproduce

1. Admin account creates a new account
1. Log into the new account
1. Click on `logged in as: [new account]` to change the password
1. **The user is shown the details of another admin account.**
1. Change the password of the other admin account
1. The admin account cannot log in with their own password, but they can use this new password

## Versions
  - Authoring Tool Version: 0.10.4
  - Framework Version: 5.7.1

deltanetdan:
I can reproduce this issue. Reversing [this](https://github.com/adaptlearning/adapt_authoring/pull/2516/) PR seems to fix the problem.

moloko:
nice find @simondate

moloko:
Looks like it gives you the account details of whoever created the account. Currently having lots of fun changing @guywillis password to random things ;-)

PR #2541, opened 2020-10-22
Reintroduce fixed Mongoose dependency

tomgreenfield:
The PR reintroduces the `mongoose@5.8.13` dependency, which in turn pulls down the last version of `mongodb` that's known to work with the domain/session code in the authoring tool. Reference: #2504.

This does surface the following warnings in recent versions of Node.js:
```console
(node:20860) Warning: Accessing non-existent property 'count' of module exports inside circular dependency
(node:20860) Warning: Accessing non-existent property 'findOne' of module exports inside circular dependency
(node:20860) Warning: Accessing non-existent property 'remove' of module exports inside circular dependency
(node:20860) Warning: Accessing non-existent property 'updateOne' of module exports inside circular dependency
```
However, this seems like an adequate compromise having already done a sizeable chuck of debugging to try and ascertain why the domain goes missing in the middleware with later versions of `mongodb`.

This prevents the 'sticky session' problem detailed in #2540 if logging in to different user accounts using the same device (and browser and OS account).

Tested in Node 12 & 14.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4115,15 +4115,15 @@
       "integrity": "sha1-D3ca0W9IOuZfQoeWlCjp+8SqYYE="
     },
     "mongoose": {
-      "version": "5.9.18",
-      "resolved": "https://registry.npmjs.org/mongoose/-/mongoose-5.9.18.tgz",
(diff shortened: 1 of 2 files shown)


deltanetdan (approved):

taylortom (approved):
worksforme :tada:

deltanetdan (approved):

PR #2582, opened 2021-12-09
Run grunt using npx

### LINKED ISSUE #2581: Remove global grunt dependency
Everything can (and should) be built using the locally installed grunt dependency and not require grunt to be installed globally. All grunt commands should be run via npx.

taylortom:
Fixes #2581

Using npx to run a local copy of grunt. Means we no longer need the global grunt CLI dependency.

### Note that this is a breaking change for the end-user, as `npx grunt build:prod` etc. will be needed to build the UI.


oliverfoster (approved):

canstudios-nicolaw (approved):
:eye:

PR #2591, opened 2021-12-14
Support --skipDependencyCheck during install process

### LINKED ISSUE #2590: Support --skipDependencyCheck during install process
This is currently only used by the app startup process, but would be good to be able to use for the install

taylortom:
Fixes #2590


taylortom:
Merging so can be tested with release

PR #2595, opened 2021-12-14
Remove RequireJS timeout limit

### LINKED ISSUE #2594: RequireJS occasionally times out on very slow connections
See [config-waitSeconds](https://requirejs.org/docs/api.html#config-waitSeconds) in the RequireJS docs.

taylortom:
Fixes #2594

### DIFF
--- a/frontend/src/core/config.js
+++ b/frontend/src/core/config.js
@@ -90,5 +90,6 @@ require.config({
     'templates/templates': {
       deps:['handlebars']
     }
-  }
+  },
+  waitSeconds: 0
 });


oliverfoster (approved):

canstudios-nicolaw (approved):
:eye:

clara-am:
Hi there, 
I'm having a similar issue, but I have doubts. Is this adding the waitSeconds, inside the scriptLoader.js file, when you publish the package? Or should it be inside adapt-framework code?
Currently, I'm including that line, after publishing, and with that I fix the problem. But I would like to find a solution to do that process automatically.

oliverfoster:
No, this is for the AAT user interface. We can do similar in the framework though.

oliverfoster:
https://github.com/adaptlearning/adapt-contrib-core/pull/81

PR #2636, opened 2022-08-11
Updated the Gruntfile to include a uglify task

koenignicolas:
Updated the Gruntfile to include a uglify task as a separate step and excluded
the uglify in the compilation of requirejs step
To test you need to include a frontend plugin that
uses es6. 

Since the goal of the change is to enable us to use ES6 in the frontend plugins of the authoring tool.


taylortom:
We actually already had similar functionality awaiting review. See https://github.com/adaptlearning/adapt_authoring/pull/2645

We're hoping to roll this into a release very soon

PR #2645, opened 2022-09-07
Add babel for ES6 support

taylortom:
This is one of Tom G's that's been lying around for a while. Would be good to get this into legacy.

Adds babel so we can use ES6 in the front-end code.


deltanetdan (approved):
👀

PR #2654, opened 2022-09-08
Replace migration mechanism

### LINKED ISSUE #2653: Fix migrations
The migrations package we use is unsupported now, and has broken at some point with our current dependency tree. It needs replacing asap.

taylortom:
Fixes #2653 

Unfortunately I've had to rip out the old migration mechanism due to incompatibility issues. I've added a new npm task (`create-migration`) to set up new migrations, everything else is run internally, so nothing changes.

All existing migrations have been updated to use the new module (and renamed for consistency).


deltanetdan:
@taylortom if I install using this branch (issue/2653)) I get the following error after "Web application built successfully" log:

```
ERROR:  Error: Migrator#runTask expects a callback parameter
    at runMigrations (C:\Users\dan.gray\Documents\Source\Adapt\local\adapt-test-final\at-migrate\lib\installHelpers.js:317:11)
```

Tried with node v14.19.3 and v16.15.1.

PR #2752, opened 2024-02-20
Fix: ckeditor bump to v4.17.0 (fixes #2751)

### LINKED ISSUE #2751: CK Editor not loading correctly
## Expected Behaviour

CK Editior is not loading correctly meaning that 'TextArea' fields are not interactable

image

[Also reported here in Gitter](https://app.gitter.im/#/room/#adaptlearning_adapt_authoring:gitter.im/$JtH23jiAlrGPYujLXUFMqBrbFC_GtDpCSI1XJgFwZ1M)

Issue seemed to have been introduced in AAT 0.11.2. Likely an issue with a npm dependecy.

Issue can be fixed by just refreshing the page, but it's not really a great experience for the user!

oliverfoster:
fixes #2751

### Fixes
* Bump version of ckeditor from v4.16.2 to v14.17.0 
* Wait for textarea to become attached to the dom before replacing with ckeditor

### DIFF
--- a/frontend/src/modules/scaffold/backboneFormsOverrides.js
+++ b/frontend/src/modules/scaffold/backboneFormsOverrides.js
@@ -66,7 +66,20 @@ define([
   Backbone.Form.editors.TextArea.prototype.render = function() {
     textAreaRender.call(this);
 
-    _.defer(function() {
+    function until(conditionFunction) {
+      const poll = resolve => {
+        if(conditionFunction()) resolve();
+        else setTimeout(_ => poll(resolve), 400);
+      }
+      return new Promise(poll);
+    }
+    function isAttached($element) {
+      return function() {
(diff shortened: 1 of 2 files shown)


nachocinalli (approved):

nachocinalli (approved):

joe-allen-89 (approved):
Struggling to replicate the original issue, but the PR is working as expected.

oliverfoster:
> Struggling to replicate the original issue, but the PR is working as expected.

I can only replicate on a new mcq. I imagine that any long schema will cause the _.defer to lapse before the rendered template is attached to the Dom, where ckeditor errors replacing an unattached textarea. The fix is to perform the ckeditor replacement of the textarea only after the parent element is attached to the body element.

PR #2773, opened 2024-07-09
Update: From ckeditor4 4.17.0 to ckeditor5 42.0.0 (fixes #2772)

oliverfoster:
fixes #2772

Updated from ckeditor4 4.17.0 to ckeditor5 42.0.0 

image

PR #2780, opened 2024-08-15
Enables CKeditior html elements

simondate:
## Proposed changes

Adds support for inline HTML elements such as `class` `span` `style`.

Decided to create as a PR rather than commit straight to on the branch as I wasn't sure about the approach I used. This [article](https://ckeditor.com/docs/ckeditor5/latest/features/html/general-html-support.html#enabling-all-html-features) recommends only allowing certain elements to prevent a security risk.

The attribute previously used was set to `span(*)` . I'm fairly certain this wasn't working properly as other values were working such as `class`.

Ref: https://ckeditor.com/docs/ckeditor5/latest/features/html/general-html-support.html

config.json:
```json
{
  "ckEditorHtmlSupport": {
    "allow": [
      {
        "name": "^(blockquote|ul|ol|span)$"
      },
      {
        "name": ".*",
        "classes": true,
        "styles": true
      },
      {
        "name": "^a$",
        "attributes": {
          "target": true
        }
      }
    ]
  }
}
```

Adds support for ckeditor htmlSupport:
```
        htmlSupport: {
            allow: [ /* HTML features to allow. */ ],
            disallow: [ /* HTML features to disallow. */ ]
        }
```
Strings will be processed as regex body values, sub objects and arrays are allowed, booleans are supported:
```
[
    {
        // The element name to enable or extend with
        // the following styles, classes, and other attributes.
        name: "regexp",

        // Styles to allow (by name, name and value, or just all).
        styles: "regexp"|boolean,

        styles: {
            name: "regexp"|boolean
        }.

        // Classes to allow (by name or just all).
        classes: "regexp"|boolean,

        classes: [ "regexp", "regexp" ],

        // Other attributes to allow (by name, name and value, or just all).
        attributes: "regexp"|boolean

        attributes: {
            name: "regexp"|boolean
        }
    }
]
```


frontend/src/modules/scaffold/backboneFormsOverrides.js
```diff
@@ -90,7 +90,16 @@ define([
...
         enterMode: CKEDITOR[Origin.constants.ckEditorEnterMode],
         entities: false,
-        extraAllowedContent: Origin.constants.ckEditorExtraAllowedContent,
+        htmlSupport: {
+          allow: [
+            {
+              "name": new RegExp(Origin.constants.htmlSupportAllowed.name),
+              "attributes": Origin.constants.htmlSupportAllowed.attributes,
+              "classes": Origin.constants.htmlSupportAllowed.classes,
+              "styles": Origin.constants.htmlSupportAllowed.styles
```
oliverfoster, line 99: I don't understand where you're getting the sub properties of this object from `Origin.constants.htmlSupportAllowed.*` `name, attributes, classes, styles` when you've defined the value as a string `htmlSupportAllowed: '.*',`

image

simondate: Sorry, overlooked this. This file is only used when installing a new AAT. I was using my own `conf/config.json` file, which doesn't get added to git.

lib/configuration.js
```diff
@@ -81,7 +81,12 @@ function Configuration() {
   this.conf = {
     root: this.serverRoot,
     maxLoginAttempts: 3,
-    ckEditorExtraAllowedContent: 'span(*)',
+    htmlSupportAllowed: {
+      name: ".*",
+      attributes: true,
+      classes: true,
+      styles: true
+    },
```
oliverfoster, line 89: Is this intentional? It looks like you've completely ignored the security section you've mentioned in the pr body.

https://ckeditor.com/docs/ckeditor5/latest/features/html/general-html-support.html#security

simondate: I wanted to get in place a working PR, then could have a discussion with you and other developers about what would be a sensible default.

I guess we could at the minimum restrict the `script` tag?

oliverfoster: Generally it's best to go with a permissive, rather than a restrictive attitude to things like this. The list of potential things to restrict is always hard to determine, changes over time and tends to be more insecure as time passes.

simondate: We'll need everything currently visible in the UI (`blockquote` `ul` `ol`) assuming they weren't working already.

Then `class` `span` `style` comes to mind. I've worked with teams who've done some weirder things, but I guess we should make it so they have to "opt in" those things themselves.

Finally my client mentioned that by default with [PR 2773](https://github.com/adaptlearning/adapt_authoring/pull/2773), that links were opening in new tabs rather than former standard behaviour of opening in the same tab.

I noticed from testing this PR that issue was fixed, so I'm wondering if the attribute `target` was getting removed. If so, we should also add it to the whitelist.

oliverfoster: I have a minor update to support these things, would you mind if i commit them?

simondate: Gor for it!

oliverfoster: I think, there's probably an issue with restricting attribute values with the way I've done this.

oliverfoster:
I think this is good now. Could you test please?

hanshillen:
I would like the following to be allowed (or be able to override the provided CKE instance with a custom configuration without having to modify the AAT's source): 

* Be able to add inline code (part of [basic text styles](https://ckeditor.com/docs/ckeditor5/latest/features/basic-styles.html) feature ) and code blocks ([Code blocks](https://ckeditor.com/docs/ckeditor5/latest/features/code-blocks.html) feature). Ideally combined with the [Auto formatting](https://ckeditor.com/docs/ckeditor5/latest/features/autoformat.html) feature, which lets you type `` ` `` or ```` ``` ```` to automatically start these elements.
* Custom classes ([Styles](https://ckeditor.com/docs/ckeditor5/latest/features/style.html) feature)
* sup, super (part of [basic text styles](https://ckeditor.com/docs/ckeditor5/latest/features/basic-styles.html))
* kbd
* mark ([Highlight](https://ckeditor.com/docs/ckeditor5/latest/features/highlight.html) feature)
* cite
* Ability to automatically add `target="_blank"` to external links (e.g. like in this [link decorator demo](https://ckeditor.com/docs/ckeditor5/latest/features/link.html#demo-2))

simondate:
@hanshillen sorry for the delay. I just had time to sit down with this again and look through your suggestions.

Some of your requests look really cool, but since the legacy branch is meant to have ceased development, I would be hesitant to adding any new features, this PR was just to ensure that functionality that worked before continues to.

`sup` and `super` are present in the editor and everything else is possible to do (in some format) by editing the source code.

The only thing I'm not particularly happy with is the link editor, as it's much more basic than the old one and doesn't allow `target="_blank"`, but as long as the author is taught how to edit the source, it's possible to add it there.

Therefore @oliverfoster , I think we should just focus on merging this as is. I'm going to test this with a fresh install and see it works as is and everything that was possible in earlier CK editior versions still works.

oliverfoster (approved):

Issue #2815, opened 2025-02-12
CKEditor toolbar element 'RemoveFormat' not loaded in


simondate:
We list the RemoveFormat as a toolbar element, but it's not actually imported. This leads to a warning error. I would imagine this feature would be good to have

mrblonde77:
We have updated our system and now also have this problem, so I would be very interested in a solution.

mrblonde77:
Just want to mention, that you can not format / style a table in the CKeditor, without this Toobar element.
Any plan to check this problem?

Issue #2814, opened 2025-01-16
WYSIWYG editor stripping out html


fosterc1:
The WYSIWYG editor appears to be stripping out HTML added within text boxes.

FW: v5.42.1 / AT: v0.11.5

When entering html text within the Source of the WYSIWYG editor and then saving the HTML code is being removed.

```html
As an example, we've added the following in the Body text in Project settings and within a text component:
<p>Welcome!We are committed to creating an environment where equity, diversity and inclusion (ED&amp;I) are non-negotiable, allowing our unique and authentic team members to bring their whole selves to work and feel a real sense of belonging. We’re on an ED&amp;I journey, and this learning experience is just the start.</p><p>Throughout this experience, you’ll have opportunities to reflect on what ED&amp;I means for you. When prompted, add your thoughts to your personal diary in the top right of the screen using the<span class="icon icon-checklist inline-icon" aria-hidden="true"></span>icon. You’ll be asked to share some of your thoughts and learning with your line manager or team leader at your next catch-up with them.</p>

After saving the editor stips out the HTML <span class="icon icon-checklist inline-icon" aria-hidden="true"></span>
```

We have tried selecting Source adding the text and saving and selecting Source adding the text, selecting Source again and saving, nothing seems to be preserving the HTML code.

oliverfoster:
This is an authoring tool issue, moving.

oliverfoster:
~~There is no html in the example you've given @fosterc1 could you be more specific please.~~ Have wrapped the example in appropriate tags.

oliverfoster:
@simondate are you able to add and edit the same html in your aats?

simondate:
None of my customers have mentioned any issues.

Looking at Caleb's code, I wonder if the `aria-hidden` is the thing that is being backlisted?

StuartNicholls:
It looks as though the WYSIWYG is stripping out empty tags, so the example above gets removed for that reason. Adding something inside the tag helps, but as @simondate mentioned it looks like `aria-hidden` is ALSO blacklisted as that gets removed even with something inside the `<span>` tag.

@taylortom has something changed in the last few months as the editor didn't quite have this issue previously?

ambivalent-axiom:
This is specifically in Authoring tool, CKEditor has its security settings and strips most of the HTML unless You define otherwise in config.
```nano [path to...]/adapt_authoring/conf/config.json```
find and edit:
Example:
```  "ckEditorHtmlSupport": {
[...]

PR #2816, opened 2025-02-12
Fixes 2814, 2815

simondate:
Fixes -

* #2814 
* #2815 

Two issues with CKEditor.


simondate:
Just updated this PR as it was still stripping the `aria-hidden="true"` attribute.

Note for testers - You'll need to update the `conf/config.json` file, to include this -

```
 "ckEditorHtmlSupport": {
    "allow": [
      {
        "name": "^(blockquote|ul|ol|span)$",
        "attributes": true,
        "classes": true,
        "styles": true
      },
      {
        "name": ".*",
        "classes": true,
        "styles": true
      },
      {
        "name": "^a$",
        "attributes": {
          "target": true
        }
      }
    ],
    "allowEmpty": [
      "span", 
      "i"
    ]
  },
  ```

File: frontend/src/modules/scaffold/backboneFormsOverrides.js [hunk 1]
```diff
@@ -107,6 +107,7 @@ define([
         htmlSupport: {
           // Convert all allow/disallow strings to regexp, as config is json only
           allow: convertStringsToRegExDeep((Origin.constants.ckEditorHtmlSupport && Origin.constants.ckEditorHtmlSupport.allow) || []),
+          allowEmpty: ['span', 'i'],
```
StuartNicholls: Hey @simondate, I'm curious as to why we're allowing the <I> tag here?

simondate: What do you mean? Do you think we should also expose options here in `config/config.json`?

I guess that will make this release slightly more future proof? As if in the future more attributes are added we can extend it further there.

StuartNicholls: Sorry, that was supposed to say, `<i>` tag, rather than making the rest of the line italicised!

So was just asking why HTML italic tag is there.

simondate: This was just in the example that I copied from CKEditior's website. I know that some other icon packs such as Font Awesome use `<i>`, so perhaps someone else may find it useful.

simondate:
Just discovered that the any <a> links to Glossary objects also weren't working.

The latest commit fixes this too.

StuartNicholls (approved):

lib/configuration.js
```diff
@@ -83,7 +83,10 @@ function Configuration() {
     maxLoginAttempts: 3,
     ckEditorHtmlSupport: {
       allow: [{
-        name: "^(blockquote|ul|ol|span)$"
+        name: "^(a|blockquote|ul|ol|span)$",
+        "attributes": true,
```
oliverfoster, line 87: Won't this allow `onclick` attributes on all of the aforementioned tags? Which means that one AAT user could add JavaScript that another user could execute?

StuartNicholls: This was added to allow aria-hidden, if we think the above is a risk then we can add specific attributes. I believe like this:

`attributes: { foo: 'true', bar: true }`

oliverfoster: It definitely is a risk if it allows all attributes.

StuartNicholls: @simondate , so allow, aria-hidden, aria-label?

simondate: I would want `data-glossaryterm` included as it's the way you [open glossary items](https://github.com/adaptlearning/adapt-contrib-glossary?tab=readme-ov-file#usage) within elements on the page.

If `onclick` is the only thing we're concerned about could we just disable that?

oliverfoster: I wouldn't go with disabling things. Something new is always added that makes denies redundant. Permissive allows are generally a better idea (as per the French vs British legal system).

How about we don't set the default config to allow/deny extra things that are specific to a vendor, but we instead just add `allowEmpty` and `RemoveFormat` to support the given use-case?

simondate: `allowEmpty` would fix the original issue, but I feel the glossary extension should be supported, considering it's a Core extension

oliverfoster: Does this work?

```js
      {
        "name": ".*",
        "attributes": "^(id|role|aria-.+|data-.+|target|href)$",
        "classes": true,
        "styles": true
      },
```

In theory that should allow all `aria-` and `data-` prefixed attributes as well as `role` and `id`.

mrblonde77:
Any update on this topic? We switched to the current Authoring Tool and Framework Version, and now we have problems with tables in the ckeditor. I assume that it is related to this problem: CKEditor toolbar element 'RemoveFormat' not loaded" #2815

lib/configuration.js
oliverfoster: bump

simondate: I've been using this PR for months with my own clients, no issue.

I just thought that we weren't doing any more merges to the "Legacy" AAT anymore?

oliverfoster: It's been waiting for you to respond to the above question. It has security implications.

simondate: We might also need `target` so that links can open in a new tab/window.

@StuartNicholls / @fosterc1 can you think of anything else you would want?

